PR #71, opened 2015-01-12
s/MiniTest::Unit::TestCase/MiniTest::Test/g

mambocab:
I've been getting the following deprecation warning:

```
MiniTest::Unit::TestCase is now Minitest::Test. From hamming_test.rb:4:in `<main>'
```

So I've run

```bash
grep -rl MiniTest::Unit::TestCase . | sort | uniq | xargs sed -i s/MiniTest::Unit::TestCase/MiniTest::Test/g
```

I have not tested all of these -- I ran a couple by hand but make no promises.

If this causes compatibility problems, is there a way to suppress the deprecation warning?

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,15 +61,6 @@ Please do not add a README or README.md file to the problem directory. The
 READMEs are constructed using shared metadata, which lives in the
 [exercism/x-common](https://github.com/exercism/x-common) repository.
 
-## Minitest vs MiniTest
-
-Minitest v4.x (`MiniTest::Unit::TestCase`) ships with the language, whereas
-Minitest v5.x (`Minitest::Test`) does not. This means that people who have
-installed v5.x will get deprecation warnings.
-
-There's no really good way around this at the moment. We can't upgrade without
-forcing people to install an extra dependency.
-
 ## Contributing Guide
 
 For an in-depth discussion of how exercism language tracks and problem sets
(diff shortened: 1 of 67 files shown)


kytrinyx:
This is tricky. When we launched exercism, the stable ruby included MiniTest 4.x, and if we require Minitest 5.x then we need to explain `gem install` in the setup.

Also, in order to not get the deprecation warning, we'll need to say `gem 'minitest', '~> 5.0'` before we require anything, in case people have an older ruby with a newer minitest.

Since the most recent ruby now ships with a recent Minitest, I'm open to making the change, but I'd like to make sure that we get the documentation and everything fixed as well.

kytrinyx:
I've decided to go ahead and switch to using the recent Minitest.

Would you rebase onto master, please?

Also, the new minitest syntax is `Minitest::Test`, not `MiniTest::Test`.

mambocab:
Done -- sorry for the delay. I used this command:

    grep -rl MiniTest::Unit::TestCase . | sort | uniq | xargs sed -i s/MiniTest::Unit::TestCase/Minitest::Test/g

I've tested a few manually, and they run fine (and without deprecation warnings! :tada:), but I haven't done anything exhaustive.

I've also completely removed the caveat about testing libraries from the README.

kytrinyx:
Sweet, thanks!

Issue #77, opened 2015-01-24
gigasecond: use times (not dates) for inputs and outputs


kytrinyx:
A duration of a gigasecond should be measured in seconds, not
days.

The `gigasecond` problem has been implemented in a number of languages,
and this issue has been generated for each of these language tracks.
This may already be fixed in this track, if so, please make a note of it
and close the issue.

There has been some discussion about whether or not gigaseconds should
take daylight savings time into account, and the conclusion was "no", since
not all locations observe daylight savings time.

kotp:
When converting to Time, 46 minutes 40 second (approximately) discrepancy.

Compared to former tests, I mean
.

kytrinyx:
That sounds like daylight savings time. Is the Time in UTC or a local timezone?

kotp:
Daylight savings would not be off by 46 minutes, though.  An hour, maybe, as that is what DST does, but , it sounds like minute and second leaps.  Date class handles things like that, time class, not so much, from what I understand.

kytrinyx:
> A billion seconds, he told us, equals 31 years, 259 days, 1 hour, 46 minutes and 40 seconds.

I don't know if the `Date` tests are correct or the `Time` ones. I would have thought that `Time` accounts for these things, but I have not looked into it.

kotp:
So it looks like I have something good to work from, thanks.  Though for one, I did get 39 seconds, so that is what got me thinking about the leap seconds that get put into place.

kytrinyx:
Interesting. I really need to dig into all of this to get a better understanding of it. Time/Date/Timezones always makes me want to hide in a corner somewhere.

kotp:
Yes, it is definitely not a shallow subject.  Should have a pull request for you in a matter of moments.

re: #78

kotp:
I can't close this, but the pr #78 was merged and I assume this is closed?

kytrinyx:
Ah, yes. Thanks!

PR #78, opened 2015-01-25
Looking for time not date objects

kotp:
Now shows the 46 hours and 40 seconds, rather than a date.

Re: #77

### DIFF
--- a/gigasecond/example.rb
+++ b/gigasecond/example.rb
@@ -1,7 +1,7 @@
 class Gigasecond
   SECONDS = 10**9
 
-  def self.from(date)
-    (date.to_time + SECONDS).to_date
+  def self.from(time)
+    (time.to_time + SECONDS).to_time
   end
 end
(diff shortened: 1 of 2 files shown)


kytrinyx:
I'm curious - does it show the same offset if you use `Time.utc` as with `Time.new`?

kotp:
```
  1) Failure:
GigasecondTest#test_1 [gigasecond_test.rb:10]:
Expected: 2043-01-01 00:46:40 UTC
  Actual: 2043-01-01 00:46:40 -0500
```
When I use UTC it overrides the local time for the system it is running on.

kytrinyx:
Right, then the solution would have to use UTC as well. That might avoid some timezone confusion, actually.

kotp:
No, when I use utc for both the gs expectation and then use utc in the assertion, I still get the offset.

Looking into it, but I think that is due to the example solution... investigation.

kotp:
It shouldn't be, the time given is the time given.

kotp:
Leaving the utc out should leave the solution passing... so the request as it is should deal with the local run and pass.

kytrinyx:
Cool, thanks for digging into this!

PR #79, opened 2015-01-31
Remove lower-case letters from regex

IanWhitney:
In the README for this problem the letters are always in upper case. And
in the example solution, the `alphabet` method only returns upper case.
So I didn't think the tests needed to check for lower case letters.

### DIFF
--- a/robot-name/robot_name_test.rb
+++ b/robot-name/robot_name_test.rb
@@ -4,7 +4,7 @@ require_relative 'robot'
 class RobotTest < MiniTest::Unit::TestCase
   def test_has_name
     # rubocop:disable Lint/AmbiguousRegexpLiteral
-    assert_match /^[a-zA-Z]{2}\d{3}$/, Robot.new.name
+    assert_match /^[A-Z]{2}\d{3}$/, Robot.new.name
     # rubocop:enable Lint/AmbiguousRegexpLiteral
   end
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
Some people might implement a solution with lower case letters, though, and I'm not sure that would be wrong. It raises an interesting question, though: if they implement a solution with both uppercase and lowercase letters, would "aa000" be equivalent to "AA000"?

IanWhitney:
My typical approach when starting a new exercise is to get each test to pass in order. And since the test specified both lower & upper case, I implemented that.* But the example solution, which I know a lot of people look at, does just uppercase. And all of the README examples do the same.

In the end, I don't think it matters which way you go: either mixed case or all upper. I just think the example, readme and tests should all agree.

*Which, did you know there's no easy way to get a range of upper & lower case letters in Ruby? You end up with something like `('a'..'z').to_a + ('A'..'Z').to_a`.

kytrinyx:
I agree that the readme, solution, and test suite should all agree.

I'm fine with restricting the solutions to only upper-case, but in that case the test suite should protect against names with lower-case letters in them.

IanWhitney:
I think that's what my change does. Unless I'm misunderstanding you. 

```ruby
lower_name = "ab123"
upper_name = "AB123"
mixed_name = "aB123"
/^[A-Z]{2}\d{3}$/.match(lower_name)
#=> nil
/^[A-Z]{2}\d{3}$/.match(upper_name)
#=> #<MatchData "AB123">
/^[A-Z]{2}\d{3}$/.match(mixed_name)
#=> nil
```
So if a person implements a lower-cased or mixed-case name generator, the test suite will fail. 

Or are you asking for an explicit test of case so that the person sees a "RobotTest#test_name_is_upper_case" message when they run their tests?

kytrinyx:
Oooooh. Sorry, I completely misunderstood the change. Not sure what I was thinking.

Yes, this works. Thank you!

PR #80, opened 2015-02-04
Change test names to be easier to understand

benoittgt:
I had some problems when doing the exercice because of the DST. I think more explicit method's name should help.

Thanks


kytrinyx:
This works for me. Thanks!

PR #83, opened 2015-02-05
prevent empty linked lists from being represented as nil

zvkemp:
List operations `reverse` and `to_a` should be valid on an empty list. `to_a` is handled by `NilClass#to_a`, but `reverse` is not. The new test expects that an empty list can be represented by a non-nil object. The updated example passes this test.


kytrinyx:
I have a pretty strong aversion to having any object who is not actually `nil`, the instance of `NilClass` respond to `nil?` with true.

@markijbema Thoughts?

zvkemp:
@kytrinyx, in general, I agree. This seemed the simplest way to make this change without a more significant rewrite, but I'll give it some more thought.

zvkemp:
Here's another version with a custom `null?` query method and updated assertions:

https://github.com/zvkemp/xruby/commit/70d1674cef594503863c3a30c562d8e007d8b07a

Up to you whether you feel it's worth it, or if this is an appropriate level of effort required for this test. I felt there should be a way to maintain the list context even when empty, which `nil` doesn't provide.

kytrinyx:
> I felt there should be a way to maintain the list context even when empty, which nil doesn't provide.

I agree. I think that a NullObject is appropriate, it's only that I feel that `nil?` should respond `true` only if it's an instance of `NilClass`.

Would there be a more appropriate interface? How about a `last?` method on Element and NullElement? Or perhaps `has_next?`? Then we could avoid ever exposing nils.

kytrinyx:
@zvkemp We had a bit of an accident with some very large binaries getting written to the history of the repo. Let me know if you need a hand in getting master cleaned up (so sorry about the mess).

Did you have a chance to think about a way to never expose the nil/null?

kotp:
I think [this is the only commit](https://github.com/zvkemp/xruby/commit/b98cf15547d11116d7146a7eabab0614c040d9f4) that really matters in this entire PR.  if this is true, can we clean it up to look at it?

kytrinyx:
I've cherry-picked @zvkemp's commit into a separate branch, and submitted it as a PR here: https://github.com/exercism/xruby/pull/139 -- the commit itself is still attributed to @zvkemp, but the PR has my name on it.

PR #84, opened 2015-02-08
Change buffer.write to buffer.write!

stevepm:
I was working on this problem earlier and found that changing this method seemed to make sense in terms of what the tests were asserting.

### DIFF
--- a/circular-buffer/circular_buffer_test.rb
+++ b/circular-buffer/circular_buffer_test.rb
@@ -95,7 +95,7 @@ class CircularBufferTest < MiniTest::Unit::TestCase
     buffer.read
     buffer.write '4'
     buffer.read
-    (5..8).each { |i| buffer.write String(i) }
+    (5..8).each { |i| buffer.write! String(i) }
     buffer.write! 'A'
     buffer.write! 'B'
     (6..8).each do |i|


kytrinyx:
Would you elaborate on why it makes sense, please?

stevepm:
Based on the assertions on lines 101-103.
We would only be asserting that '6', '7', and '8' are in the buffer, if the `write!` method were called.

If the `write` method was called, those numbers would not be in the buffer because it would already be full.

kytrinyx:
I still don't see it.

The buffer has 5 slots

```
  _ _ _ _ _
^
```

```
  1 2 3 _ _
^
```


We read twice:

```
_ _   3 _ _
     ^
```

Then we write once
```
_ _   3 4 _
     ^
```

Read once:

```
_ _ _  4 _
        ^
```

Then we write four times:

```
6 7 8  4 5
        ^
```

The buffer was not full, so all of the numbers go into it, and there is no `CircularBuffer::BufferFullException`.

stevepm:
I see. I was mistaken. I didn't realize that the `read` method removed that element from the buffer. It wasn't clear to me from the readme. No issue then, thanks for clearing it up!

kytrinyx:
Thanks for taking the time to dig into it.

PR #85, opened 2015-02-13
Missing value for expected variable in verse 8

alexpchin:
Missing value for expected variable in verse 8


kytrinyx:
Nice catch, thank you!

PR #86, opened 2015-02-13
Skipping setup for later tests

alexpchin:
Skipping setup for later tests so that user only needs to deal with one test at a time.


kytrinyx:
Thanks!

PR #88, opened 2015-02-17
Make references to minitest module consistent across all exercises

chrisvroberts:
Three of the exercise directories (gigasecond, largest-series-product and protein-translation) contain *_test.rb files which refer to the Minitest module rather than the MiniTest module. For consistency these references have been changed to MiniTest.


kytrinyx:
Thanks!

PR #90, opened 2015-02-17
Add new scale-name test to scale_generator_test.rb

chrisvroberts:
Only one test (test_naming_scale) currently asserts against Scale#name meaning that a compliant implementation could contain a hard coded value of 'C chromatic'. This change adds a new test (test_naming_major_scale) which asserts against Scale#name with a different tonic and scale-name.


fluxusfrequency:
Looks good, but can you add a `skip` to the top of that test to match the others?

kytrinyx:
Cherry-picked in 4d650524c7af3495c5a060d50d60ebdbc9309e17

I fixed the `skip`.

PR #91, opened 2015-02-23
squares: subsequent calls should have same value

bjmllr:
Since the API calls for a `Squares` object to be instantiated with a specific max value, it seems to be implied that the instance methods of a given instance should have the same result on subsequent calls. While not common, I have seen more than one submission that didn't have this property.


kytrinyx:
Thanks!

PR #94, opened 2015-03-13
Add exception handling for wrong nucleotides.

dkinzer:
I noticed that other language versions of this exercise handle argument errors
for the nucleotides.  This commit adds some new tests to check against bogus
nucleotides.


kytrinyx:
Thanks

dkinzer:
Thanks!

Issue #95, opened 2015-03-15
Using the Prime module on the nth-prime test


Carpetsmoker:
Do we want to encourage the usage of Ruby's standard `Prime` module for the nth-prime test?

I wanted to make a patch to prevent using the `Prime` module, as I would consider that to be slightly cheating and defeating the purpose of the exercise, but the `example.rb` seems to use it as well ...

kotp:
I saw that in recent solution, and I thought the same thing.  But only if the [readme](https://github.com/exercism/x-common/blob/master/nth-prime.yml) states to use built in language items only, otherwise in my opinion, it would be fair game.

https://github.com/exercism/xruby/tree/master/nth-prime for ease of reference.

And for ease of reference to which '[example.rb](https://github.com/exercism/xruby/blob/master/nth-prime/example.rb)' uses it just in case that changes as well. 

This isn't to say that you can't test for it and report encouragement.  But again, this may be a function of the 'nitpick' activity.

kytrinyx:
> Do we want to encourage the usage of Ruby's standard Prime module for the nth-prime test?

No, I would want to discourage that. I would welcome a patch that prevents the use of `Prime`. I already did this in `Gigasecond` to prevent the use of date.

kotp:
Looks like it is adequately fixed.

PR #96, opened 2015-03-15
Prevent people from using Ruby's Prime class to implement the nth-prime exercise

Carpetsmoker:
See: Issue #95


nth-prime/nth_prime_test.rb
```diff
@@ -1,4 +1,22 @@
 require 'minitest/autorun'
+
+require 'prime'
+
+Error_message = "Using Ruby's Prime class is probably the best way to do this in a 'real'\n" +
```
kytrinyx: I think the constant should either be CamelCase or SHOUT_CASE, but not a mix of Camel and snake_case.

kytrinyx:
Cherry-picked in 0e55df100d238c7f8dbed1e4cdea6bb183125b15

I rewrote the history to remove some very large, accidentally committed binaries.

PR #97, opened 2015-03-18
Added acronym problem

monkbroc:
I had an idea for a problem: shorten a string to its acronym. Like Portable Network Graphics :arrow_right: PNG.

I wrote tests and an example solution in Ruby. I don't know where a README should go, but I'm submitting this so you can review the idea.

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,6 +8,7 @@
     "gigasecond",
     "rna-transcription",
     "raindrops",
+    "acronym",
     "difference-of-squares",
     "roman-numerals",
     "robot-name",
--- a/acronym/acronym_test.rb
+++ b/acronym/acronym_test.rb
@@ -1,0 +1,33 @@
+require 'minitest/autorun'
+require_relative 'acronym'
+
+class AcronymTest < Minitest::Test
+  def test_png
+    assert_equal 'PNG', Acronym.abbreviate('Portable Network Graphics')
+  end
+
+  def test_ruby_on_rails
+    skip
+    assert_equal 'ROR', Acronym.abbreviate('Ruby on Rails')
+  end
+
+  def test_html
+    skip
+    assert_equal 'HTML', Acronym.abbreviate('HyperText Markup Language')
+  end
+
+  def test_fifo
+    skip
+    assert_equal 'FIFO', Acronym.abbreviate('First In, First Out')
+  end
+
+  def test_php
+    skip
+    assert_equal 'PHP', Acronym.abbreviate('PHP: Hypertext Preprocessor')
+  end
+
+  def test_cmos
+    skip
+    assert_equal 'CMOS', Acronym.abbreviate('Complementary metal-oxide semiconductor')
+  end
+end
--- a/acronym/example.rb
+++ b/acronym/example.rb
@@ -1,0 +1,15 @@
+class Acronym
+  def self.abbreviate(phrase)
+    [].tap do |letters|
+      each_word(phrase) do |word|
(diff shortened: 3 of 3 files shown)


kotp:
The readme goes into https://github.com/exercism/x-common as per the README file.

monkbroc:
Ok. Thanks. Should I make a PR to that repository or wait until this new
problem is discussed and ok'ed before doing that?

kotp:
Go ahead and create the README with a pull request, so @kytrinyx can see the whole picture.  Then it is a simple accept or deny action if it is that simple, and any issues can be addressed at the proper location, perhaps with references to the other in the comments.

monkbroc:
OK. I added the metadata to [the x-common repository in this pull request.](https://github.com/exercism/x-common/pull/61).

I also added the new problem to the Ruby track to get the Travis build to pass.

kytrinyx:
Neat!

In terms of the API, I think I'd prefer to make it a class or module function. Then people can decide whether or not to delegate to an instance or to leave it as a purely functional thing.

Also, how about `abbreviate` rather than `short`?

kytrinyx:
Oh, one more thing. Let's move it even earlier in the progression, perhaps right after raindrops?

config.json
```diff
@@ -37,6 +37,7 @@
     "scrabble-score",
     "nucleotide-count",
     "hexadecimal",
+	"acronym",
```
kytrinyx: Would you fix the indentation here?

monkbroc:
I renamed the method to `abbreviate` and made it a class method.

When you say "put it earlier in the progression" I assume it means the problems are supposed to go in increased difficulty? It's not something that comes across to me at all. It felt more like one exercise after the other, all of them fairly straightforward. Is it something that is important to the experience of Exercism in your opinion?

kytrinyx:
> I assume it means the problems are supposed to go in increased difficulty

I can only say "ish". For an experienced programmer they're all trivial, but for beginners there are definitely concepts that get harder further out. It's very hand-wavey, and I don't know that it's core to the experience.

Merging. Thanks!

monkbroc:
Yeah, my feeling while doing the Ruby exercises is that the best ones are the ones that give more opportunities to comment on style while not being tedious. Difficulty doesn't seem to play a big role.

kytrinyx:
Agreed. Some problems have a lot of different opportunities, others are like "write this one liner. Okay? Okay!"

I'd actually be interested in figuring out which ones could be retired.

monkbroc:
What about a simple +1/-1 vote button for each exercise? "This is a great exercise", "Not so good"

This way the data will show which ones are most appreciated and which ones are most disliked.

kytrinyx:
Yeah, I think that could work. We could have it both in the exercise page and on the dashboard (for exercises that you haven't voted on). Would you mind opening a new issue for this?

kytrinyx:
Actually, maybe this is part of this discussion: https://github.com/exercism/exercism.io/issues/1081

monkbroc:
@kytrinyx This problem is too early in the progression. The combination of test cases is causing headaches to people not too familiar with regular expressions. [One person said he worked on it for 90 minutes and couldn't get all test cases to pass.](http://exercism.io/submissions/d618971d10984a1fb59c5c3596047ecd)

I think it should be moved back later or the problem made easier. What's your thought?

kytrinyx:
Wow, yeah, 'm happy to move it to a later spot.

PR #100, opened 2015-03-26
Auto run all tests for CI.

dkinzer:
This commit ports over the auto test solution implemented in the xphp
track.

In order to be able to to autorun all these tests in a sensible manner
I made changes to filenames and imports such that names of imported
files corresponded to their test name and the name of their respective 
directories.

Some of the tests are currently failing because of bugs in the example code.
These test will not get run unless we replace `skip("<Reason>")` with `skip`.

The following tests fail:
```
crypto_square/crypto_square_test.rb:72:    skip("Skip")
crypto_square/crypto_square_test.rb:84:    skip("Skip")
crypto_square/crypto_square_test.rb:100:    skip("Skip")
custom_set/custom_set_test.rb:55:    skip("No method uniq for Range(1..3)")
gigasecond/gigasecond_test.rb:32:    skip("Keep")
ocr_numbers/ocr_numbers_test.rb:6:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:17:    skip("Expected 1, Actual ''")
ocr_numbers/ocr_numbers_test.rb:28:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:39:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:50:    skip("Expected 4, Actual ''")
ocr_numbers/ocr_numbers_test.rb:61:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:72:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:83:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:94:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:105:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:116:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:127:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:138:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:149:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:160:    skip("Undefined method [] for nil")
ocr_numbers/ocr_numbers_test.rb:171:    skip("Undefined method [] for nil")
```


dkinzer:
Hmmm... not sure why travis build is failing with test not being able to run.  They run fine locally.  Maybe I need to prep the travis config with something ruby specific?

dkinzer:
Ah, travis language is set to bash.  Maybe that's it?

dkinzer:
Looks like I've made matters worse.  I'm stumped as to why this is failing on Travis. Any ideas?

dkinzer:
I think it may have something to do with the version of minitest being used:

https://forums.manning.com/posts/list/32024.page;jsessionid=1B8C60D05F0DDED33162957EC9B8DDF0

dkinzer:
OK, adding minitest as a gem did the trick.  Now to fix config.json

kytrinyx:
This looks great, sorry I didn't see the notifications along the way. I'll pull this down and test it out when I get home.

dkinzer:
Cool. I'm not 100% sure that the whole dash to underscore is the right thing to do, but that can be changed.

kytrinyx:
When I looked at this briefly earlier I was in the middle of a hackathon, and hadn't taken the time to read all the way to the end.

We can't change the names of the exercises in config.json -- these refer to

a) the directory that the problem lives in, and
b) the metadata files in https://github.com/exercism/x-common

dkinzer:
OK, this can still be solved without needing to change these names.  So is it just the names of the problems that can't be changed or is it also the names of the directories that the problems live in that cant' be changed?  Are the names of the directories and the names of the problem always related in some way?  I think that a solution that doesn't follow some kind of convention will always be brittle or not be applicable across all the different tracks (that be nice).

Hmmm, now that I know about config.json I realize that it may actually be possible to solve this just relying on that fact.

dkinzer:
Sorry, didn't read your answer carefully enough.  You already said that "the slug in the config file, refers to the directory of the problem" :)

dkinzer:
OK, I'm closing this in favor of #102

kytrinyx:
Yeah, eventually it might be nice to have something less brittle than naming conventions, but up until now it has worked remarkably well, and the entire problem API is backed only by the filesystem; no database necessary. I think there are other, much larger problems, to solve before this becomes important.

PR #101, opened 2015-03-27
Fix test suite for robot-name excercise

StevenXL:
Consider a robot whose `reset` state is set to *true*, and this state is used to determine if a new name should be generated and returned. 

The current test suite will pass even if the programmer forgets to "unset" the robot's `reset` state after a call to name, thereby generating and returning a new name whenever the `name` method is called, ad infinitum. 

The proposed changes to the test suite would test that the `reset` state has been set to "off" or its equivalent.

### DIFF
--- a/robot-name/robot_name_test.rb
+++ b/robot-name/robot_name_test.rb
@@ -29,6 +29,7 @@ class RobotTest < Minitest::Test
     robot.reset
     name2 = robot.name
     assert name != name2
+    assert_equal name2, robot.name, "Command/Query Separation: query methods should generally not change object state"
     # rubocop:disable Lint/AmbiguousRegexpLiteral
     assert_match /^[A-Z]{2}\d{3}$/, name2
     # rubocop:enable Lint/AmbiguousRegexpLiteral


robot-name/robot_name_test.rb
```diff
@@ -29,6 +29,7 @@ def test_reset_name
     robot.reset
     name2 = robot.name
     assert name != name2
+    assert_equal robot.name = name2
```
kytrinyx: Did you try running this assertion? The api of `assert_equal` takes two arguments, plus an optional argument for the failure message.

Since the code does `name2 = robot.name` on line 30, it doesn't really make sense to assign the same value back with `robot.name = name2`.

StevenXL:
@kytrinyx see updated commit.

robot-name/robot_name_test.rb
```diff
@@ -29,6 +29,7 @@ def test_reset_name
     robot.reset
     name2 = robot.name
     assert name != name2
+    assert_equal robot.name, name2
```
kytrinyx: I feel like this tests Ruby itself. You do the assignment on line 30, only to verify that the assignment worked.

I may not understand the edge case that you're trying to avoid. Could you show a code example that has the bug that would pass the old test suite and fail with this one?

StevenXL:
Sure.

Below is a solution that passes all the tests in the current test suite, but will fail the proposed test suite. 

```ruby
class Robot
  ALBHABET = ("A".."Z").to_a
  DIGITS = (0..9).to_a

  def initialize
    @name = generate_name
    @reset = false
  end

  def reset
    @reset = true
  end

  def name
    if @reset
      @name = generate_name
    end
    @name
  end

  private

  def generate_name
    "#{ALBHABET.sample(2).join}#{DIGITS.sample(3).join}"
  end
end
```

The key here is the `name` method. It checks the `@reset` variable, and if that variable evaluates to true, it generates and returns a new name for the robot. That's all well and good, except that it has to then set the `@reset` variable to evaluate to false. Otherwise, all subsequent calls to `name` will generate a new name. 

To make the code above pass with the proposed test suite, the `name` method would have to look like this:
```ruby
  def name
    if @reset
      @name = generate_name
      @reset = false
    end
    @name
  end
```

StevenXL:
I'm worried now that I messed something up. The Travis CI is taking too long to pass.

kytrinyx:
Every once in a while Travis blocks on fetching dependencies.

kytrinyx:
OK, so this is interesting. It never occurred to me to implement the `name` method in this way. A method that returns a value should generally not change the state of the object (this is often referred to as _command/query_ separation).

I think that it's interesting to catch this, but I would want to add a failure message that encourages the author of the solution to look into the concept of command/query separation.

StevenXL:
@kytrinyx Great - I'll add the message and push the changes this evening. (And of course, I'll look into command/query separation myself since the solution I posted is my own).

StevenXL:
Updated with failure message.

kytrinyx:
Thanks for the contribution, and enjoy the exploration of command/query :)

PR #102, opened 2015-03-27
Auto run all tests for CI.

dkinzer:
This pull request accomplishes the same as exercism/xruby#100 without changing
any slugs in the config.json folder.

I still ended up having to change some of the name of some of the files loaded and also the names of some of the test files to conform to what's on the config.json.


dkinzer:
Hmm.. I don't think that failure is real.  All I did was remove script I used to clean up some of the name discrepancies.

kytrinyx:
Yeah, you can change the names of all the files inside the test folder. The only restriction is that the example file needs to have /[Ee]xample/ in the name, and the test file is assumed to follow the .*_test.rb format (though I think that the code allows for a much looser definition).

kytrinyx:
This looks very nice. Using hyphens for the filenames is unusual, but that might be worth it. The only worry that I have is that newbies will see this and think it's a convention, and then use it in their projects.

What do you think of saying `tr '-' '_'` when renaming the example file to the assignment slug so that we follow ruby conventions?

dkinzer:
yeah, that makes senses.

dkinzer:
@kytrinyx this now uses _ for all tests and source files by tr-ing the - to _.

kytrinyx:
I love this, thanks so much for taking the time to work it out!

dkinzer:
Thanks!

PR #103, opened 2015-03-27
Space Age: Don't compare floats using ==

monkbroc:
It's bad form to compare floats with each other using == and Exercism shouldn't be showing this to aspiring programmers.

[As the standard lib docs says](http://ruby-doc.org/stdlib-1.9.2/libdoc/minitest/unit/rdoc/MiniTest/Assertions.html#method-i-assert_in_delta) `asset_in_delta` should be used to compare floats.

The modified test pass with code designed for the old test (i.e. with `round(2)` after the division) and allows new code not to have to round the division which feels unnatural (it should not a responsibility of
`SpaceAge` to round its output).

### DIFF
--- a/space-age/example.rb
+++ b/space-age/example.rb
@@ -17,7 +17,7 @@ class SpaceAge
   }.each do |planet, orbital_period|
 
     define_method("on_#{planet}") do
-      (seconds / orbital_period).round(2)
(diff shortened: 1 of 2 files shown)


kytrinyx:
Yes, this is a definite improvement. Thanks!

PR #106, opened 2015-03-30
Fetch appropriate configlet binary.

dkinzer:
I noticed that when I am working on my Mac the fetched configlet binary fails.

This pull request adds a limited but smarter configlet binary fetcher that
tries to determine the OS and architecture and fetch the appropriate
configlet binary.


dkinzer:
@kytrinyx I wrote a script to propagate the pull request across all the tracks but I triggered some kind of abuse detection thing.  I need to wait to create pull requests for the remaining repos.

bin/fetch-configlet
```diff
@@ -2,7 +2,29 @@
...
+        echo "windows";;
+    *)
+        echo "linux";;
+esac)
+
+ARCH=$(
+case $(uname -m) in
+    *64*)
+        echo 64bit;;
+    *32*)
```
kytrinyx: Are you sure about this one? I think `i386` is the classic response for a 32 bit architecture.

kytrinyx:
This is great. Anything to reduce the friction in getting a clean build!

kytrinyx:
@dkinzer A couple things:

- Would you double-check the correct `uname -m` for 32 bit architectures?
- I apparently had a configlet binary in there from before, from waaay back... so lots of new history here. Let me know if I should write up a quick how-to on cleaning this up against the new master.

dkinzer:
@kytrinyx you're totally right, `uname -m` will return either i386 or i686 for 32 bit. (I need to stop making guesses).  I'll correct the patches.

dkinzer:
I think it will be easier to close this and start a new one because of the merge conflicts.

kytrinyx:
OK, cool. Sorry about the mess!

dkinzer:
No problem!

PR #107, opened 2015-03-31
Move the acronym problem later

monkbroc:
As I mentionned in #97 the acronym problem is too early and is causing frustration to beginners.

This PR moves it to the middle of the progression (after clock).


kytrinyx:
Good call, thanks!

PR #109, opened 2015-04-09
Accumulate: add test to forbid changing in place

broadfootp:
Without this test, the test suite will pass even if #accumulate is altering the original object.

This is my first pull request to a public project, apologies if I made a mistake :-)


kytrinyx:
This looks great, thank you!

PR #110, opened 2015-04-14
use refute rather than "assert !"

dalexj:
it's a little more descriptive


kytrinyx:
I totally agree. Thanks!

PR #112, opened 2015-04-15
Add link to exercism help in setup.md

tejasbubane:
It will be better if the link to exercism help page for ruby is provided
in the README. In fact, it should be provided for other languages as
well.

Also make it clear that minitest is required for running the tests and
not for solving the exercise in particular.


tejasbubane:
Oops! I should have skipped CI.

kytrinyx:
This is better, thank you.

PR #114, opened 2015-04-16
Modify tests so that identical instances are not necessary

StevenXL:
The current test suite requires new instances to be created by sending the `new` message to the `Grains` class. 

These new instances are exactly identical at time of creation, and won't differ in the future. It therefore doesn't make sense to create them. 

I have changed the test suite to test class methods instead of instances methods, and therefore avoid a solution that requires new instances. This rewrite is in keeping with the other exercises in the Ruby suite.

### DIFF
--- a/grains/example.rb
+++ b/grains/example.rb
@@ -1,11 +1,9 @@
 class Grains
-  def square(number)
+  def self.square(number)
     2**(number - 1)
   end
 
-  def total
-    @total ||= (1..64).inject(0) do |sum, number|
-      sum + square(number)
-    end
+  def self.total
+    square(65) - 1
   end
 end
--- a/grains/grains_test.rb
+++ b/grains/grains_test.rb
@@ -3,41 +3,41 @@ require_relative 'grains'
 
 class GrainsTest < Minitest::Test
   def test_square_1
-    assert_equal 1, Grains.new.square(1)
+    assert_equal 1, Grains.square(1)
   end
 
   def test_square_2
     skip
-    assert_equal 2, Grains.new.square(2)
+    assert_equal 2, Grains.square(2)
   end
 
   def test_square_3
     skip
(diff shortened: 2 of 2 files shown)


grains/example.rb
```diff
@@ -1,10 +1,10 @@
 class Grains
-  def square(number)
+  def self.square(number)
     2**(number - 1)
   end
 
-  def total
-    @total ||= (1..64).inject(0) do |sum, number|
+  def self.total
+    @@total ||= (1..64).inject(0) do |sum, number|
```
kytrinyx: I know this is just an example, but I would like to avoid using class instance variables. These are an invitation to create race conditions, and I feel like they should be avoided.

There really is no reason to save state at all in this exercise:


```
def self.square(number)
  1 << (number-1)
end

def self.total
  square(65) - 1
end
```

kytrinyx:
Agreed, there's absolutely no reason to introduce state in this problem.

StevenXL:
Pull request modified per feedback.

kytrinyx:
Thanks!

kotp:
And when it is stateless, it indicates a Module rather than a Class, I think.

kytrinyx:
Yeah, I'm always sloppy about that, but you're totally right.

StevenXL:
Thanks Victor. I see that the code has been updated. Glad to have that
nugget of knowledge.

On Thu, Apr 16, 2015 at 5:01 PM, Katrina Owen <notifications@github.com>
wrote:

> Yeah, I'm always sloppy about that, but you're totally right.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/exercism/xruby/pull/114#issuecomment-93838676>.
>



-- 
Steven Leiva
305.528.6038
leiva.steven@gmail.com
http://www.linkedin.com/in/stevenleiva

PR #119, opened 2015-04-25
Hamming generate tests

r11runner:
hamming exercise:
 - added an example generator
 - regenerated the tests with it

### DIFF
--- a/hamming/example.rb
+++ b/hamming/example.rb
@@ -1,6 +1,7 @@
 class Hamming
   def self.compute(strand1, strand2)
-    (0...[strand1.length, strand2.length].min).count do |i|
+    raise ArgumentError, "The two strands must have the same length." if strand1.length != strand2.length
+    (0...strand1.length).count do |i|
       strand1[i] != strand2[i]
     end
   end
--- a/hamming/example_gen.rb
+++ b/hamming/example_gen.rb
@@ -1,0 +1,81 @@
+require 'json'
+
+PREFIX_PART1="require 'minitest/autorun'
+begin
+  require_relative 'hamming'
+rescue LoadError => e"
+PREFIX_PART2 ='
+  puts "\n\n#{e.backtrace.first} #{e.message}"
+  puts DATA.read
+  exit 1
+end
+
+class HammingTest < Minitest::Test
+'
+
+SUFFIX="end
+
+__END__
+
+*****************************************************
+You got an error, which is exactly as it should be.
+This is the first step in the Test-Driven Development
+(TDD) process.
+
+The most important part of the error is
+
+      cannot load such file
+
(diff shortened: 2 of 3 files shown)


kytrinyx:
`make test` is failing with the following error:

```
  1) Failure:
HammingTest#test_disallow_first_strand_longer [/tmp/hamming.BsRYhY7oVw/hamming_test.rb:61]:
ArgumentError expected but nothing was raised.
  2) Failure:
HammingTest#test_disallow_second_strand_longer [/tmp/hamming.BsRYhY7oVw/hamming_test.rb:65]:
ArgumentError expected but nothing was raised.
```

Is the solution updated to raise an argument error?

r11runner:
Sorry, I didn't think of it.
Now, I've committed a further change which changed the example implementation accordingly.

kytrinyx:
Awesome!

@markijbema @henrik I know we've been back and forth on whether or not we should have an error case. I'm back to wanting to explicitly handle the error.

In particular, the reason that I wanted to not handle the error, is that this is the first exercise. But now we're about to add "hello world" as the first exercise, so that's not going to be a problem any more.

Thoughts?

henrik:
Was this ever the first exercise? Thought that was Bob?

Is it a goal of Exercism to have a progression from easier to more challenging exercises? Or to at least have that for the first few exercises?

I like about the error case that it introduces another special case to be handled, perhaps making your beautiful solution less so. Just like the real world ;) But I can see that we should skip it if that helps with this progression.

kytrinyx:
This has been the first exercise for a while. It's simpler than Bob, because there's no state in Hamming.

kytrinyx:
Thanks @r11runner -- the Hello World exercise went live yesterday, so this is ready to go.

r11runner:
Thx, great :)

markijbema:
In my opinion, hamming has no meaning for strands of different lengths. So it depends on whether you see this as a standalone program which checks user input, or an internal method (which should be confident) which can just crash by assuming equal lengths. Personally I'm in favour of not checking, but I don't care too much either way.

kytrinyx:
I've gone back and forth, and have ended up leaning more towards the error checking (treating it like a stand-alone library). I think that a good argument can be made either way, and I like the slightly added complexity of having to think about the errors.

markijbema:
I think the most important risk is that people write 'insecure' code (as in opposite of what @avdi talks about in confident code). Since the examples never get large the difference between checking internally and externally isn't really visible. Maybe an exercise where this difference is clear is useful, but I wouldn't really know how to do that without making it large.

kytrinyx:
> I wouldn't really know how to do that without making it large.

Same here.

PR #120, opened 2015-04-25
Add introductory hello-world exercise

ramonh:
This addresses exercism/rossconf#2 towards goal [#12](exercism/rossconf#12).

The hello-world exercise has three test cases, one for no name given, one for the name 'Alice' given and one for the name 'Bob' given. This is to avoid the hardcoding of a name into the solution.

### DIFF
--- a/hello-world/example.rb
+++ b/hello-world/example.rb
@@ -1,0 +1,5 @@
+class HelloWorld
+  def self.hello(name = '')
+    name == '' ? 'Hello, world!' : "Hello, #{name}!"
+  end
+end
--- a/config.json
+++ b/config.json
@@ -4,6 +4,7 @@
   "repository": "https://github.com/exercism/xruby",
   "active": true,
   "problems": [
+    "hello-world",
     "hamming",
     "gigasecond",
     "rna-transcription",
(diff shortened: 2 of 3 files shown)


kytrinyx:
This looks great. I'm going to hold off on merging it until we have the code in place to ignore hello-world in the nitpicking list, and to submit an encouraging comment by rikki-.

Would you submit a separate PR with the `SETUP.md`? That would be useful to have even before hello world goes live.

ramonh:
Thanks for the feedback! I've removed the SETUP.md commit.

kytrinyx:
:heart: here we go!

ramonh:
Wahey! Awesome, still getting you that Obj-C version. :smile:

kytrinyx:
Niiiice!

hello-world/example.rb
```diff
@@ -0,0 +1,5 @@
+class HelloWorld
+  def self.hello(name = '')
+    name == '' ? 'Hello, world!' : "Hello, #{name}!"
```
kotp: This is a little bit "tricky". You already have a default value, why not just use it?

``` ruby
def self.hello(name = 'World')
  "Hello, #{name}!""
end
```

Least amount of complication.

PR #122, opened 2015-04-28
gigasecond: added example generator and updated the tests

r11runner:
side effects:
 - 1 additional test method
 - the naming of the test methods has changed


kytrinyx:
Thanks!

PR #124, opened 2015-05-02
Add tests for write! edge cases in Circular Buffer

### LINKED ISSUE #123: Missing edge cases in Circular Buffer tests.
None of the tests for the Circular Buffer exercise use `write!` with a `nil` argument or on a buffer that isn't full. I'm assuming that in both cases `write!` should perform identically to `write`, but many simple implementations of `write!` are likely to get these cases wrong.

I'd be happy to add test cases to address those issues myself!

JulianAndrews:
Tests for "CircularBuffer.write!" with "nil" argument and on non-full
buffers. Fixes #123.

### DIFF
--- a/circular-buffer/circular_buffer_test.rb
+++ b/circular-buffer/circular_buffer_test.rb
@@ -86,6 +86,26 @@ class CircularBufferTest < Minitest::Test
     assert_raises(CircularBuffer::BufferEmptyException) { buffer.read }
   end
 
+  def test_forced_writes_of_nil_should_not_occupy_buffer
+    skip
+    buffer = CircularBuffer.new(2)
+    (1..2).each { |i| buffer.write String(i) }
+    buffer.write! nil
+    assert_equal '1', buffer.read
+    assert_equal '2', buffer.read
+    assert_raises(CircularBuffer::BufferEmptyException) { buffer.read }
+  end
+
+  def test_forced_writes_to_non_full_buffer_should_behave_like_writes
+    skip
+    buffer = CircularBuffer.new(2)
+    buffer.write '1'
+    buffer.write! '2'
+    assert_equal '1', buffer.read
+    assert_equal '2', buffer.read
+    assert_raises(CircularBuffer::BufferEmptyException) { buffer.read }
+  end
+
   # rubocop:disable Metrics/MethodLength
   def test_alternate_read_and_write_into_buffer_overflow
     skip


circular-buffer/circular_buffer_test.rb
```diff
@@ -86,6 +86,26 @@ def test_overwriting_oldest_item_in_a_full_buffer
     assert_raises(CircularBuffer::BufferEmptyException) { buffer.read }
   end
 
+  def test_forced_writes_of_nil_should_not_occupy_buffer
+    skip
+    buffer = CircularBuffer.new(2)
+    (1..2).each { |i| buffer.write String(i) }
+    buffer.write! nil
```
kytrinyx: This is actually an interesting question overall... it seems odd to me now that a `nil` would be treated differently than  any another value, but I see that we're already doing this for the `write` with a `nil`.

Let's do this for now, but also be open to having the discussion about what `nil` actually means in the context of a circular buffer.

kytrinyx:
Thanks!

circular-buffer/circular_buffer_test.rb
JulianAndrews: I had the a similar thought. It's notable that all of the test cases only use strings as values for the buffer. If the intent is a circular buffer for strings only, then probably instead of silently failing on a `nil` write, the buffer should throw an exception on any non-string value.

If instead the intent is a buffer that can accept arbitrary objects, then `nil` should be a valid thing to write. Of course, that would make the exercise a notch more difficult since the simple approach of using a `nil` filled array wouldn't work anymore, but it would also be more general an correct.

Since the buffer is suppose to raise an error if you try to read from it empty, there's no scenario where `nil` is used to represent an empty value or anything like that.

kytrinyx: Agreed, I like the idea of making this a circular string buffer, and to have any non-string values raise errors. We could have several tests to that effect (nil, fixnum, Object.new, whatever), or one test with several different types of things.

JulianAndrews: I could write a fix to replace the `nil` `write` tests with tests that check a representative set of non-string values all throw an `ArgumentError` (or some custom exception). In that case it might also make sense to have some of the tests test longer strings (not just single character strings) just to make it clear that any strings (but only strings) are valid inputs.

kytrinyx: Yes, longer strings would help clarify.

PR #126, opened 2015-05-09
Fixes the incorrect crypto-square tests

alexkalderimis:
The tests for crypto-square expected incorrect rows in the normalised ciphertext, contradicting the README and other more conformant implementations. In addition the example itself was failing some of the existing (correct) tests.


kytrinyx:
Thanks for catching this.

PR #127, opened 2015-05-09
Use travis.ci containers for faster builds.

alexkalderimis:
This change-set adds the `sudo` flag which allows travis to use containers rather than VMs, which means
much faster build allocation.


kytrinyx:
Awesome! Thanks

PR #128, opened 2015-05-11
Default name of 'world' solves the no name case

kotp:
A follow up to my comments in PR #120

### DIFF
--- a/hello-world/hello_world_test.rb
+++ b/hello-world/hello_world_test.rb
@@ -1,8 +1,12 @@
 require 'minitest/autorun'
 require_relative 'hello_world'
 class HelloWorldTest < Minitest::Test
+  def test_no_strange_name
+    assert_equal 'Hello, !', HelloWorld.hello(''), 'Greeting given an empty string as a name'
+  end
+
   def test_no_name
-    assert_equal 'Hello, world!', HelloWorld.hello(''), 'Greeting given no name'
+    assert_equal 'Hello, world!', HelloWorld.hello, 'Greeting given an empty string as a name'
   end
 
   def test_sample_name
--- a/hello-world/hello_world_test.rb
+++ b/hello-world/hello_world_test.rb
@@ -2,18 +2,18 @@ require 'minitest/autorun'
 require_relative 'hello_world'
 class HelloWorldTest < Minitest::Test
   def test_no_strange_name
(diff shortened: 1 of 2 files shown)


hello-world/hello_world_test.rb
```diff
@@ -1,8 +1,12 @@
 require 'minitest/autorun'
 require_relative 'hello_world'
 class HelloWorldTest < Minitest::Test
+  def test_no_strange_name
+    assert_equal 'Hello, !', HelloWorld.hello(''), 'Greeting given an empty string as a name'
+  end
+
   def test_no_name
-    assert_equal 'Hello, world!', HelloWorld.hello(''), 'Greeting given no name'
+    assert_equal 'Hello, world!', HelloWorld.hello, 'Greeting given an empty string as a name'
```
kytrinyx: The test feedback in the case of a failure is a bit confusing here. Can we make it `no name`?

kotp: I agree, and paused on this.  But a blank name is not no name.  Nil would be no name.  Subtle but important distinction.

kytrinyx: The way I see it, this is a blank name: `hello("")`, whereas this is a nil name: `hello(nil)`, whereas this is no name: `hello`. Nil is sometimes nothing, sometimes it represents something. That's context-dependent, and we don't have context here, so it's hard to say whether the nil case actually means no name.

I think the failure messages should probably describe either what went wrong, but these are really used to identify which particular assertion failed. If we don't have trouble with context we should just delete the messages altogether.

kotp: I think my prior messages (wherever I submitted messages in tests) were trying to be descriptive of the reason the test failed.).  Having independent tests are important, that gives the behavior wanted.  The message, feedback on why it failed.

So the action points on this then are to have tests for default name (or no name) and a purposefully given blank name.  But no name given should be covered due to the 'world' being the default name?  Looking at the code again... 

With one failure, with changes, it can read like this:
```
HelloWorldTest#test_no_name [hello_world_test.rb:9]:
When given no name, the default should happen.
Expected: "Hello, world!"
  Actual: "Hello, !"
```
I will do a push. :)

kytrinyx: I really like this, thanks for helping clarify both the thinking and the result.

PR #131, opened 2015-05-16
Moves the 'file not found' and getting_started from hamming and updates the getting_started.md

### LINKED ISSUE #130: Should the 'file not found' message in hamming be included in hello_world instead?
It seems like it will be more useful in hello_world rather than hamming.
Here is the error message:
```ruby
__END__

*****************************************************
You got an error, which is exactly as it should be.
This is the first step in the Test-Driven Development
(TDD) process.

The most important part of the error is

      cannot load such file

It's looking for a file named hamming.rb that doesn't
exist yet.

To fix the error, create an empty file named hamming.rb
in the same directory as the hamming_test.rb file.

Then run the test again.

For more guidance as you work on this exercise, see
GETTING_STARTED.md.
*****************************************************
```

trayo:
The file `GETTING_STARTED.md` was moved from the hamming directory to the hello-world directory.
The `rescue from LoadError` was moved from `hamming_test.rb` to `hello_world_test.rb`.

I also updated the `GETTING_STARTED.md` to walk the user through how to solve the first test. I made the first test the one that has the user solve for `HelloWorld.hello` because is the simplest.
I can change it back to the empty string test if we want `HelloWorld.hello` to be extended functionality, but expecting to write the solution for `HelloWorld.hello('")` as the code `"Hello, !" is pretty confusing initially. I think it should start with either "Hello, World!" or "Hello, Alice!" in that case. I took the liberty of fixing a couple of sentences that sounded confusing when I was reading them.
Please give this a quick read, in case I missed anything.

In addition, I thought the tests could use more descriptive names.

closes #130

### DIFF
--- a/hello-world/GETTING_STARTED.md
+++ b/hello-world/GETTING_STARTED.md
@@ -1,0 +1,117 @@
+# Getting Started
+
+These exercises lean on Test-Driven Development (TDD), but they're not an
+exact match.
+
+The following steps assume that you are in the same directory as the test
+suite.
+
+You must have the `minitest` gem installed:
+
+    $ gem install minitest
+
+## Step 1
+
+Run the test suite. It's written using the Minitest framework, and can be
+run with ruby:
+
+    $ ruby hello_world_test.rb
+
+This will fail, complaining that there is no file called `hello_world`.
+
+To fix the error create an empty file called `hello_world.rb` in the same
+directory as the `hello_world_test.rb` file.
+
+## Step 2
+
+Run the test again. It will give you a new error.
+
+    1) Error:
+    HelloWorldTest#test_no_name:
+    NameError: uninitialized constant HelloWorldTest::HelloWorld
+      hello-world/hello_world_test.rb:5:in `test_no_name'
+
+Within the first test, we are referencing a constant named `HelloWorld` when
+we say `HelloWorld.hello`. When Ruby sees a capitalized name like
+`HelloWorld`, it looks it up in a big huge list of all the constants it knows about,
+to see what it points to. It could point to anything, and often in Ruby we have
+constants that point to definitions of classes or modules.
+
+When it looks `HelloWorld` up in it's list, it doesn't find anything, so we need
+to make one.
+
+There are several ways that this error message can be made to go
+away, one of which is to define a `HelloWorld` class.
+
+Open up the hello_world.rb file and add the following code:
+
+    class HelloWorld
+    end
+
+## Step 3
+
+Run the test again.
+
+    1) Error:
+    HelloWorldTest#test_no_name:
+    NoMethodError: undefined method `hello' for HelloWorld:Class
+        hello_world_test.rb:5:in `test_no_name'
+
+This time we have a `HelloWorld`, but we're trying tell it to `hello`, and
+`HelloWorld` doesn't understand that message.
+
+Open up hello_world.rb and add a method definition inside the class:
+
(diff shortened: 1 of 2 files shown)


kotp:
"In addition", indicates another pull request, really.  What if the new names are loved, but you missed something with the other?  Then you have tied two independent problems into one pull request and perhaps one commit (since there is only 1 commit in this pull request.)

That said, I like what you have done with the test names, and if it were another pull request and if it were *I* who had the ability to merge it in, that would already be done.

Unfortunately, that may be tied up while the PTB (powers that be)  decide all of it as a package is good to go.

trayo:
Oh yeah, my bad. I got a little carried away when I moved the tests around a bit. :smile_cat:

kytrinyx:
@kotp makes a good point. Would you separate this into separate pull requests, please? I'd like to see each atomic change in a separate commit. When looking at the history of the repo, there would be no commit that describes the changes to the test names, which could be confusing.

trayo:
I updated the commit and removed the different test names.

hamming/hamming_test.rb
```diff
@@ -1,11 +1,5 @@
 require 'minitest/autorun'
-begin
-  require_relative 'hamming'
-rescue LoadError => e
-  puts "\n\n#{e.backtrace.first} #{e.message}"
-  puts DATA.read
-  exit 1
-end
+require_relative 'hamming'
```
kytrinyx: Hamming gets generated from shared data. Take a look at the `example_gen.rb`. (I'm actually in the middle of refactoring this, so if you're OK with holding off until @kotp has a chance to review the generator PR (https://github.com/exercism/xruby/pull/134) then it might be easier to clean the new template up than the old `example_gen.rb` file.

kytrinyx:
This looks good. The only comment is the bit about hamming being generated.

kotp:
PR #134 is good to go, aside from a clarifying point about the version number in the solution file the programmer generates (or not).

trayo:
Sounds good! :+1:

I'm going to push a PR for the name changes in the test. Is it preferred that the PR be based off of this commit or off of master and fetch/rebase it later?

kotp:
It is up to you @trayo as @kytrinyx indicated in the comment.  It may be easier/simpler to make the changes from the new version, once it is merged, base the branch from updated master.  You will likely want to base your changes from the result of #134

kytrinyx:
Off of an up-to-date master is preferred

kytrinyx:
Wait--which test names are we talking about? If this is for hamming, then we would need to update the descriptions in the metadata (https://github.com/exercism/x-common/blob/master/hamming.json) since that's what the test names are generated from.

kotp:
Perhaps this is confusing due to effecting two different exercises.  Can we split this into two different pull requests?  It will help isolate things specifically to the beginning exercise of "HelloWorld" and the "Hamming" exercise.  Especially since there may be external code forces to deal with.  @kytrinyx peek at the 4 files and 2 directories affected in the commit?

kytrinyx:
Yeah, that might be easier.

If this were to simply add the files to hello-world, I could merge that immediately. Removing files from hamming is a bit tricker, and it's OK to leave the duplicated getting started stuff in hamming until we've resolved the generator piece in the other pull request.

trayo:
That's a good point! I went ahead and changed it so hamming stays the same.

kytrinyx:
Sweet! Thanks for being so patient with this whole process.

PR #132, opened 2015-05-17
Clean up 'food-chain'

kytrinyx:
- Normalize "expected" strings.
- Replace instance methods with class methods.
- Implement example as a module rather than a class.
- Rename module to FoodChain instead of FoodChainSong.
- Rename `sing` to `song` (noun for consistency with other methods).
- Remove trailing newline on last line of each verse (if I were to write the song down in a file, I wouldn't have a blank line at the very end of the song).
- Add a VERSION constant for book-keeping purposes.
- Replace logic with value in test for 'song'


kytrinyx:
@kotp I overhauled the entire problem. If you have any thoughts on this, I'd love to discuss the choices.

kotp:
Peeking.

kotp:
The above pull request is only a pull request against this branch, not master.

kytrinyx:
Thanks for helping think this through, @kotp. I think the result is a huge improvement to what we had earlier.

NobbZ:
The biggest problem with this approach I see, there are so many solutions out there, that just open the file and return its content. No one cares for `#verse` and `#verses` anymore, they all just do `#song`.

kytrinyx:
Interesting. So we should probably add tests back in for those, to enforce some sort of algorithmic approach.

kotp:
Verses is still being tested though, I will have to look at recent submissions for this exercism to note what @NobbZ is stating.  @NobbZ any idea on how to test to enforce a more dynamic solution?

PR #133, opened 2015-05-17
Clean up food chain - continued

kotp:
re #132

Continuing to normalize the strings and in so doing making the Heredoc
the One Pattern here.  It is a little bit of overkill maybe for the
first 2 or three tests, but soon makes sense to do this, May as well get
them all in the same pattern.


kytrinyx:
I deliberately did not use the heredoc for the individual verses, since I don't particularly like using heredocs in the example solution (which tends to be what happens because the easiest thing to do is to copy/paste from the tests). The place where it makes sense to stop copy/pasting is the multiple verses strung together.

I may just be over-thinking things. What do you think?

kotp:
I definitely had the same feeling, that the test can direct the solution.  For what I want to do was to make an array and test it against the array and clean up the tests, as they are not DRY... and since this is not a  normal project, it would be appropriate to do so.  Still, it is something that can bias the solution.

The solution to that, then, becomes having an external source for the song lines, and forming the expected result from that.  And using the library as the actual result.

The test becomes more cryptic, but the reporting from the test remains helpful, hopefully not biasing the solution.

kytrinyx:
I'd prefer it if the tests don't get clever. Ideally, I'd only want one test--the last one. The rest are really just training wheels.

What if we only have three tests:

1. a place-holder test that says: "you know how this works, go ahead and TDD it yourself"
2. the final test that shows where we're headed.
3. the book-keeping test (`VERSION`).

kotp:
Details in the commit message.  Let me know what you think.

Are we thinking that the self-drive TDD will eventually also be shared to be nitpicked?  Or a self-evaluate kind of thing?

Hoping you find the song text being in the external file is not any more clever (perhaps even less clever) than using the `\" character to escape the new line being used to continue a line of code, in the other versions of the test.

kotp:
One other thing, I did hear you about not lilking to use the heredoc solution in the example.rb.  It may be less 'clever' than using the `" \ <newline>"` syntax to continue the line.

kytrinyx:
I usually find HEREDOCs to be completely fine. I've avoided them with beginners because it seems to confuse and distract them. Probably by this point it won't be confusing.

I think the example looks good. I also think it's nice to read from song.txt. It's not clever, and it keeps the test file easier to read.

> Are we thinking that the self-drive TDD will eventually also be shared to be nitpicked? Or a self-evaluate kind of thing?

My thought is just self-evaluate.

I don't think we need more than the single test that reads from song.txt. At this point they've done enough problems that we can say "write whatever tests will help you solve this".

food-chain/food_chain_test.rb
```diff
@@ -1,166 +1,11 @@
...
-She swallowed the dog to catch the cat.
-She swallowed the cat to catch the bird.
-She swallowed the bird to catch the spider that wriggled and jiggled and tickled inside her.
-She swallowed the spider to catch the fly.
-I don't know why she swallowed the fly. Perhaps she'll die.
-
-I know an old lady who swallowed a horse.
-She's dead, of course!
-    SONG
+    expected = IO.read("song.txt")
```
kytrinyx: Try using `File.expand_path('../song.txt', __FILE__)`. `make test` runs from the root of the repository, and would be looking for it relative to there, rather than relative to the test file.

kytrinyx:
Oh. Hm. I wonder if it's failing because the song.txt doesn't get copied when `make test` puts it in a temp dir. Investigating.

kytrinyx:
Try adding this right after line 23 of the Makefile (`@echo "running tests for: $(ASSIGNMENT)"`).

```bash
@cp -r $(ASSIGNMENT)/ $(OUTDIR)
```

kytrinyx:
You can test it locally with `make test-assignment ASSIGNMENT=food-chain`

PR #134, opened 2015-05-18
Refactor hamming test generator

kytrinyx:
This uses an erb template (example.tt), and moves supporting code into a
new lib directory.

The actual generator lives in bin/generate-hamming

I also added a book-keeping test (`VERSION`) and a comment that lets us know which SHA the shared metadata was at when the problem was last generated.


kotp:
Code looks good.

It should be made clear that the point at which the version needs to be incremented.  We have 3 occurrences of this happening.  The `.version` file, the test and the solution file the user is writing.  Are there any that need to be manually done?  (The user solution file?)

Otherwise, it looks pretty solid from what I can see at the moment.

kytrinyx:
The .version doesn't need to be incremented manually, though if you regenerate the solution multiple times you'll want to git checkout the .version file. I'll add something more to the README around this.

The example solution does have to be incremented manually at the moment, but we could easily solve that with the generator script. I'll fix that.

kytrinyx:
I've updated the generator to also update the example.rb file.

kotp:
What about the coders file? (This is doable due to the required name being in the test itself in the require line?)  Or is this strictly  user responsibility?

kytrinyx:
The coder's file is the user's responsibility, in my opinion.

kotp:
This looks good to me, added some style changes. Gets a :shipit: vote from me.

kytrinyx:
Yeah, those style changes make sense. There's one thing that needs to be changed in the text template, I'll fix it tonight if you don't get to it before I do.

kotp:
Waiting on a :shipit:  from you or another member.  Things should now be persistent.

kytrinyx:
w00t

Issue #135, opened 2015-05-19
Missing test for Binary?


bmulvihill:
I noticed while working on this solution that the case where a binary string is only numbers, i.e. 123 isn't covered.  I noticed this while trying to determine if a string was valid I could pass the tests by doing something along the lines of:

```ruby
   def to_decimal
    return 0 if n.to_i == 0
    n.reverse.each_char.with_index.map { |char, index| char.to_i * 2**index }.reduce(:+)
  end
```

the test for string carrot123 returns 0, which is fine, but if i put 123 as a string my method would return a value,11 in this case.

kytrinyx:
Nice, yepp. That seems like a missed test case.

kotp:
@bmulvihill Do you have a test that covers this?  Would you be up for submitting a pull request?

bmulvihill:
Sure I'll submit one in a bit!

kotp:
:shipit:

PR #136, opened 2015-05-19
Add numeric string test

bmulvihill:
Base on https://github.com/exercism/xruby/issues/135

I noticed while working on this solution that the case where a binary string is only numbers, i.e. 123 isn't covered. I noticed this while trying to determine if a string was valid I could pass the tests by doing something along the lines of:
```ruby
   def to_decimal
    return 0 if n.to_i == 0
    n.reverse.each_char.with_index.map { |char, index| char.to_i * 2**index }.reduce(:+)
  end
```
the test for string carrot123 returns 0, which is fine, but if i put 123 as a string my method would return a value,11 in this case.

### DIFF
--- a/binary/binary_test.rb
+++ b/binary/binary_test.rb
@@ -36,8 +36,13 @@ class BinaryTest < Minitest::Test
     assert_equal 1128, Binary.new('10001101000').to_decimal
   end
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks!

PR #137, opened 2015-05-19
Removes the getting started and 'file not found' code from the hamming exercise

trayo:
This PR is for when the hamming exercise is ready to become the second exercise.

#130


kotp:
:+1:

kytrinyx:
Sweet!

Go ahead and rebase onto the most recent master, and then check out this section of the README to see if it makes sense:

https://github.com/exercism/xruby#generated-problems

trayo:
I went ahead and pushed the rebase. The example makes more sense now. Do I need to update this PR by editing the example generators instead?

kytrinyx:
If you edit the `example.tt` file in addition to what you have here, then we're all good.

trayo:
I updated the example.tt and removed the same code that was removed from hamming_test.rb

kotp:
:shipit: Looking good!  Thanks, @trayo for all the hard work!

kytrinyx:
Thanks so much @trayo!

Issue #74, opened 2015-01-13
Document how to get coloured output in tests in SETUP.md


JikkuJose:
I am using `minitest-reporters` gem to colour the test output, I think adding that would make it easier for many; any objections in submitting a pull request?

The only down side I see is that users needs to install the dependency; may have to update the instructions elsewhere.

kytrinyx:
Unfortunately colours don't work on Windows, it just outputs very ugly meta characters.

kotp:
It can output colors in Windows, but we could do a platform check to shut it off for Windows users.

And the characters that it outputs are ANSI characters, just to know what to search for, if you are curious.

kytrinyx:
Thanks @kotp -- Maybe the gem that didn't work on windows was pride. I'd be open to a pull request that adds something like that optionally -- meaning don't bail and don't complain if it triggers a `LoadError`.

If it works for Windows I don't see a reason to shut it off for Windows users. I'd love to have someone test that, though, to verify before we change 70 test suites :)

kotp:
I think ANSI support is an environment question so we shouldn't worry too much about it.

Since we have to open the test files anyway, would we just document that if you want colored tests, you can insert the line:

    require 'minitest/pride''

As we are using minitest already and minitest comes with pride automatically.  No gem required.

JikkuJose:
@kytrinyx @kotp Then let me see if I can test `minitest-reporters` in Windows; frankly have never worked in Ruby on Windows yet! It seems the best option may to just disable it in Windows?

**Off topic**: Also, isn't the usage `MiniTest::Unit::TestCaset` [deprecated](http://stackoverflow.com/a/17461423/2602869), what about changing to `Minitest::Test`?

kotp:
It won't be so much 'disabling it' on Windows, but making it an opt-in which is not hard to do, you simply install a gem or require the minitest/pride library.

There is already discussion on the syntax of the Minitest unit tests.

kytrinyx:
The discussion about minitest versions is here: https://github.com/exercism/xruby/pull/71

Carpetsmoker:
I also have the problem that many colours don't work very well on white backgrounds; I regularly get things like bright yellow text on a white background :-/ A light (white) background colour is the default for many terminal emulators (such as `xterm`, `gnome-terminal`, `xrvt`, etc).

I don't know what MiniTest's colours are, since I only used rspec before, but in the interest of "safe defaults that work for as many people as possible" it's something to consider.

[...]

PR #138, opened 2015-05-19
How to get color tests...

kotp:
This is the easiest first step to get color.  May have some requirements
for those on Windows.

Re: #74


kytrinyx:
Yeah, I remember removing this because pride wasn't compatible with Windows at the time. That might have changed.

kotp:
I think Minitest is no longer shipped with Ruby either.  So this pull request (title) may evolve to manage a couple of things I am investigating.

kytrinyx:
Interesting, thanks!

PR #140, opened 2015-05-20
Convenience library

kotp:
Use this to disable skip directives in the tests so we don't need to
remember to replace or delete code while we are evaluating.


kotp:
This still needs the update in the readme, if we can use this.  Around here it says to define the skip method to disable the skips in the test.

kytrinyx:
This is handy, thanks!

PR #142, opened 2015-05-24
Capitalized W in "Hello, World" .

skeskali:


kotp:
Will this change the underlying goal?  Does this drive a solution where the person that solves it will want to automatically capitalize the first letter of the given argument?  I, for one, am not excited to tell someone else how to capitalize, or not, their names.  Also, it is correct to reference, or even address, the world without a capitalized letter, as 'world' is not generally a proper noun in English.

That change would cause the example solution to fail, as well.

Let us know what your thoughts are on this.

skeskali:
It won't change the underlying goal. See the top post in this issues thread for the reason behind the change: https://github.com/exercism/x-common/issues/87

kotp:
There we go.  exercism/x-common#87 mentions that we also need to change the example.rb file to accommodate the change. I believe this is why we get a CI failed message.

You can push to the same branch as the commits here and this PR will automatically be updated.  It may be preferred to create a topic branch for this work though (on your repository).

kotp:
You are Awesome!  Thanks for the help!

skeskali:
I'm happy to help. :smiley:

PR #144, opened 2015-05-25
Exempt bin directory from style violations

kytrinyx:
I prefer to have commands use a hyphen rather than snake case, but rubocop complains
about things like generate-hamming.


kotp:
Works well.

PR #145, opened 2015-05-25
Fix style violations in hello world

kytrinyx:
The hello world test suite triggered "long line" violations.

This uses a heredoc to move the actual failure message below the assertion.
Another option would have been to assign a temporary variable on the line above the assertion.


kotp:
I am really getting to like the HereDoc for failure messages.  It never really struck me before.

PR #146, opened 2015-05-25
Fix style violation in nth prime

kytrinyx:
The error message is a string, but it was defined as an array. The style violation that it triggered was that the array should have a comma after the last element. Instead of fixing that, I decided to make the constant a string right off the bat.

### DIFF
--- a/nth-prime/nth_prime_test.rb
+++ b/nth-prime/nth_prime_test.rb
@@ -1,11 +1,12 @@
 require 'minitest/autorun'
 
 require 'prime'
-ERROR_MESSAGE = [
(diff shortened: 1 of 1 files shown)


kotp:
I noticed this in the robot-name exercise as well.  The HereDoc actually provides a nice indenting for the messages that are invoked when it fails as well, so it solves a couple of concerns.  

I like it.

kotp:
I am realizing that I like that...  it provides indenting in the case of failure, and so it solves more than one concern.

kytrinyx:
Cool, I like the indentation bit as well. That was unintentional, but a nice side effect.

PR #147, opened 2015-05-25
Fix style violations in robot-name

kytrinyx:
Rubocop was complaining about a long line. At first I changed
the failure message to be a heredoc, which triggered a "long method"
violation.

This feels a bit dirty, in that the method is legitimately long, but
perhaps it's not too bad to make the message a constant.

### DIFF
--- a/robot-name/robot_name_test.rb
+++ b/robot-name/robot_name_test.rb
@@ -7,10 +7,10 @@ class RobotTest < Minitest::Test
     Query methods should generally not change object state.
   MSG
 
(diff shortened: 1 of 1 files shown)


kotp:
The regular expression used can be named as well, which will take care of the other style issue.  It is used at least 2 times, so it is appropriate to name it anyway, right?

Perhaps as e03ede2 to continue the idea?

kytrinyx:
Nice! Yes, naming the regex is a definite improvement.

PR #148, opened 2015-05-25
Simplify twelve-days

kytrinyx:
This takes the same approach as food-chain, reducing the required API to a single
module-level or class-level method, and moving the lyrics into a static file.

It provides a single acceptance test, as well as a book-keeping test ("VERSION", so that people know which version of the test suite people are solving against).

Coincidentally, this fixes style violations ('long line') in the test suite.

### DIFF
--- a/twelve-days/example.rb
+++ b/twelve-days/example.rb
@@ -1,15 +1,17 @@
-class TwelveDaysSong
-  def sing
+module TwelveDays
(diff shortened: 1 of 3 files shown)


kotp:
:shipit: I was looking at those long lines.  I like this idea.

kytrinyx:
Sweet :) Shipping it.

PR #150, opened 2015-05-26
Rewrite gigasecond problem generator

kytrinyx:
This uses the new style of generator based on lib/generator.

Incidentally, it also fixes the last rubocop style violation.


kotp:
Was failing CI, added VERSION in `example.rb`.

kytrinyx:
Gah, I thought I tested, but I must have tested _before_ adding the book-keeping test. Thanks for fixing it.

kotp:
No worries, it was definitely tested afterward as well.  CI for the win...

kytrinyx:
OK, shipping this. I think this means we can add rubocop to the travis build.

PR #151, opened 2015-05-26
Require Minitest gem, w/ explanatory error message

kotp:
First Exercise introduces Minitest gem requirement.


kytrinyx:
The indentation in the `__END__` portion of the hello world exercise is inconsistent. Sometimes it indents two spaces, sometimes it doesn't: https://github.com/exercism/xruby/blob/ExplicitMinitestGemRequirement/hello-world/hello_world_test.rb#L46-L65

Issue #93, opened 2015-03-12
Make *_test.rb executable and add a hashbang


Carpetsmoker:
Is it possible to make the `*_test.rb` files executable by default and to add a hashbang in the form of `#!/usr/bin/env ruby`?

That way I can just type `./<Tab>`, which makes running tests ever so slightly easier ;-)

kotp:
Yes, it is.

But you are a good systems coder, you know this already.  In fact, I would be almost surprised if you did not already have a pull request ready.

If you want easy tests running though, I would suggest using guard or some other auto-testing mechanism, then you don't even need to type those 3 characters every time, and that makes running tests extremely easy.  (I think you already know this as well.)

Carpetsmoker:
@kotp  I am slightly confused by your comment; do you mean "go make a patch!" or "meh; it's not that useful"?

kotp:
Yes, the comment was kind of two sided.  If you have a patch ready, I would be interested in seeing it.

But I also think that solving the problem is a great exercise left to each person to solve as they learn more about the language.

kotp:
@Carpetsmoker Any thoughts on what you want to do with this?  If a solution will benefit those on  [Li|U]n[u|i]x as well as Windows that would be awesome as well.

kotp:
Closed by #155

kytrinyx:
I'd like to revisit this.

Back when it was proposed I said I saw no reason not to include it, however I've later been convinced that it's not really great to give people the habit of executing something they downloaded without checking it first.

There are two parts to this:

1. When downloading the tests, the CLI will not set the executable bit, so people would have to chmod the file themselves.
2. Should we remove the shebangs and encourage people to add it themselves if they want to (or use guard)?

Carpetsmoker:
It seems to me that it's merely a convenience matter of typing `ruby test_foo.rb` vs. `./test_foo.rb`. I prefer the latter, as it's not uncommon for me to type `python foo.rb` or `sh foo.py` and it works well with tab-completion (You just have to type `./<Tab>`, instead of `ruby test_<Tab>`).

The issue of trusting random files "downloaded from the internet" is not really related to this, IMHO...

> Should we remove the shebangs and encourage people to add it themselves if they want to (or use guard)?

The issue is that you need to add it to a *lot* of files. I never played with Guard, but isn't that a Ruby-only thing? At any rate, I just would *prefer* to type `./foo.rb`; it's nice and simple and doesn't require any setup.

[...]

PR #155, opened 2015-05-29
Make test executable

kotp:
Changes execution bit and adds magic string

Re: exercism/xruby#93


kytrinyx:
Cool, yeah I see no reason not to do this.

PR #156, opened 2015-05-30
Clarify dependency or error.

kotp:
For Minitest less than version 5.0.0 the error states "Missing
Dependency" and for the first error, it states "Error" as eluded to in
the message.


kytrinyx:
Subtle and useful. I like it.

PR #157, opened 2015-05-30
93 make tests executable

kotp:
Actually changes the mode of the test files to executable bit, adds some tools to bin, and includes a test for each test file to check if it is in fact executable.

re #93 and #155


kytrinyx:
Yeah, I think that a tool for this is a good approach, especially seeing as we have other tools already.

Maybe we can add a make task that runs all the various tools to make it easy/quick to check everything before committing.

kotp:
A git commit hook?

kotp:
Thanks @kytrinyx I was hoping you liked it.

kytrinyx:
I would definitely make it a git commit hook. I always forget, otherwise.

kotp:
See #159

Issue #158, opened 2015-05-30
Some files are skipping tests in Travis


kotp:
`custom-set` and `ocr-numbers` both have tests that are being skipped in Travis

kotp:
Currently these two problem sets, if the skips are disabled, fail.  Unsure why the example file does not address these tests, as I thought these were used as the aluminum standard<sup>[1](#aluminum_standard)</sup> and required for inclusion.

* * * *
<a name="aluminum_standard">1</a>: Aluminum Standard - In programming, the code which demonstrates an idea, perhaps given as an example of usage, but not up to the 'gold standard' which could be brought on by refactoring or redesign.

kotp:
@kytrinyx When you review this, understand that if you pull these in, we will likely have some long standing Travis breakage reports... we should hold off until ocr-number is either suspended or corrected.

Contact the author for working `example.rb`?

kytrinyx:
Yeah, I'd like to avoid having long-standing CI failures. I think I originally made ocr-number, but I bet the test suite has changed a few times over the past few years. I'll take a look.

kytrinyx:
The problem with OCR is that someone's editor at some point removed trailing whitespace, which is crucial to actually solving the problem.

kotp:
Thansk @kytrinyx everything is cleaned up and working well locally and with Travis.

PR #160, opened 2015-05-30
Allow tests to reveal broken problem examples

kotp:
re: #158

This first step is to reveal the skips that are not disabled allowing
bad example code through our checks.

To allow for this, the monkey patch on Skip should adhere to the
contract set forth in `Minitest::Assertions`

The Makefile is now using the same process for disabling skips.


kotp:
As is wanted, Travis is failing.

Issue #159, opened 2015-05-30
Git hook to pre-evaluate before Travis


kotp:
This will help track authors with preparing as expected to pass CI check.

kytrinyx:
Git hooks aren't distributed with the repo, but we could certainly have a little script that initializes a git hook with the correct behavior to make it very, very easy to set up.

kotp:
We can have a tool that sets it up, no problem at all.  It would be distributed in bin, executable, versionable, and just do the right thing.  On a *n?x system, it can be a soft link but on Windows would have to be a copy maneuver.

The `Makefile` can even potentially be used for this, if we wanted to go that way rather than a tool in `bin` but I think the bin folder is the way to go.

The question that we need to answer is if we want it to be pre-push or post-commit or pre-commit.  I am leaning toward pre-push hook, myself.

Locally, I have simply been watching for file system events in the folder and running `make` when the file system changes.  Easy enough for continuous testing.  I think doing the checks per commit are just a little bit too soon, and a pre-push will give a 'last chance' stage.

kotp:
Current solution is a copy, should work on Windows.  I prefer a soft link, but it may not work well with Windows.

The difference may only affect how  maintenance of the hook is done.

kytrinyx:
> I think the bin folder is the way to go.

Yeah, I agree.

> The question that we need to answer is if we want it to be pre-push or post-commit or pre-commit.

I think pre-push is fine, so long as people are happy to rebase/squash/edit etc. I was initially leaning towards pre-commit, but the folks who will flail with their commits (who I think pre-commit would help) would probably not set up the hook in the first place, so we'd end up helping them squash etc in the PR anyway.

> Current solution is a copy

I think that's fine.

kotp:
Ready to fall into place for anyone that wants to use it.

kytrinyx:
Nice!

PR #161, opened 2015-05-31
Pre-Push and local check setup.

kotp:
They are one and the same file, but used automatically or manually.

This does not link the file... but does a physical copy to `.git/hooks`.

solves #159

### DIFF
--- a/bin/local-status-check
+++ b/bin/local-status-check
@@ -1,0 +1,21 @@
+#!/usr/bin/env ruby
+
+puts "Starting #{__FILE__} happening now."
+
+commands_to_run = [
+  'make test',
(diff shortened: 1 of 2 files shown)


kotp:
Next step is to update the README for developer setup.  Issue number:  #56

bin/local-status-check
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env ruby
+
+puts "Starting #{__FILE__} happening now."
```
kytrinyx: For consistency with the output at the end, perhaps this should read `"Started #{__FILE__} checks..."`

PR #163, opened 2015-06-01
Add trailing whitespace back to ocr-numbers

kytrinyx:
The trailing whitespace is part of the spec for the numbers. Each number is defined as a grid.

Addresses (perhaps only in part) #158


kotp:
I knew it had to be passing when it came in...  Thanks for fixing it.  It was on my todo list for Monday.

PR #167, opened 2015-06-03
Clean up temporary files created for tests

### LINKED ISSUE #164: Local checks not cleaning up their temporary files
This presumably happens on the Travis server as well, but may be automatically cleaned due to a reboot (just like on any Unix machine, if it honors traditional behavior).

However, like many traditional *nix machines, my machine can store many of these files, as the likelihood of a reboot happening for many days, perhaps many months, is slim.  We should be good citizens and remove the temporary files we create, as they are meant to be temporary.

kotp:
fixes #164


kytrinyx:
Nice, thanks!

PR #169, opened 2015-06-14
Add skip to hamming version test

kytrinyx:
The book-keeping test in hamming was missing a skip, which meant that one of the first
errors people got was the NameError for Hamming::VERSION. That seems a bit excessive.

It would be better to tack that on as the last thing.


kotp:
Looks good from here!

Issue #10, opened 2014-03-16
largest-series-product forces implementation


markijbema:
The tests for this exercise test a lot of the helper methods. I think given that it's this late in the path, it isn't needed. Also, it takes away a lot of freedom from the implementer. I think the tests should be guiding more by functionality needed than by methods defined.

kotp:
Related information: 

* [largest series exercise test and soltuion](https://github.com/exercism/xruby/tree/master/largest-series-product)
* [Meta for exercise](https://github.com/exercism/x-common/blob/master/largest-series-product.yml)
* [Readme data](https://github.com/exercism/x-common/blob/master/largest-series-product.md)

Listed here for my convenience.

kotp:
@markijbema and @kytrinyx hopefully this is what you had in mind.

kotp:
BTW: in the spirit of "Look to exercism for awesome code" I took the following measurements:

```
Flogging your code!
    44.7: flog total
     5.6: flog/method average

    11.9: Series#separate                  example.rb:30
     7.1: Series#largest_product           example.rb:9
     6.7: Series#collection_of_digits      example.rb:36
     6.4: Series#analyzer                  example.rb:17
Your code reeks this much!
example.rb -- 2 warnings:
  [19]:Series#analyzer contains iterators nested 2 deep (NestedIterators)
  [26]:Series#reduce_to_product doesn't depend on instance state (UtilityFunction)
```

kotp:
```
   42.6: flog total                                     
    4.3: flog/method average                            
                                                        
    6.7: Series#collection_of_digits      example.rb:45 
    6.5: Series#analyzer                  example.rb:22 
    6.4: Series#validate_length           example.rb:17 
    6.1: Series#separate                  example.rb:39 
```

kytrinyx:
That's a much nicer result.

PR #170, opened 2015-06-16
No longer drives the solution

kotp:
Does not rely on on any specifically named 'helper' methods.  Tests
things through the implementation via public facing API

re: #10 

I removed the test for "identity" as well, as there was nothing that indicated that a empty string should return 1, which is what the test stated.

The example file was changed against the original specification provided by the README and the tests prior to these changes.

### DIFF
--- a/largest-series-product/example.rb
+++ b/largest-series-product/example.rb
@@ -9,7 +9,9 @@ class Series
   def largest_product(length)
     @length = length
     validate_length
-    analyzer
+    return 1 if @digits.empty?
+    collection_of_digits
+    select_max { reduce_to_product  { validate  { separate } } }
   end
 
   private
@@ -17,11 +19,6 @@ class Series
   def validate_length
     @length > digits.length and
       fail(ArgumentError.new 'Not enough digits')
-  end
-
-  def analyzer
-    collection_of_digits
-    select_max { reduce_to_product { validate { separate } } }
   end
 
   def validate
--- a/largest-series-product/example.rb
+++ b/largest-series-product/example.rb
@@ -3,34 +3,37 @@
 class Series
   attr_reader :digits
   def initialize(numeric_string)
-    @digits = convert_to_digits(numeric_string)
+    @digits = numeric_string
   end
 
   def largest_product(length)
-    fail ArgumentError.new('Not enough digits') if length > digits.length
-    products = []
-    slices(length).each do |slice|
-      products << slice.inject(1) do |product, n|
-        product * n
-      end
-    end
-    products.sort.last
-  end
-
-  def slices(length)
-    result = []
-    i = -1
-    begin
-      i += 1
-      i2 = i + length - 1
-      result << digits[i..i2]
-    end while i2 < digits.size - 1
-    result
+    @length = length
+    fail ArgumentError.new('Not enough digits') if @length > digits.length
+    analyzer
   end
 
   private
 
-  def convert_to_digits(s)
-    s.chars.to_a.map(&:to_i)
+  def analyzer
+    collection_of_digits
+    reduce_to_product { validate { separate } }.max
+  end
+
+  def validate
+    yield.take_while { |array| array.size == @length }
+  end
+
+  def reduce_to_product
+    yield.map { |array| array.inject(1, :*) }
+  end
+
+  def separate
+    0.upto(digits.length - 1).map.with_index do |_, index|
+      digits[index, @length]
+    end
+  end
+
+  def collection_of_digits
+    @digits = digits.chars.map(&:to_i)
   end
 end
--- a/largest-series-product/example.rb
+++ b/largest-series-product/example.rb
@@ -8,15 +8,20 @@ class Series
 
   def largest_product(length)
     @length = length
-    fail ArgumentError.new('Not enough digits') if @length > digits.length
+    validate_length
     analyzer
   end
 
   private
 
+  def validate_length
+    @length > digits.length and
+      fail(ArgumentError.new 'Not enough digits')
+  end
+
   def analyzer
     collection_of_digits
-    reduce_to_product { validate { separate } }.max
+    select_max { reduce_to_product { validate { separate } } }
   end
 
   def validate
@@ -27,8 +32,12 @@ class Series
     yield.map { |array| array.inject(1, :*) }
   end
 
+  def select_max
+    yield.max
+  end
+
   def separate
-    0.upto(digits.length - 1).map.with_index do |_, index|
+    digits.map.with_index do |_, index|
       digits[index, @length]
     end
   end
--- a/largest-series-product/largest_series_product_test.rb
+++ b/largest-series-product/largest_series_product_test.rb
@@ -7,18 +7,7 @@ require_relative 'largest_series_product'
 # rubocop:disable Lint/ParenthesesAsGroupedExpression
 #
 class Seriestest < Minitest::Test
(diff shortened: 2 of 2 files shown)


largest-series-product/largest_series_product_test.rb
```diff
@@ -7,56 +7,8 @@
...
-    assert_equal expected, series.slices(3)
-  end
-
-  def test_other_slices_of_3
-    skip
-    series = Series.new('982347')
-    expected = [[9, 8, 2], [8, 2, 3], [2, 3, 4], [3, 4, 7]]
-    assert_equal expected, series.slices(3)
+  def test_required_method_called_largest_product
+    assert_respond_to Series.allocate, :largest_product
```
kytrinyx: What does this test mean?

kotp: It is the one test that specifies the signature that Sieve will have an instance method called `largest_product`.  The signature of the test itself is `object, method_name, message`.
```
(from gem minitest-5.7.0)                                                      
=== Implementation from Minitest::Assertions                                   
------------------------------------------------------------------------------ 
  assert_respond_to(obj, meth, msg = nil)                                      
                                                                               
------------------------------------------------------------------------------ 
                                                                               
Fails unless obj responds to meth.                                             
```

The documentation via RDoc demonstrates it well.

```
==== (Uses superclass method RDoc::AnyMethod: MiniTest::Assertions#assert_respond_to) 
------------------------------------------------------------------------------        
                                                                                      
Tests if the given Object responds to method.                                         
                                                                                      
An optional failure message may be provided as the final argument.                    
                                                                                      
  assert_respond_to("hello", :reverse)  #Succeeds                                     
  assert_respond_to("hello", :does_not_exist)  #Fails                                 
```

I could have allocated the object separately, and used a variable, but it is succinct enough here, or I could have instantiated it, knowing the signature we want, but it would then fail with an argument error for initialize, and I wasn't sure we wanted to start driving the initialize method here.  Though of course we could.  If the important thing is the method itself, better to have it only address exactly what we want to address.

It is perhaps not the first test that should be ran, though.  Perhaps it should be the tests that would drive the creation of the `initialize` method.

kytrinyx: Thanks for walking me through that!

There's another test that makes an assertion about the result of calling `largest_product` on the instance. Do we need this test in addition to the other one?

kotp: I think it depends on the purpose of the test.  If it is to get someone started with prerequisites, with that being the name of the method itself, this addresses that.  If it is an aside due to testing the method outright, then it accomplishes the same thing, by side-effect.

It really goes to how much the tests help drive the specification and communication, more than duplication.

Line 67, for example, accomplishes the same thing, in effect, but concentrates that the method operates as expected, not that it exists.  Of course, if it doesn't exist, it will also drive that eventuality. 

Cleaner, perhaps, to state that requirement of existence before how it will operate.  For sure, in a TDD manner, it would be appropriate to have some test state that a method exists, before testing for functionality.

Of course, I am up for suggestions.

I have other changes to make as well, just waiting for feedback on this, then I can make the update.  So this isn't quite ready to go yet... I will commit to this branch other changes that are pending currently, as well as any test changes you see fit.

kytrinyx: It hadn't ever occurred to me to specify the API of a library explicitly, separately from the behavior. I'm still not convinced about it, but I'm open to the possibility that there's some benefit that I'm missing.

At this point I'd rather keep the tests about the behavior, with the API specified as a side effect.

kotp: I will make the change.  Thanks.

kytrinyx:
> I removed the test for "identity" as well, as there was nothing that indicated that a empty string should return 1, which is what the test stated.

I think that the largest product of nothing should be the identity. I'm not sure why you removed it. Because this isn't in the README?

kotp:
Yeah, that was the only reason, there was nothing to indicate what it should be other than the test.  The largest product of nothing should be 0 perhaps?  I don't know what the set theory is, or if it applies, or just the arithmetic theory.

largest-series-product/example.rb
```diff
@@ -3,34 +3,46 @@
...
-    result
+  private
+
+  def validate_length
+    @length > digits.length and
+      fail(ArgumentError.new 'Not enough digits')
   end
 
-  private
+  def analyzer
```
markijbema: little naming nitpick: i think it's clearer when a method which returns something is named after the thing it returns, so maybe 'maximum' or if you want to show that the analysis is the important thing here I would say something like analysis_result. When you have a method called analyzer I would expect it to return a lambda (or object responding to call)

largest-series-product/largest_series_product_test.rb
```diff
@@ -77,6 +18,12 @@ def test_another_tiny_number
     assert_equal 9, series.largest_product(2)
   end
 
+  def test_largest_product_of_2
+    skip
+    series = Series.new('0123456789')
+    assert_equal 72, series.largest_product(2)
```
markijbema: i think it might be better to write 9*8 here to provide a little hint

markijbema: and maybe for the 270 as well

kotp:
I was considering adding an actual message to clarify, what do you think @markijbema ??

markijbema:
I like this way better. One thing which immediately strikes me is that I would implement this differently now. I would just sort the digits, take the last few, and multiply them. A solution of which I didn't think before (at least).

markijbema:
Adding an explicit message would even be better, agreed.

kotp:
Can't sort the digits, as the product group changes when you do.  I had initially mis-read the information given, and was doing that.

markijbema:
ah yes, you're right. Forgot the exercise, but it says series of digits. I think given the tests, it would still be my first attempt for the third test ;) I'm really liking this!

kotp:
Pair with me real quick?

markijbema:
Sorry, I'm at work (but always start the day by github notifications of both work and outside of it), so don't have the time. Btw, Katrina is right about the identity, it should be 0. If you're interested, it's indeed group(/field) theory. Over addition, 0 should be the identity, over multiplication it should be 1.

largest-series-product/example.rb
kotp: Agreed, naming is hard.  But analyzer_result is a duplication, as result is assumed, probably, as it is the return of the method.

Maybe largest_value or even _largest_product to avoid the duplication of the method name itself?

largest-series-product/largest_series_product_test.rb
kotp: Message are better, as mentioned... add the largest series as a hint?

kotp:
It was 1. so I will restore that, but should there be a mention of "why" in the README or a reference somewhere?

kotp:
Thanks @markijbema for clarifications.  Added messages and restored the identity test.  Also refactored method I think it is better.

markijbema:
I think this is good to merge. I don't feel to strong either way over the explanation of the identity, but then again, I did mathematics ;)

@kytrinyx also ok with merging this? Do you want explanation of the identity?

kytrinyx:
I think this is good like this. Identity can be a bit of a rabbit hole.

kotp:
I was just getting ready to squash, when I was checking the state of master and was surprised by the traffic coming in.  Nice surprise.

Thanks @kytrinyx !

PR #171, opened 2015-06-20
binary: Add error handling

kytrinyx:
This adds some additional test cases to clarify the behavior around
the conversion of valid/invalid binary input.

See https://github.com/exercism/x-common/issues/95

### DIFF
--- a/binary/binary_test.rb
+++ b/binary/binary_test.rb
@@ -4,6 +4,10 @@ require 'minitest/autorun'
 require_relative 'binary'
 
 class BinaryTest < Minitest::Test
+  def test_binary_0_is_decimal_0
+    assert_equal 0, Binary.new('0').to_decimal
+  end
+
   def test_binary_1_is_decimal_1
     assert_equal 1, Binary.new('1').to_decimal
   end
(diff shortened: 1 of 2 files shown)


kotp:
The example does not adhere to the "do the conversion yourself" but it is a sanity check that everything is working as it should be... trusting the implementers of the language more than a 'self made' solution.  The educational value is not in the example.rb file itself, so I think this is OK to do.

kytrinyx:
> The example does not adhere to the "do the conversion yourself" but it is a sanity check that everything is working as it should be... trusting the implementers of the language more than a 'self made' solution.

I'm not sure what you mean. The example does the conversion manually.

kotp:
Yes, but it uses a built in method that does this conveniently, rather than writing the implementation explicitly.  Using to_s(base) is probably not really the goal, based on the README.  But it is definitely good as the definitive example that is expected in the way that it is to work.

kotp:
Specifically, my changes to `example.rb` as:

```
  def to_decimal
    digits.to_i(2)
  end
```

kytrinyx:
The point is to _not_ use `to_s(base)`, though.

kytrinyx:
Oh, I didn't see your changes. That makes sense. I don't want to use digits.to_i(2) since the whole point of the exercise is to understand binary conversion enough to do it manually.

kotp:
Yes, but the nitpickers don't see the example, right?  It is a different view point?

kotp:
if I do the 'thing to measure against' in the tests, then it sort of exposes this.  If I leave it in the more 'hidden' example.rb file, then it does what is intended, cleanly.

kytrinyx:
Right now the example solution is only used to test the test suite. I guess that means that it's OK to use a solution that would be discouraged on the site, but it seems wrong to me in a way that I'm unable to articulate.

kotp:
I have the same little nag as well.  I think it is because of the conflicting point of view.  It does potentially avoid bugs that a 'homegrown' solution could introduce though.

kytrinyx:
I see how it could avoid the bugs that a homegrown solution might have, but this is really a trivial problem, and the homegrown solution shouldn't have bugs.

kotp:
Then just the style changes then...

kytrinyx:
Okay :)

PR #172, opened 2015-06-29
change ranges to use string versions

dalexj:
it's a little bit more concise to read without have `String(i)` and you see the `'quotation marks'` instead


kytrinyx:
I like it, thanks!

PR #173, opened 2015-07-03
explicit reference to error on Windows

abartov:
when running the test on an empty file on Windows (Ruby 2.2), there is an error that was not mentioned in this very first help text.  It is important to acknowledge it and reassure the exercist, so that they don't give up.


kytrinyx:
Thanks, I didn't realize the error was different on Windows. This is a great catch.

Issue #176, opened 2015-08-02
Distinguishing between quotations and contractions in Word Count


Calvyn82:
The final spec test in Word Count looks for the use of apostrophes, which is a great idea that got me thinking.

The quiz could have one more test that included the use of apostrophes inside of contractions, but also had words in single quotes.

This would make for an interesting level of difficulty at the end of the quiz, requiring users to write something that checked how the punctuation was being used. If this seems like a good idea, I'm willing to write the pull request for it, but I didn't want to proceed unless it was something desired by others.

If creating a breaking test in a quiz that is in production isn't something that is acceptable, I understand.

kytrinyx:
We've been very easy-going about breaking changes in the test suites. We've started introducing the idea of a [version constant for book-keeping purposes](https://github.com/exercism/xruby/blob/master/hamming/hamming_test.rb#L79-L84), which lets us avoid the "your code doesn't pass the test" discussion in the nitpicks.

I think the idea is interesting, and I'd be curious to hear what others think. @kotp -- thoughts?

One thing that we might notice is that if the exercise becomes more challenging, we may want to move it slightly further along in the track. I don't think that this change would [change the classification of the exercise](https://github.com/exercism/x-common/issues/63).

kotp:
I would be interested in seeing at least a description of the test that is being considered here.   A pull request whether accepted, considered for a new exercise, or not accepted is no where near a waste of time, in my opinion.  It clarifies what the ideas are.

kotp:
I am still trying to figure out how 'parenthesis' falls into this, from the title?  @Calvyn82 any ideas to show a working test?

Calvyn82:
I'll write a pull request up when I get off tonight, but the test sting might be something like:

"Joe can't tell between 'large' and large."

Or

"He said, 'That isn't what he said.'"

Calvyn82:
I see your confusion now. It's because I'm an idiot sometimes and I said parenthesis in the title when I meant quotations.

kotp:
Fixed the title.

kytrinyx:
Closed by #117

Issue #92, opened 2015-03-03
for...in loop nitpicks


KaoruDev:
Not sure if this is the right place to make this issue but I'm see a lot of automated nitpicks with:

> Rubyists tend to prefer enumerable methods over for loops. This isn't
> a rule, it's more of a strong cultural preference.

Which isn't a 100% true, there is a reason why most don't use `for...in` and it's because it does not create a closure. i.e.
```
x = 0
for x in (0..4)
end
x #=> 4
```

Which can cause some weird bugs.

kotp:
So how would you suggest changing the wording?  It maybe able to be improved.

I would say that most do not use it because they aren't solving problems where the last iterated value is not needed to be kept.  Or because of the same reason that people don't tend to use 'or' and 'and' as opposed to `||` and `&&` especially early on in their development as Rubyists.

I have noticed that as people become more experienced with Ruby they start to use each of these things as intended and where appropriate, because they do see the features and nuances given by each.

And though it can be the source of some weird bugs, I don't think it is the cause of those bugs.  The cause is simply not keeping track of your variables.

KaoruDev:
ah yes i forgot to put my suggestion!

Perhaps pointing out the behavior of `for..in` along with your comment: 

>  And though it can be the source of some weird bugs, I don't think it is the cause of those bugs. The
>  cause is simply not keeping track of your variables.

would suffice. I suppose another good way is to make a concrete example. i.e.

```
x = 0
for x in 0..4
end
x #=> 4
```
vs.
```
x = 0
(0..4).each do |x|
end
x #=> 0
```

I argue the reason we don't use `for..in` is it can change the state of our variables without us realizing it. It's not so obvious.

kytrinyx:
This is a really great distinction. The comment lives here: https://github.com/exercism/rikki/blob/master/comments/ruby/for_loop/for_loop.md

Do you want to take a stab at improving it?

KaoruDev:
ah you found my kryptonite: English! haha I'll take a stab sometime today, thanks for the link =)

kytrinyx:
LOL, I'm happy to tweak once you've given it a first pass :)

KaoruDev:
Closing, address in: https://github.com/exercism/rikki/pull/8

kytrinyx:
Yepp, thanks! Also: deployed.

PR #177, opened 2015-08-04
Wrote test for word count.

Calvyn82:
Here is an example of what I had in mind.


kotp:
I think it needs the code as to pass the tests as well.  This will give us an idea if it shifts the target problem enough to create a new exercism problem. 

re #176  this does not address the parenthesis part of your issue title.  Perhaps the title is wrong?

Calvyn82:
It is. I don't even have the excuse of English not being my first language. So sorry.

Calvyn82:
Did you mean that I needed to update the example code to make it pass the test as well?

If so, I updated it. The example now passes all of the tests.

word-count/word_count_test.rb
```diff
@@ -82,4 +82,14 @@ def test_with_apostrophes
...
   end
+
+  def test_with_quotations
+    skip
+    phrase = Phrase.new("Joe can't tell between 'large' and large.")
+    counts = {
+      'joe' => 1, "can't" => 1, 'tell' => 1,
+      'between' => 1, 'large' => 2, 'and' => 1
+    }
+    assert_equel counts, phrase.word_count
```
kotp: assert_equal is misspelled so Travis is failing.

Calvyn82:
Thought I had that fixed. Must have reverted it when I put the skips back in. Should work now.

kotp:
Looks good to me, @kytrinyx you think the wording should change in the exercise description?  It is now looking for only 'proper' words exclusive of punctuation now.

I think it is close enough to the original intent of the exercise.  We aren't trying to do sentence count, etc, nor looking at hyphenated words.  This does address a pretty simple idea of what we would look for in a word count program, I think.

kotp:
Version information should be added in though.

Calvyn82:
I'm getting the x-common stuff setup and looked at right now for the versioning.

kytrinyx:
> I think it is close enough to the original intent of the exercise.

Yeah, agreed.


About the versioning--why did you remove it? If you add the constant to the example solution, it should pass the tests.

Calvyn82:
I'm open to suggestions about what I'm doing wrong, here.

kotp:
I am not sure that the `VERSION` constant was there yet, @kytrinyx .  So not sure that it was removed.

@Calvyn82 Can you look at another exercise set to see an example of how the VERSION constant works?

Calvyn82:
So, In line 7 of hamming/hamming_test.rb it has a commented out merge statement.
# 11a7e6f Merge pull request #92 from kotp/116_Verbal_Word_Problem

 Is that something Travis needs for word-count/word_count_test.rb? If so, what pull request should I match it up with?

kotp:
The current (master) branch is what you should compare with.  That comment is just a comment.  It is not read by anything to produce anything.

kytrinyx:
The comment is used by the generator to show what version of the shared data it's using. You don't need to pay attention to that.

kytrinyx:
The failing build is unrelated to this--it's updates to the rubocop gem (and we haven't pinned it). Once this is squashed, it's safe to merge. I'll go fix the rubocop stuff in a separate branch.

PR #178, opened 2015-08-08
Pin rubocop to a known good version

kytrinyx:
There were some backwards incompatible changes that caused new offenses to be triggered
on CI, which in turn causes PRs to fail for reasons unrelated to the code they're changing.

I was going to fix the new offenses, but one of them is actually triggered by
a bug in the rubocop library (see https://github.com/bbatsov/rubocop/issues/2116), so this turned
out to require more time than I could spare for it today.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ rvm:
   - 2.1.2
 
 install:
-  - gem install minitest
-  - gem install rubocop
+  - gem install minitest --no-ri --no-rdoc
+  - gem install rubocop -v 0.31.0 --no-ri --no-rdoc
 
 script:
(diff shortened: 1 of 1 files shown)


kytrinyx:
I'm merging this, since it passed CI, and will unblock us on merging https://github.com/exercism/xruby/pull/177

kotp:
When I get a dedicated computer again, I will look at this further, @kytrinyx

kytrinyx:
Thanks. I'll push up the fix for most of the warnings. There's no news on the rubocop issue just yet. It seems like their implementation for that cop is pretty brittle, and they might need to rethink it. I didn't want to jump in and add a hack.

PR #180, opened 2015-08-09
Remove unnecessary rubocop disablings

kytrinyx:
These are some of the warnings that come up if you use a version of rubocop that is higher than 0.31.0.


kotp:
Looks good to me, glad that some of these are gone.  But I need to review the recent changes for Robocop to see how they have changed.  What version are we pinned to currenty?

kytrinyx:
We pinned to 0.31.0: https://github.com/exercism/xruby/blob/master/.travis.yml#L10

Issue #179, opened 2015-08-09
README for food-chain is way too vague


Rojo:
Most persons are just loading the included `song.txt` file and returning its content to make the test pass. Is it the intended way to work the exercise or should people generate the song programatically?

If the later is the intended task, I think it should be stated clearly in the README and _test file_ for the exercise. Also, displaying a warning message when `File.open` or `IO.read` are used would be nice.

kytrinyx:
Yeah, the idea was to generate the song programmatically. I think we should rethink the test-suite and also (as you suggest) take another look at improving the README.

Rojo:
Addressed on https://github.com/exercism/xruby/pull/181 and https://github.com/exercism/xruby/pull/184.

PR #181, opened 2015-08-14
food_chain: An error is raised if File.open or IO.read are used.

Rojo:
This would be the first part to address https://github.com/exercism/xruby/issues/179.

As a lot of people solves the _kata_ by just loading the file `song.txt` with either of `File.open` or `IO.read`, this version of the tests raises run-time error, explaining that the desired solution must be to generate the song verses with the program.

It is left to improve the wording in the [metadata](https://github.com/exercism/x-common/blob/master/food-chain.yml).

### DIFF
--- a/food-chain/food_chain_test.rb
+++ b/food-chain/food_chain_test.rb
@@ -4,6 +4,36 @@ require 'minitest/autorun'
 
 require_relative 'food_chain'
 
+module NoCheating
+  ERROR_MESSAGE = <<-MSG
+  This exercise intends to help you improve your hability to work
+  with data generated from your code, so DON'T just open or read
+  the song.txt file.
+  MSG
+
+  class File
+    def self.open(*)
+      fail ERROR_MESSAGE
+    end
+  end
+
+  class IO
+    def self.read(*)
+      fail ERROR_MESSAGE
+    end
+  end
+end
+
+if FoodChain.class.eql? Class
+  class FoodChain
+    prepend NoCheating
+  end
+else
+  module FoodChain
+    prepend NoCheating
+  end
+end
+
 class FoodChainTest < Minitest::Test
   # This test is an acceptance test.
   #
@@ -23,6 +53,6 @@ class FoodChainTest < Minitest::Test
   # giving feedback know which version of the exercise you solved.
   def test_version
     skip
-    assert_equal 1, FoodChain::VERSION
+    assert_equal 2, FoodChain::VERSION
   end
 end
--- a/food-chain/food_chain_test.rb
+++ b/food-chain/food_chain_test.rb
@@ -4,27 +4,14 @@ require 'minitest/autorun'
 
 require_relative 'food_chain'
 
-module NoCheating
-  ERROR_MESSAGE = <<-MSG
-  This exercise intends to help you improve your ability to work
-  with data generated from your code, so DON'T just open or read
-  the song.txt file.
-  MSG
-
-  class File
-    def self.open(*)
-      fail ERROR_MESSAGE
-    end
-  end
-
-  class IO
-    def self.read(*)
-      fail ERROR_MESSAGE
-    end
+class NoCheating < IOError
+  def message
+    "The use of File.open and IO.read is restriced.\n"                \
+    'This exercise intends to help you improve your ability to work ' \
+    'with data generated from your code. Your program must not read ' \
+    'the song.txt file.'
   end
 end
-
-FoodChain.prepend NoCheating
 
 class FoodChainTest < Minitest::Test
   # This test is an acceptance test.
@@ -37,8 +24,16 @@ class FoodChainTest < Minitest::Test
   # specifically want feedback on them.
   def test_the_whole_song
     song_file = File.expand_path('../song.txt', __FILE__)
-    expected = IO.read(song_file)
+    expected  = IO.read(song_file)
     assert_equal expected, FoodChain.song
+  end
+
+  # Tests that an error is efectively raised when IO.read or
+  # File.open are used within FoodChain.
+  def test_read_guard
+    ["IO.read 'song.txt'", "File.open 'song.txt'"].each do |trigger|
+      assert_raises(NoCheating) { FoodChain.send :class_eval, trigger }
+    end
   end
 
   # This is some simple book-keeping to let people who are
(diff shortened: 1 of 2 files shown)


food-chain/food_chain_test.rb
```diff
@@ -4,6 +4,36 @@
...
+
+if FoodChain.class.eql? Class
+  class FoodChain
+    prepend NoCheating
+  end
+else
+  module FoodChain
+    prepend NoCheating
+  end
+end
```
kotp: Wouldn't lines 27..35 be cleaner if you simply wrote

```
FoodChain.prepend(NoCheating)
```

kytrinyx:
I've updated the metadata in https://github.com/exercism/x-common/commit/ddc57c953a062d277173aa63e48f3d7ab191b4bf

kotp:
@Rojo shouldn't we be seeing some tests for the new patches you wrote for IO and File classes?  I was surprised when I deleted those 11 lines and no tests failed.

The "hability" word there I am unfamiliar with, maybe a typographical error? (Line 9)

kytrinyx:
I've always thought "hability" meant "the ease with which you do something", but I looked it up in the dictionary, and it says that it's "an obsolete form of _ability_", so maybe _ability_ is better here.

Rojo:
"hability" was a typo originated to Spanish.

@kotp, can you elaborate about the tests you mentioned? do you mean the project has tests for the tests?

kotp:
As for the tests, you introduced some functionality that states that if IO.read is used, we should fail with a message.  But you have no tests to show this functionality.  So you have code introduced to the project that is untestted.

And this is pretty important tests, I think, as it is a major change.  You would want to show that your code that is being introduced works properly and as expected.

Also, because current tests use IO.read, you want to make sure that those errors would be raised appropriately, and not interfere with "normal operation" of the language.

Does that make sense?

Rojo:
Blame my bad wording. What I like are directions for that test suite. I skimmed over the `xruby` tree and docs, but was not able to find a `spec` or `test` directory, nor a mention of the test suite.

kotp:
@Rojo Sent you an invite to chat.

Basically, I am saying that you can write Ruby code in that tests that shows what is not allowed, and verifies that it will raise the exception you want to raise.  Raising `Exception` though probably is not the one you want.  Also, with a custom Exception that you create, you can also provide the message there.

kotp:
I just went through a few of the submissions for food-chain for this week, and the prevalent solution is using the text file to read the song in.  Acceptance Test Pass!  Goal stated in README, Fail!

@kytrinyx I have invited @Rojo Rojo to pair program if he can and has time. This should be corrected as soon as possible.  Hopefully we come up with a good way to guide this, without guiding the implementation while steering it away from simply reading the `song.txt`.

kytrinyx:
Thanks, @kotp -- good call on suggesting pairing!

Rojo:
Oh! Now I see what you mean, @kotp.

I was following the example in the tests for [Nth Prime](https://github.com/exercism/xruby/blob/master/nth-prime/nth_prime_test.rb#L12), that raises an exception when the implementation tries to use _primes_ related methods from the `Prime` and `Integer` classes.

As you say, the main problem was to make it restrain the use of `File.open` and `IO.read` without disturb the execution of the test suite, as it uses `IO.read`. With _RSpec_ it would be as easy as to define a couple of tests that expect that that those methods were not called but I don't know of such feature in _MiniTest_, so, prepending the module with the changes to `FoodChain` was the only way I found to pull it out.

@kytrinyx, sure, I can pair with @kotp to see if there is a better option to implement these changes.

@kotp, get in touch to schedule the session and tools to use.

kotp:
@Rojo I have sent an invite to chat on gmail so that should be pending your acceptance.  Sent you (another) email.

kotp:
Oh, and thanks.  Now I should go look at Primes exercise and make sure that if the code is deleted it at least fails the tests... For the same reason.  If you delete code and nothing fails (and nothing breaks), the code is not needed.

Rojo:
The changes worked in the pair programming session were added. (Thanks a lot, @kotp):

* A test verifies that the code that makes `FoodChain` raise a custom exception is in place. So if someone removes such code, the test will fail.
* The message displayed by the custom exception was improved.
* The code that sets the restrictions was sent to the bottom of the file, to give more relevance to the tests.

I've checked this with four scenarios:

* The solution in `example.rb`: All tests pass, as expected.
* A solution that uses `File.open`: Fails at `test_the_whole_song` with the custom exception, as expected.
* A solution that uses `IO.read`: Fails at `test_the_whole_song` with the custom exception, as expected.
* Removing the code that sets the restrictions on `FoodChain`: Fails at `test_read_guard`, as expected.

@kytrinyx, @kotp: Let me know what you think.

food-chain/food_chain_test.rb
```diff
@@ -4,6 +4,15 @@
 
 require_relative 'food_chain'
 
+class NoCheating < IOError
```
kytrinyx: I love the class name!

food-chain/food_chain_test.rb
```diff
@@ -15,14 +24,38 @@ class FoodChainTest < Minitest::Test
   # specifically want feedback on them.
   def test_the_whole_song
     song_file = File.expand_path('../song.txt', __FILE__)
-    expected = IO.read(song_file)
+    expected  = IO.read(song_file)
```
kytrinyx: Is the extra space here intentional?

kytrinyx:
I really like this approach!

food-chain/food_chain_test.rb
kotp: Probably only there for visual alignment.  Not as apparent if there were 3 or more.

kytrinyx: Oh, yepp. Works for me.

kotp:
I had not yet time to go through this.  For a solution where IO.read is being used, as expected, the message comes up nice and clear.

For a solution that uses File.read it is not so nice.

We get the message:

```
Run options: --seed 29534

# Running:
ES.
Finished in 0.002807s, 1068.7041 runs/s, 712.4694 assertions/s.

  1) Error:
FoodChainTest#test_the_whole_song:
NoMethodError: undefined method `read' for RestrictedClasses::File:Class
    /home/nitpicker/Projects/exercism_org/xruby/food-chain/food_chain.rb:6:in `song'
    food_chain_test.rb:28:in `test_the_whole_song'

3 runs, 2 assertions, 0 failures, 1 errors, 1 skips
```

It would be nice to get the same message.  So I still want to tweak this a little bit.

PR #182, opened 2015-08-15
Add tests to Rna Transcription for partially invalid input.

connorlay:
The current test suite checks for an `ArgumentError` on completely invalid input, but not partially invalid input. For example, this solution passes all of the existing tests but accepts input that has a mix of valid and invalid characters:
~~~ruby
class Complement

  DNA_REGEX = /[CGTA]/
  RNA_REGEX = /[CGAU]/

  KEYS = {
    'C' => 'G',
    'G' => 'C',
    'T' => 'A',
    'A' => 'U'
  }

  def self.of_dna(dna)
    raise ArgumentError unless dna.match(DNA_REGEX)
    dna.gsub(DNA_REGEX, KEYS)
  end

  def self.of_rna(rna)
    raise ArgumentError unless rna.match(RNA_REGEX)
    rna.gsub(RNA_REGEX, KEYS.invert)
  end

end
~~~
These additional tests might discourage novices from simply checking for the arguments present in the tests.

### DIFF
--- a/rna-transcription/example.rb
+++ b/rna-transcription/example.rb
@@ -1,5 +1,5 @@
 class Complement
-  VERSION = 1
+  VERSION = 2
 
   def self.of_dna(strand)
     DNA.new(strand).tr('CGTA', 'GCAU')
--- a/rna-transcription/rna_transcription_test.rb
+++ b/rna-transcription/rna_transcription_test.rb
@@ -68,15 +68,25 @@ class ComplementTest < Minitest::Test
     assert_raises(ArgumentError) { Complement.of_rna('XXX') }
   end
 
-  def tes_dna_raises_argument_error_on_completely_invalid_input
+  def test_dna_raises_argument_error_on_completely_invalid_input
     skip
     assert_raises(ArgumentError) { Complement.of_dna('XXX') }
+  end
+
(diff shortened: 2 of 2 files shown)


kytrinyx:
I like this change. @kotp what do you think?

Would you mind amending the commit to have the name of the exercise in it? E.g.

```
rna-transcription: add tests for partially invalid inputs
```

kotp:
I like it, plus a misnamed test method was found and corrected as well.

Test version needs to be added.

kytrinyx:
I didn't even notice the mis-named test. That test would never actually get run! Good find.

For the test version, check out how it's done here: https://github.com/exercism/xruby/blob/bb819ff5f5a3dff8b135b44228d9549e65c2f880/acronym/acronym_test.rb#L26-L30

connorlay:
@kytrinyx Should I add a `.version` file as well or just bump the version number in the test and example files?

kytrinyx:
Don't bother with the `.version` file--that's used by generators to generate the new test suite correctly. Just bumping the number should be fine.

connorlay:
Ok version bumped to 2.

kytrinyx:
Thanks!

PR #187, opened 2015-09-08
Add skip to version test of gigasecond_test.rb

kickinbahk:
This way it is not one of the first failing tests seen.


kytrinyx:
Good catch, thank you!

Issue #188, opened 2015-09-24
VERSION constant - how can we clarify this?


kytrinyx:
Someone incremented the `VERSION` number in their Hamming solution so that iteration 1 had `VERSION=1`, and iteration 2 had `VERSION=2`.

The current test says:

```
  # This test is for the sake of people providing feedback, so they
  # know which version of the exercise you are solving.
  #
  # Define a constant named VERSION inside of Hamming.
  # If you're curious, read more about constants on RubyDoc:
  # http://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/constants.html
```

Could we come up with a better explanation that clarifies that they should not change the test, and that the version is not about the solution, but about the test suite?

Should we name the constant something else?

kotp:
The version number in the solution should match the (exercise authors) version number in the test suite that you are solving for.

kotp:
I suppose if we generated the program file that they are editing, then this could automatically be included.  And that would automatically match the test version, if it is there.  Solve it in a technical way, having a note to indicate that it should only be changed if they have read 'whatever'.

kytrinyx:
Even if we generate the program file, there's still the possible confusion about incrementing the `VERSION` in their subsequent iterations. Maybe if we just say "Do not edit this test. For more about this read XYZ."

kotp:
Something that communicates it is for track maintainers only.  But they are welcome to maintain the track too.

Maybe "This number usually doesn't need to change.  Edit the test only to modify the "skip" portions -- See "How to update the track tests.""

kotp:
But if we generate that file initially, it can have the information you want in there, the hint about what the version number is there for.  I would be careful to not drive anything that makes their decision of a module or class though, or something other.

kytrinyx:
Right, but we can also put the information in the test. That's the piece that we don't want people to change (if they don't change it in the test, they can't change it in the solution).

kotp:
True..  Then it comes down to the track maintainers being informed to do so.

This is very likely system wide update?  Not sure how many language tracks have the multiple tests implemented.

kytrinyx:
At the moment I would settle for just fixing it in Ruby. I think that it has been used in Go as well, but I haven't seen any evidence that gophers are confused by it.

sguermond:
"Problems in exercism evolve over time as we find better ways to ask questions. The version number refers to the version of the problem you solved, not your solution."

kytrinyx:
I like it, @sguermond!

[...]

PR #189, opened 2015-09-29
Add version test for Accumulate with updated comment for clarity

kickinbahk:
See [ticket](https://github.com/exercism/xruby/issues/188) for where this was recommended.

### DIFF
--- a/accumulate/accumulate_test.rb
+++ b/accumulate/accumulate_test.rb
@@ -45,5 +45,4 @@ class ArrayTest < Minitest::Test
     original.accumulate { |n| n * n }
     assert_equal copy, original
   end
-
 end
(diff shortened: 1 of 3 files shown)


kytrinyx:
It's passing in the very last commit, but I'd really like to do all of these things separately.

Would you mind making a change to just the rna-transcription exercise, and only about updating the `VERSION` documentation?

That way if there's a complaint from travis about that, we can focus in on exactly why.

kickinbahk:
Okay sure! I am still a little new to this open source stuff...would the way that I should do it be to open a branch for each exercise to update the comments?

kytrinyx:
Yepp, a branch for each would be a great idea.

Sometimes (if the change is trivial) it's fine to make the exact same change to all the different exercises at once, but it's easier to review things and easier to merge when the PRs are very distinct.

kytrinyx:
Oh, and I'm always happy to help sort out the details, don't hesitate to ask!

PR #190, opened 2015-09-29
Update RNA_transcription version test comments to be more clear for user

kickinbahk:
Clarify the comments in the test for the version number so the user knows that they do not need to increment the version number


kytrinyx:
Thanks!

PR #191, opened 2015-09-30
Update word-count version test comments to be more clear for user

kickinbahk:


kytrinyx:
This looks great, thank you!

PR #195, opened 2015-10-01
Update food-chain version test comments to be more clear for user

kickinbahk:
Found that users were getting confused by the version test and it's purpose (see [ticket 188](https://github.com/exercism/xruby/issues/188)). Goal is to make it clear to users that there only need to return the number that is asked. They do not need to bump the versions themselves at any point. We have also included a link to the ruby docs on contstants so if a user is curious at what they are inserting, they can read more about them

### DIFF
--- a/food-chain/food_chain_test.rb
+++ b/food-chain/food_chain_test.rb
@@ -36,8 +36,14 @@ class FoodChainTest < Minitest::Test
     end
   end
 
-  # This is some simple book-keeping to let people who are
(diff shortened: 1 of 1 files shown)


kytrinyx:
Would you mind squashing these commits? They're both a change on food-chain, as far as I can tell.

kickinbahk:
@kytrinyx I think this one should be good to go too.

kytrinyx:
Yepp, this looks great. Thanks!

PR #197, opened 2015-10-01
Update acronym version test comments to be more clear for user

kickinbahk:
Found that users were getting confused by the version test and it's purpose (see ticket https://github.com/exercism/xruby/issues/188). Goal is to make it clear to users that there only need to return the number that is asked. They do not need to bump the versions themselves at any point. We have also included a link to the ruby docs on contstants so if a user is curious at what they are inserting, they can read more about them

### DIFF
--- a/acronym/acronym_test.rb
+++ b/acronym/acronym_test.rb
@@ -23,8 +23,14 @@ class AcronymTest < Minitest::Test
     gem 'minitest', '>= 5.0.0'
   end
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
Would you mind squashing these commits?

kickinbahk:
@kytrinyx I think this is squashed?

kytrinyx:
Yepp, thanks!

PR #198, opened 2015-10-01
Change refute to assert

mamenama:
By adding the error messages, we now need to assert a match to the 'No...' message instead of refuting the return value. As is, the refute test fails when the statement is matched.


mamenama:
NM - I was not understanding the nature of the error message. Doh.

kotp:
+1 At least you are contributing!

PR #199, opened 2015-10-06
Update hamming version test comments to be more clear for user

kickinbahk:
Found that users were getting confused by the version test and it's purpose (see ticket https://github.com/exercism/xruby/issues/188). Goal is to make it clear to users that there only need to return the number that is asked. They do not need to bump the versions themselves at any point. We have also included a link to the ruby docs on contstants so if a user is curious at what they are inserting, they can read more about them.

### DIFF
--- a/hamming/example.tt
+++ b/hamming/example.tt
@@ -12,11 +12,13 @@ class HammingTest < Minitest::Test<% test_cases.each do |test_case| %>
     assert_equal <%= test_case.expected %>, <%= test_case.do %><% end %>
   end
 <% end %>
-  # This test is for the sake of people providing feedback, so they
-  # know which version of the exercise you are solving.
+  # Problems in exercism evolve over time,
+  # as we find better ways to ask questions.
+  # The version number refers to the version of the problem you solved,
+  # not your solution.
   #
   # Define a constant named VERSION inside of Hamming.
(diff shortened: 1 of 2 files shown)


kickinbahk:
@kytrinyx I think this is the last version commit that needs to happen. I know that you said that hamming is a bit more complicated due to a generator. It passed travis, so not sure if I am missing something or it just is not affected by touching the comments only. Let me know if I need to do anything else for this one!

kytrinyx:
The hamming_test.rb file is generated from the example.tt file, so we need to add the change to the example.tt file as well, otherwise the comment will revert the next time we regenerate the test suite (e.g. if the data in the x-common repo changes).

kickinbahk:
Cool, I would love to understand how that works. I have updated the comments for both files. Let me know if I need to fix anything else.

kytrinyx:
Take a look at the README here in the xruby repository, I've tried to describe how the generator works.

In order to try it out, clone the x-common repository and change something in the `x-common/hamming.json` file, and then run bin/generate-hamming, and take a look at what changed.

I'd love to have generators for more of these problems! I opened a ticket about this a while back, if you're interested in working on something like this: https://github.com/exercism/todo/issues/8

kytrinyx:
:+1:

PR #202, opened 2015-10-09
Grade school patch

remcopeereboom:
Changed grade-school problem to use `#to_h` instead of `#to_hash`.

### DIFF
--- a/grade-school/example.rb
+++ b/grade-school/example.rb
@@ -1,4 +1,6 @@
 class School
+  VERSION = 1
+
   def initialize
     @db = Hash.new { |db, grade| db[grade] = [] }
   end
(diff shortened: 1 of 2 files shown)


kytrinyx:
@remcopeereboom Do you have a very old version of the ruby repository? We had a bit of a snafu where we had to edit the history: https://github.com/exercism/xruby/issues/28

Would you mind resetting --hard with master and then cherry-picking your patch into a fresh branch?

remcopeereboom:
@kytrinyx - Strange, I just forked the same day. It should be okay now.

remcopeereboom:
@kytrinyx Strange. I forked only minutes before. I have rebased and all should be well now.

kytrinyx:
Very strange! This looks great now. Thanks!

PR #203, opened 2015-10-10
Cleaned up references to problems and test-suites

mikeyjcat:
The terms problems and test suites were being used interchangeably, which confuses new contributors. Have corrected this to be clearer about what is being referred to, either problems or test suites.


kytrinyx:
This is fantastic, thank you so much!

PR #204, opened 2015-10-10
Ruby leap tests gen

mikeyjcat:
Created test generator for ruby leap problem


leap/example.rb
```diff
@@ -1,25 +1,12 @@
 class Year
```
kytrinyx: Any particular reason why you changed the program to use class methods instead of delegating to the instance?

leap/leap_test.rb
```diff
@@ -15,26 +18,44 @@ def leap?
 
 class YearTest < Minitest::Test
   def test_leap_year
-    assert Year.leap?(1996), 'Yes, 1996 is a leap year'
+    assert_equal true, Year.leap?(1996)
```
kytrinyx: If the value you are asserting against is `true` or `false`, it's idiomatic in minitest to use `assert` and `refute`.

leap/leap_test.rb
```diff
@@ -15,26 +18,44 @@ def leap?
...
+    assert_equal false, Year.leap?(1900)
   end
 
   def test_fourth_century
     skip
-    assert Year.leap?(2400), 'Yes, 2400 is a leap year'
+    assert_equal true, Year.leap?(2400)
+  end
+
+  def test_Y2K
```
kytrinyx: I'd leave the method name all lowercase.

leap/leap_test.rb
```diff
@@ -15,26 +18,44 @@ def leap?
...
+  # as we find better ways to ask questions.
+  # The version number refers to the version of the problem you solved,
+  # not your solution.
+  #
+  # Define a constant named VERSION inside of Leap.
+  # If you are curious, read more about constants on RubyDoc:
+  # http://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/constants.html
+  def test_bookkeeping
+    skip
+    assert_equal 2, Year::VERSION
```
kytrinyx: If the API of the problem hasn't changed, let's leave the version as 1.

kytrinyx:
Neat! I'm very excited to see another generator here!

I've left a few style comments, but there's nothing major.

The test suite is failing on CI for two reasons:

```
rubocop -fs -D
make test
```

leap/example.rb
mikeyjcat: The existing solution was very wordy, with attr_reader, initialise, etc. Happy to revert to the original solution if it's better style.

leap/leap_test.rb
mikeyjcat: OK. I kept the tests the same as they were previously, will adjust.

leap/leap_test.rb
mikeyjcat: As per previous comment, this was the name as per the json test definition, I'll sort it out.

leap/leap_test.rb
mikeyjcat: Pff, I thought I'd corrected that. It seems as each time you generate the tests, it gets incremented and I must have lost track of resting it. Will sort it out.

mikeyjcat:
OK, I'm clueless here. Rubocop style issues all good, fixed all the issues referred to above. Failed on execution bit not set, and I have no clue on what that is:)

kytrinyx:
> the learning curve is a bit steep.

Yeah, sorry about that :( Thanks so much for being patient and working out all the various issues!

The execution bit is a linux/mac thing (it might exist on windows, but I'm not familiar with the platform).

It means that the file is marked as being "runnable" like a script or program.

On mac/linux the fix for it is to say:

```
chmod +x path/to/file
```

leap/example.rb
kytrinyx: The thing that I like about the existing solution is that the `divisible_by?` is much more readable. In the version that you added there are two numbers, and it's not clear which is the number being divided and which is the factor that is being checked. With the other one, there's no ambiguity.

leap/leap_test.rb
```diff
@@ -15,26 +17,38 @@ def leap?
 
 class YearTest < Minitest::Test
   def test_leap_year
-    assert Year.leap?(1996), 'Yes, 1996 is a leap year'
+    assert Year.leap?(1996)
```
kytrinyx: We lose the friendly error messages. Would it be worth adding them to the leap.json file over in x-common? I don't know whether or not they're useful to people, though they are a good practice, and I like modeling good practices where possible.

mikeyjcat: Good point, sorry I missed that. Again, I think it requires a design decision. The existing pattern does not have the messages in either the JSON file or the test generator template. While one could take either approach, I think adding the messages to the JSON file would provide consistency across languages and push less logic into the generators.
Once we have a call made, then I'd need to back-fill on the existing Ruby test generators and JSON files.
Trying the Go ones would be very dangerous:)

mikeyjcat:
Two of these issues seem to be design patterns for the test generation. Maybe it would be good to document these somewhere so there would be consistency across tests within a language, and between the various languages? Not sure were the right place would be to touch across all languages?

leap/example.rb
mikeyjcat: Makes sense, will revert it.

mikeyjcat:
In summary, looks like three things outstanding:
- Pattern for the case on test names in the JSON file which map to test method names in Ruby (but not go). I believe we should have lower case, and the Ruby generator works on that assumption. Languages who use camel case would need to capitalise where needed. Go seems to not create test method names from this to no issues introduced by this.
- Pattern for whether test error messages should be included in the JSON file. I believe this makes sense and happy to fix leap and the other JSON files, and back fill the existing Ruby generators.
- Decision on where/how to keep these design decisions. Again, happy to implement, just don't have the full picture yet so need a steer from you.

Travis checks passing though, so good progress.

mikeyjcat:
One last comment. It seems as though the rubocop config does not check the example.rb. Is this intended?

leap/leap_test.rb
kytrinyx: Let's start by adding it to the json file and to this generator. We can backfill later. I'll do the Go ones :)

kytrinyx:
> It seems as though the rubocop config does not check the example.rb. Is this intended?

Yes, we intentionally ignore style in the reference solutions.

> Pattern for the case on test names in the JSON file which map to test method names in Ruby 

Think about the JSON file as documentation. The descriptions should be readable English that explain context. Then each language can do with that as they wish. If Ruby wants to use this for test names, then Ruby will need to process it (downcase, snake case, etc).

> Pattern for whether test error messages should be included in the JSON file.

Yeah, I think that having a friendly failure message means that we can happily ignore it if we like, but it gives us the option to use it where it makes sense.

> I believe this makes sense and happy to fix leap and the other JSON files, and back fill the existing Ruby generators.

That would be lovely!

> Decision on where/how to keep these design decisions.

I think that for the JSON files, we should put something in the x-common README about what data we have found to be useful, and that each language is welcome to keep/ignore as they wish.

For the decisions about the Ruby generators, let's keep adding to the xruby README.

kytrinyx:
This looks good. I'd like to wait to merge it in until we have the friendly failure messages included.

Then I'd like to have the commits squashed, since there's a lot of going back and forth on the same thing.

mikeyjcat:
Yes to all of the above. I'll fix up the bits and pieces, and when all sorted squash the commits.

kytrinyx:
Wonderful! :heart: Thanks so much for taking the time to not only work through this, but to figure out how to improve each of the pieces along the way.

mikeyjcat:
I think that's done now. All thats left is to squash the commits once you're satisfied.
I'll also update the two readme files later, need to go to work now.

kytrinyx:
This looks great. Thank you!

PR #205, opened 2015-10-14
Included messages in test generator

mikeyjcat:
I''ve only included the generator template on the understanding the the test generation would be run to crate the new set of tests. I I should include the other affected files, please let me know (.version, example.rb and gigasecond_tests.rb).


kytrinyx:
No, I think this is fine.

PR #208, opened 2015-10-18
Add more test generators

mikeyjcat:
Enhanced generator to cater for two part problem names

Added test generators for hello_world, na_transcription and raindrops


kytrinyx:
This seems like the same PR that you submitted and closed. What happened?

Please submit PRs that are only about one thing at a time. Here you're:

- fixing rubocop violations (were they introduced in other commits in the PR? Were they already a problem?)
- adding 3 different generators
- refactoring the generator

This should probably be 3 PRs, one for each generator.

PR #210, opened 2015-10-19
Add hello world test generator

mikeyjcat:
Created hello world test generator incorporating feedback.


hello-world/hello_world_test.rb
```diff
@@ -12,33 +12,25 @@
...
+# 7668b09 Added hello world test definition
+
 class HelloWorldTest < Minitest::Test
   def test_no_name
-    assert_equal 'Hello, World!', HelloWorld.hello, <<-MSG
-    When given no name, it should greet the world!
-    MSG
+    assert_equal 'Hello, World!', HelloWorld.hello,
+                 'When given no name, it should greet the world!'
   end
```
kytrinyx: I really don't like it when a line is split onto a second line just to make it not exceed 80 characters. I know that it's common in Ruby, but it looks inconsistent, is much harder to scan, and seems like an exercise in box-ticking (_let's not let Rubocop complain, people!_).

If you don't like the HEREDOC's let's assign to a temporary variable above the assertion instead.

lib/hello_world_cases.rb
```diff
@@ -0,0 +1,23 @@
...
+    'test_%s' % description.gsub(/[ -]/, '_')
+  end
+
+  def do
+    defined?(name) ? "HelloWorld.hello('#{name}')" : 'HelloWorld.hello'
+  end
+
+  def raises_error?
+    expected.to_i == -1
+  end
```
kytrinyx: There are no error cases in hello world.

mikeyjcat: Will fix, happens when you clone code too early in the morning.

hello-world/hello_world_test.rb
mikeyjcat: So here's an interesting discussion, and I stopped creating automated tests on this one as this is a common problem that is going to occur repeatedly.

The problem is that for most of the tests where there are messages present, the lines are longer than 80 characters.

My choices were:
 -Use HEREDOC, which appears to be the ideal and recommended thing to use for multi-line test literals. This does introduce quite a lot of noise and density into the code, as it changes every test into three lines, and indents the test message against the right, which does not read well. 
- Split the line and indent the message. While it does affect readability, it keeps the code light, and I believe easier to grok than the the HEREDOC approach.
- Assign the message to a variable and then use the variable. I really believe that this is a box-ticking exercise that adds complexity into the code, and obscures easy reading. 
- Attempt to squish the messages which sort of defeats the purpose of having the messages at all.

It feels like the only reason to do any of the three is to tick the rubocop box, as none adds either readability nor functionality. They all feel like a compromise:(

When I looked at the code generated for something long like the RNA problem, it became clear (to me anyway), that the HEREDOC and variable approaches made the code denser, longer, and not more easy the read or understand which is the key driver for the theory behind style.

I do feel somewhat like a lawyer making a case, but this affects much of the work to be done for all of the test generation when messages come into play, so it's worth a bit of effort.

Thoughts?

mikeyjcat: @kytrinyx the above is the clarification I was looking for, thanks

kytrinyx: Thanks! I completely missed this one last time around.

You've listed several options:

* assign a temporary variable
* use a heredoc
* break the line
* make the message shorter

Every single one of these are a box-ticking exercise, because Rubocop (and the Ruby community in general) has decided that 80 characters is the line limit.

There's an additional option:

* just ditch using extra messages

Every once in a while the assertion failure doesn't have good inherent understandability, and an extra message is helpful (and a good practice). The rest of the time messages are noise. They add to the code, they add to the error message, and if they're not pulling their weight we should just ditch them.

The question is: is it worth having support for them to handle the few cases when they're useful?

If I were to rank these options in order of *I hate it the most* to *I hate it the least*, my list would look like this:

* break the line (I triple hate this with fudge on top)
* make the message shorter (defeats the purpose)
* use a heredoc (bulky, more code)
* assign a tmp (yeah, whatevs)
* just ditch messages altogether (seriously, not many people use them)

I think I've changed my mind a bit on messages overall over the past couple of years. I think that people should know how to do it, and understand how to read the assertions / test suite when they're there, and I used to think it was important enough to add this to exercism, but honestly it's not worth the headache of ticking rubocop boxes.

What do you think? tmp var or ditch them altogether?

mikeyjcat: I agree, they are all box ticking, and generally the messages aren't needed if the tests are well named.

I think the best is to ditch messages, and if they are really needed, then use a heretdoc as they probably need to be multiple lines then anyway.

Happy with this approach?

kytrinyx: Yepp, I like this approach.

mikeyjcat:
Updated and regenerated test suite with corrected test definition.

kytrinyx:
Lovely, thanks!

PR #212, opened 2015-10-22
Fixing grammatical error.

sivabudh:
Should use possessive pronoun 'its'.

Reference: http://grammarist.com/spelling/its-its/


kytrinyx:
Good catch, thanks.

PR #213, opened 2015-10-25
Added rna_transcription test generator

mikeyjcat:
Slight change in the test file is due to the original having a (very) slightly inconsistent sequence of tests.

Otherwise should address all of the previous questions, discussions.


kytrinyx:
I added some comments on the JSON cases in x-common. Once those are fixed, would you mind regenerating and then adding a comment here so I know to look at it again?

mikeyjcat:
Corrected comments in JSON files and regenerated test suite.

kytrinyx:
Thanks!

PR #216, opened 2015-10-28
Added roman numerals

mikeyjcat:
Should we continue to include the comments about the VERSION constant? This is about the fourth or fifth time it's been seen, so I imagine people have got used to it by now?


kytrinyx:
Let's make it a one-liner at this point.

mikeyjcat:
Hmm, I'm a bit stuck here. Everything seems fine apart from the line generated in the test which pulls in the git information, which is a few characters too long due to the json commit message. I have no clue how to work around this? Any thoughts or suggestions?

kytrinyx:
Hm. How about just manually shortening the name of the branch in that comment to `roman-numerals`? The critical information is the SHA1 hash and the PR number and your username.

mikeyjcat:
I can do that. Maybe I should truncate the SHA1 hash, getting best of both worlds (as much of branch name as possible)?

kytrinyx:
I like having a length of 7 for the SHA1 hash. I know that it's unlikely to have a collision even with fewer, but seven seems to be the shortest that people generally use when passing around SHA1 hashes. I don't really think that the branch name is important for several reasons:

1. you can name branches whatever you want
2. once the branch is merged, it gets deleted, so nobody goes looking for it
3. if you're going to look something up, you'd use the SHA1 or the PR #

mikeyjcat:
Closing to shorten branch name.

kytrinyx:
I was suggesting that you just hand-tweak this in the file that was causing the complaint.

mikeyjcat:
😄

PR #217, opened 2015-10-29
Correct typo in Ruby SETUP instructions

lizvdk:
I know this is extremely nitpicky, but... 'running' was missing an 'i'.


kytrinyx:
I <3 extremely nitpicky. Thanks!

PR #223, opened 2015-11-05
Add refute cases to custom-set equal test

mamenama:
Adds some counterexamples to the equal test. Since defining equality is part of the solution, some implementations may pass the assert_equal test without really testing for equality (e.g. one implementation I wrote passed the assert_equal test but incorrectly allowed sets to be equal even if one had more elements).


kytrinyx:
Yeah, this looks good. @kotp any thoughts about this one?

kotp:
Good catch, @mamenama any reasons we should also check for values of 0?

kotp:
Ready to merge this in?

kytrinyx:
yeah, good call.

PR #225, opened 2015-11-15
Add new test for trinary

haus:
Previously, certain implementations of trinary could pass all of the
tests despite doing no validation on the input. This is because "a".to_i
(and any non-numeric character) is 0 in ruby, so an all alpha test
string will evaluate to 0. This commit adds a string that has numbers
mixed in with letters, to ensure that validation is required for tests
to pass. This also updates the example to do the required validation.

### DIFF
--- a/trinary/example.rb
+++ b/trinary/example.rb
@@ -3,6 +3,7 @@ class Trinary
 
   attr_reader :digits
   def initialize(decimal)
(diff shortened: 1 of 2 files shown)


haus:
oops. looks like i broke the tests.

kytrinyx:
Nice catch, thank you!

PR #229, opened 2015-12-07
Normalize YAML in Travis CI config

kytrinyx:
The YAML was not quite valid. Travis can read it
and wasn't complaining, but this ensures that the YAML
is clean.


kytrinyx:
Updated per @kotp's comments.

PR #235, opened 2015-12-29
update tests for string output

adriennedomingus:


kytrinyx:
I've added a comment to your exercism submission--the `leap?` method should return a boolean, not a string. Check out the difference between `assert` and `assert_equal` in the minitest documentation for more details about this.

Cheers

PR #236, opened 2016-01-01
improve beer song API and tests

kytrinyx:
The final beer-song test was not actually testing anything.
This is much more straight-forward, asserting that the entire song's
output matches the expected lyrics.


kytrinyx:
I've rebased the branch onto master. Any qualms about this change?

kotp:
Simple enough.

PR #239, opened 2016-01-09
update output

Kosmas:
updated with output from running on linux.


kytrinyx:
Thanks, good catch.

Kosmas:
Thank you for having the idea in the first place :+1: 
I've been meaning to start since your talk in BaRuCo, but only managed to do it now!

kytrinyx:
Sweet. Keep the tweaks coming if you come across weird things :)

PR #241, opened 2016-01-22
Move exercise foldres to subdirectory (xruby)

kotp:
references #240


kytrinyx:
It looks like `make test` is failing.

kotp:
Does the test account for the exercises folder but not having an exercises.rb file in that folder?

> cannot stat ‘exercises/example.rb’

kytrinyx:
Good question. Does the test use the config.yaml which hasn't put "exercises" in the ignore section?

kotp:
Found it, in the travis config.

kotp:
The check for ending whitespace disabled early on in the tests, due to the current requirement to have padding in the heredocs construction.  Doesn't exactly go with the 'fix' but it slipped through earlier commits, when it should have been failing CI all along.

kytrinyx:
Right, that makes sense. Thanks!

PR #242, opened 2016-01-23
largest-series-product: Add generator

petertseng:
This generates tests from the new common test data for
largest-series-product.

The example solution had to be updated to:

* Add a version constant
* Reject negative spans
* Reject non-numeric strings


largest-series-product/largest_series_product_test.rb
```diff
@@ -3,79 +3,112 @@
...
 # Rubocop directives
 # rubocop:disable Style/AlignParameters:
 #
 class Seriestest < Minitest::Test
-  def test_largest_product_of_a_tiny_number
-    series = Series.new('12')
-    assert_equal 2, series.largest_product(2)
+  def test_can_find_the_largest_product_of_2_with_numbers_in_order
+    assert_equal 72, Series.new('0123456789').largest_product(2),
+                 'can find the largest product of 2 with numbers in order'
```
petertseng: these are redundant. removing.

largest-series-product/example.rb
```diff
@@ -17,6 +19,8 @@ def largest_product(length)
   private
 
   def validate_length
+    @length < 0 and
```
petertseng: it's a weird style of failing (using `and`). I got the logic wrong both here and below, clearly I'm not used to it.

I would have just done it with `fail(...) if ...` but I felt I needed to be consistent.

kotp:
Can you rebase this on current master branch of exercism/xruby ?

Otherwise it looks good.

petertseng:
Ah, right, the -> exercises dir move. Rebased, made sure the files ended up in the right place exercises/largest-series-product not largest-series-product now

kytrinyx:
Thanks!

PR #245, opened 2016-01-24
Deprecated cops, deprecated file patterns

kotp:
Made corrections for deprecated file patterns and style cops

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,33 +1,30 @@
 AllCops:
   Exclude:
-    - '*/example.rb'
+    - 'exercises/**/example.rb'
     - 'bin/*'
-Style/Documentation:
+Documentation:
   Enabled: false
-Style/StringLiterals:
+StringLiterals:
   Enabled: true
 Metrics/AbcSize:
   Enabled: false
-Style/SpaceAroundEqualsInParameterDefault:
+SpaceAroundEqualsInParameterDefault:
   EnforcedStyle: no_space
-Style/TrailingComma:
+TrailingCommaInArguments:
   EnforcedStyleForMultiline: comma
-Style/TrailingWhitespace:
-  Exclude:
-    - 'ocr-numbers/ocr_numbers_test.rb'
-Style/BlockDelimiters:
+BlockDelimiters:
   Enabled: true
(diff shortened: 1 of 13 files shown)


kotp:
This is not ready, gem version for Rubocop needs to be specified.

kotp:
@kytrinyx Where should we set the version requirement for Rubocop?

kytrinyx:
Is it not in the .travis.yml file? We could add a gemfile, too.

kotp:
It is indeed in the travis file.  What version should we use?  The current version is 0.36.0 and unfortunately, my local check script does not check travis.yml file.

The real problem becomes:  The local check in bin does not enforce the rubocop running version based on the travis config setting, which is the real cause of this confusion.

kytrinyx:
Yeah, let's add a Gemfile -- then we can use bundle exec locally.

If the most recent version of rubocop works with what we have, then we may as well upgrade, I think.

kotp:
This means that the travis file should use `Gemfile` to keep things up to date.  One place to do the configuration.

Then it will not be 'bundle exec locally' bundle exec no matter where.

I have the following code example for the Gemfile.

However, perhaps the travis config file should be to do the needful, with the bundle exec command instead, and then instead of the gem file getting gem information from Travis, travis configuration does its commands from there.

```ruby
source "https://rubygems.org"
require 'yaml'

# We read the travis.yml file and pull information from there
config = YAML.load(IO.read('./.travis.yml')).to_hash

config['install'].each do |g|
  gem_detail = g.split
  # Example lines from travis, ignoring the yaml array directive "- "
  # gem install rubocop -v 0.36.0 --no-document
  # gem install minitest --no-document
  # These will both be appropriate as the 2nd and 4th elements will work with
  # the gem command as parameters appropriately.
  gem gem_detail[2], gem_detail[4]
end
```

 I like the code better that will be coming in shortly though. :)

kytrinyx:
> This means that the travis file should use Gemfile to keep things up to date. One place to do the configuration.
> 
> Then it will not be 'bundle exec locally' but bundle exec no matter where.

Agreed. Also agreed that travis should just bundle and use that.

This all looks good, thanks!

PR #248, opened 2016-01-30
(The sum of) Two sides can be equal but not less than third.

kotp:
re: #247

### DIFF
--- a/exercises/triangle/example.rb
+++ b/exercises/triangle/example.rb
@@ -39,7 +39,7 @@ class Triangle
   end
 
   def violates_inequality?
-    a + b <= c || a + c <= b || b + c <= a
+    a + b < c || a + c < b || b + c < a
   end
 
   def impossible_length_side?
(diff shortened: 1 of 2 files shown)


kytrinyx:
I think that we should also have a test for the edge case where two sides exactly equal the third. Thoughts?

kotp:
Yes, I think the original triangle that was created is a good test for this.

kytrinyx:
There's more discussion in some of the other tracks about degenerate triangles: https://github.com/exercism/xerlang/issues/85

kotp:
Looking at the erlang.  Left message there.  I don't think this exercise is suited for getting into post elementary geometry, it isn't the point of the exercise.  We don't cover more than three well known and standard triangles, out of the six that we learn about in 3rd or 4th grade.  Before we make statements about degeneracy, we have room to expand on the other three types of triangles that have area.

kotp:
Leave it as is, not taking into account degenerative values, see conversation: exercism/xerlang#85

PR #251, opened 2016-02-05
Unused and unasked for to_s removed

### LINKED ISSUE #250: Exercise "simple_linked_list" has no requirements stated for `to_s` method
The method `to_s` in `Element` isn't tested for and has no specification of what it should be, though the `example.rb` file shows an implementation of it.

kotp:
fixes #250


kytrinyx:
Nice!

PR #252, opened 2016-02-05
Cleanup simple-linked-list

### LINKED ISSUE #249: simple-linked-list: overload Nil#nil? or have non-nil class return true for nil?
There are two tests that cause solutions to this problem, to have, what I think, is non-idiomatic ruby.

* https://github.com/exercism/xruby/blob/master/exercises/simple-linked-list/simple_linked_list_test.rb#L45 requires that the object returned from `Element.from_a` return `true` for `#nil?`
* https://github.com/exercism/xruby/blob/master/exercises/simple-linked-list/simple_linked_list_test.rb#L71 requires that the object returned in the previous bullet point respond to `#reverse`

There are two solutions to this:
* add `Nil#reverse` which is altering a core class of ruby, which I'm not really a fan of. Yes you could do this with refinements, but I'm not a fan of that solution either
* Create a `NullElement` or other similar `NullObject` pattern that returns `true` for `#nil?` which I consider bad form also.

The latter is the solution that the example program uses (https://github.com/exercism/xruby/blob/master/exercises/simple-linked-list/example.rb#L63-L65) and I think this is bad form in ruby because:

```
null_element = NullElement.new
puts "null element is #{null_element.nil? ? "nil" : "not nil"}"
puts "null element is #{null_element ? "truthy" : "falsy" }"
# => null element is nil
# => null element is truthy
```
And that output is just wrong.

How should we fix this?

copiousfreetime:
This is an update to simple linked list to address the issues in #249.

Changes:
* We do not test reversing a list created from an empty array since `Element.from_a([]) # => nil`
* We stop testing `Element#next` for nil and start testing `Element#tail?`
* Split all the tests up into separate methods since previously many things were being tested in each test method.
* In the example we no longer need to use a NullObject pattern.

closes #249

### DIFF
--- a/exercises/simple-linked-list/example.rb
+++ b/exercises/simple-linked-list/example.rb
@@ -1,62 +1,68 @@
 class Element
-  attr_reader :datum
-  attr_reader :next
+  attr_reader   :datum
+  attr_accessor :next
 
-  def self.from_a(arr)
-    arr.reverse_each.inject(NullElement.new) { |memo, obj| new(obj, memo) }
-  end
-
-  def self.each_datum(elem)
-    until elem.nil?
-      yield elem.datum
-      elem = elem.next
-    end
-  end
-
-  def self.to_a(elem)
-    arr = []
-    each_datum(elem, &arr.method(:push))
-    arr
-  end
-
-  def self.reverse(elem)
-    res = nil
-    each_datum(elem) { |datum| res = new(datum, res) }
-    res
-  end
-
-  def initialize(datum, next_element = NullElement.new)
+  def initialize(datum, next_element=nil)
     @datum = datum
-    @next = next_element
+    @next  = next_element
   end
 
-  def to_a
-    self.class.to_a(self)
-  end
-
-  def reverse
-    self.class.reverse(self)
+  def tail?
+    @next.nil?
   end
 end
 
-class NullElement
+class SimpleLinkedList
+  attr_reader :size
+  attr_reader :head
+
+  def initialize
+    @head = nil
+    @size = 0
+  end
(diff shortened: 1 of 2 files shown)


copiousfreetime:
Well after sleeping on it I think that this is wrong also, there are really two concepts in a linked list, the list itself and the elements of the list. The existing tests assume that they are one and the same. Or it sort of says that the Element class methods operate on the list and the instances of Element are the elements. 

I'm going to update this pull request with essentially a rewritten set of tests.

copiousfreetime:
After looking at the at the [documentation for this exercise](https://github.com/exercism/x-common/blob/master/simple-linked-list.md) it looks like the goal is to just use one class. So I'll let my rework of the tests as it is stand.

kytrinyx:
No, let's fix the documentation! I remember the exercise being submitted initially and thinking that we'd eventually have to improve it--it kind of felt like a placeholder (but I didn't have the time/bandwidth to think it through).

copiousfreetime:
Okay, that sounds good, I'll put a pull request in on the x-common side too.

kytrinyx:
Sweet, thanks!

copiousfreetime:
@kytrinyx time for the critique :smile:. also updated documentation in exercism/x-common#178

kytrinyx:
This is a great improvement! Would you mind rebasing onto master and squashing the commits? Or if you think that some of them make good stand-alone commits would you mind just editing the subject line to include the name of the exercise?

copiousfreetime:
Rebased and squashed

kytrinyx:
Thank you!

PR #259, opened 2016-02-29
Add more cases to say_test.rb

kangkyu:
in billion or million, there some missing cases I think, could be tested for "three hundred million" or "three hundred forty million", "three hundred fifteen million" etc (any of those digits in 10th or 11th place are 0, or the two digits ranged from 11 to 19)


kytrinyx:
This looks good to me. Thanks!

kotp:
These are missing the initial `skip` directives.  Can you fix these?

PR #261, opened 2016-03-06
small typo meetup_test.rb

kangkyu:


kytrinyx:
Good catch, thank you!

PR #266, opened 2016-03-22
edit GETTING_STARTED.md with proper case

gilmatic:
The example gave an incorrect solution as the test was looking for 'Hello, World!' not 'Hello, world!'


kotp:
Thanks @gilmatic !

PR #271, opened 2016-03-27
Nil is not special objects in a circular buffer

### LINKED ISSUE #257: Inconsistencies in the Circular Buffer exercise
In this exercise, `write!` forces a value in the buffer to be overridden. This behavior is reinforced by `test_overwriting_oldest_item_in_a_full_buffer`. However, this leads to what I believe is a test error in `test_forced_writes_of_nil_should_not_occupy_buffer`.

The beginning of the `test_forced_writes_of_nil_should_not_occupy_buffer ` test reads as follows (`circular_buffer_test.rb` lines 92-96, with the resulting buffer contents in the comments):

    buffer = CircularBuffer.new(2)  # [nil, nil]
    ('1'..'2').each { |i| buffer.write i }  # [1, 2]
    buffer.write! nil  # [nil, 2], I believe                  
    assert_equal '1', buffer.read  # test failure, 1 != nil
    assert_equal '2', buffer.read

The only way for both of these asserts to pass would be for the `buffer.write! nil` call to have no effect and leave the buffer as `[1, 2]`. However, previous tests of `write` with `nil` show that it should clear the value of that location. Therefore, I believe the two tests mentioned above are in conflict with one another.

kotp:
Removes tests for nil as content for the buffer

Fixes #257

### DIFF
--- a/exercises/circular-buffer/circular_buffer_test.rb
+++ b/exercises/circular-buffer/circular_buffer_test.rb
@@ -60,14 +60,6 @@ class CircularBufferTest < Minitest::Test
     assert_equal '3', buffer.read
   end
 
-  def test_writes_of_nil_should_not_occupy_buffer
-    skip
-    buffer = CircularBuffer.new(5)
-    buffer.write nil
-    ('1'..'3').each { |i| buffer.write i }
-    assert_equal '1', buffer.read
-  end
-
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks! And thanks, also, for helping think through the various red herrings (and nil clowns).

PR #273, opened 2016-04-04
Scrabble Score case insensitive

### LINKED ISSUE #272: Scrabble Score Test
One of the tests for the Scrabble Score exercise looks like this:

```ruby
  def test_scores_are_case_insensitive
    assert_equal 41, Scrabble.new('oxyphenbutazone').score
  end
```

If this is testing it will still calculate scores for words with uppercase and lower case letters, then shouldn't the test look something like this?

```ruby
  def test_scores_are_case_insensitive
    assert_equal 41, Scrabble.new('OxypHenButazoNe').score
  end
```

kotp:
Use a uppercased word to test case insensitivity

fixes #272

### DIFF
--- a/exercises/scrabble-score/scrabble_score_test.rb
+++ b/exercises/scrabble-score/scrabble_score_test.rb
@@ -40,7 +40,7 @@ class ScrabbleTest < Minitest::Test
 
   def test_scores_are_case_insensitive
     skip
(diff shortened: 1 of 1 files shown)


kytrinyx:
Yeah, good call. Thanks.

PR #277, opened 2016-04-07
Test for name collisions

callahat:


kotp:
You have brought up a good point.  Does `test_different_robotos_have_different_names` occasionally fail for the same reason that you set the random seed in your test?

Should the test name be less 'tech-speaky'?

callahat:
It is extremely unlikely that `test_different_robots_have_different_names` would produce a collision (about 1 in 457 Billion). I could see where it might make sense to combine the two tests, but one test better covers where there are very few named robots while the other covers where there are a very large number of named robots. Keeping the steps separate also encourages a bit more iterating over the solution.

As far as the test name being less tech-speaky goes, I'm open to suggestions on it. 'Collision' was a keyword used in the readme file so it seemed like a valid word choice.

kytrinyx:
I didn't realize _collision_ was jargon back when I wrote it. I personally think it's fine.

kotp:
Iit doesn't make sense to combine the two tests @callahat but I was asking if it makes sense to seed the random number generator in the other test as well.   Intermittent failures (even if extremely rare) is irritating and likely the wrong thing to do.

@kytrinyx I don't think collision is at all jargony in the file name, the test title itself seems to be tech-speaky, not reallyin  'voice agreement' with the other tests.  Such as "name sticks" and "different robots have different names".

callahat:
@kotp I think it would make sense to seed it in the other test as well.

Maybe `test_duplicate_name` or `different names when first choice is taken`?

kotp:
Your second one sounds good.  I think it leaves room for doubt about when the 'second choice' is also taken.

"duplicate_name_not_allowed" maybe concise, clear, explains what it is about the duplicate name that is being expected.

callahat:
What about `test_different_name_when_chosen_name_is_taken`?

kytrinyx:
I like that last one. The _when chosen name is taken_ has a very conversational feel.

kotp:
This passes the CI tests, but locally, when ran against `example.rb` it is failing.

@callahat Does this pass for you with the given `example.rb` file?

callahat:
It does not. It fails as it does not check if the generated name has been used previously.

kotp:
That means that I should check for why Travis is not failing.  <<-- @kytrinyx please also note.

@callahat it also means that you should update the example.rb solution, so that it behaves as it should.

I also double checked the README, just to make sure that it was still stated.  It states 'should' but for this track, we should be able to provide an example that exercises this.

kytrinyx:
> That means that I should check for why Travis is not failing.

Looking.

kytrinyx:
```
$ make test


The command "make test" exited with 0.
```

It's not running any of the tests. Probably because I forgot to change the Makefile when we moved the tests to `exercises`.

kytrinyx:
Alrighty, @callahat, if you rebase this onto the newest master and then push -f to the branch, CI should run the tests properly.

callahat:
@kytrinyx Done and green

kotp:
@callahat, again what do the seeds represent?  With the 20 million one are we using a random seed that produces the exact same name twice in a row?    What does srand 1000 give us?  Should we name those seeds?

If we can name those numbers so that they explain themselves, I think it is a good thing.

callahat:
The seeds don't represent anything in particular. A constant could be used. They are known values where the pseudo random should give good randomness. For srand 20mil+, this prevents an unlikely chance that a different seed would cause two robot names to be the same. For srand 1000, this ensures that the second robot name generated `name2` will initially be the same as the first robot name `name1`. In other words, the second time Robot.new.name is called the class will have to generate a second name internally since the first name it would generate would be the same as the first.

But probably having just one random seed as a constant would be a bit more clear.

kotp:
I like it... I was just responding to the other when I saw this change, to remind us that it does represent something and when it does we can name it.  You already identified this.

Should we them in a hash, maybe something like this:

``` ruby
ENFORCE_RANDOM_RESULT = {
  different_robot_name_seed: 1234,
  same_initial_robot_name_seed: 1000
}
```

callahat:
If there were more constants we were using across different topics maybe. `Kernel.srand ENFORCE_RANDOM_RESULT[:different_robot_name_seed]` seems a bit verbose for this use.

Cohen-Carlisle:
I noticed `example.rb` 'leaks' names when a robot is `reset`.
In other words, the name is not removed from from the list of known names.

kotp:
@callahat But it is in one area, still.  Though you do have a point... getting things done rather than perfection, right?

@Cohen-Carlisle that should be checked and a test written to guard against that from happening.  Thanks!

callahat:
@Cohen-Carlisle The readme may be slightly ambiguous on that. It states to not allow the same name twice, but doesn't call out what to do in the event of a reset (to let that name be used again or not). I would think it would still be out of the possible names as it was used once already (even if that robot were reset).

kotp:
This conversation seems familiar.  Unsure if it was on the site or in here.

Cohen-Carlisle:
I suppose since it is unclear, we should address it in another issue/PR instead of holding up this one :smile:

kotp:
Almost done now.

One last thing to do is to add a VERSION to the test and a matching on to the example.  You can see an example of this in the "food-chain" exercise.

And also squash your commits, please.  We can do it now using github's interface, but it is still 'good practice' to do so.  And Github is not the only hosting service in town, so you may need to know how to do this anyway.

callahat:
Done. Generally I avoid force pushing history modifications but in this case the risk of someone else working from my fork is acceptable.

Anything else?

kotp:
This looks much cleaner.  The version should be version 1, as the prior (without any version) is considered Version 0.

That last fix,, squash the commits one last time, and I will vote to merge!

callahat:
Updated. I'll open an issue for any other exercises which might not be versioned yet.

kotp:
No, that is fine.  Non-versioned are version 0, and it is known.  If, what you mean by that is, the existing ones that don't have version numbers.  We are going forward with versioning, not backward compliance.

kotp:
@calahat Thank you for this!

PR #278, opened 2016-04-07
add pangram exercise

Cohen-Carlisle:
I ruby-fied this after seeing the exercise on the python course.
Then I saw that the exercise existed in the common repo and generated the exercise that way.
Long story short, I decided to open this one anyway in case you prefer this hard coded one for any reason over the generated one.
However, you likely prefer the generated exercise, so feel free to close this PR if that is the case.


kotp:
Thanks @Cohen-Carlisle looks like you have been busy. :+1:

kotp:
This looks good, as does the other, with the (possible) exception of using two styles of quotes, rather than `%Q` or `%q` syntax for strings where both quote styles are used.

kytrinyx:
Niiice! I'd like to focus on the generated version, so I'm going to close this one.

PR #279, opened 2016-04-07
generate pangram exercise

Cohen-Carlisle:
I wasn't sure if I should be including the generated files, modifying `config.json`, or starting the version at 1 or 0.
Also, I hard coded single quotes around the `input` string read from the common repo... which kind of sucks as it could break later if someone includes an `input` string with apostrophes or special symbols.


kotp:
Why not use the `%q()` syntax to alleviate the worry?

Cohen-Carlisle:
Mostly I wanted to bring it up as a disclaimer of sorts.
I left `input` in single quotes because:
- its working now and YAGNI
- `%Q(...)` might be confusing to the novice Rubyist
- nothing trivial will be perfect:
  - `q%` doesn't handle escaped characters (newline, etc.)
  - `%Q()` could still fail if parens are mismatched in `input`

I also looked around the repo and it looks like everything else that is generated uses single quotes to delineate string literals.

That said, if there is a desire not to use single quotes, I'd be more than happy to change it up. Just let me know.

kytrinyx:
> I wasn't sure if I should be including the generated files, modifying config.json, or starting the version at 1 or 0.

I'd say: "yes" include the generated files, and "yes" modify the config.json, and start the version at 1.

The generated files are what get delivered when people say "exercism fetch", the generators themselves are just kept in the repo for our own benefit.

kytrinyx:
I _think_ that this is far enough along in the track that it's OK to use `%Q()`. I wonder if there's a way of using %Q in the generator, and then producing a correctly escaped string that uses double-quotes.

Cohen-Carlisle:
>I think that this is far enough along in the track that it's OK to use %Q(). I wonder if there's a way of using %Q in the generator, and then producing a correctly escaped string that uses double-quotes.

Sorry, I'm a bit confused. What do we want a user to see in the test file? It sounds like we are saying both `%Q(...)` and `"..."`.

kotp:
I think the tests should be idiomatic ruby, regardless of the skill level of the coder.  It doesn't matter if they have seen it yet or not, they will, and this project should be held to that standard.

The tests will, by behavior, validate the solution.  Feel free to use the `%Q` and related methods, they make sense and there is no reason not to expose them, especially if they should be used.

Cohen-Carlisle:
:+1: I will use `%Q(...)` syntax and put a small comment in to help alleviate any initial confusion.

kytrinyx:
> I think the tests should be idiomatic Ruby, regardless of the skill level of the coder. It doesn't matter if they have seen it yet or not, they will

Yes, this is a good point. I agree.

Cohen-Carlisle:
I appear to have angered rubocop :laughing:

kytrinyx:
I disagree with those cops. @kotp what do you think? Should we ease up on the rubocop there?

exercises/pangram/pangram_test.rb
```diff
@@ -0,0 +1,56 @@
...
+# 180638f Merge pull request #217 from ErikSchierboom/patch-2
+
+class PangramTest < Minitest::Test
+  # Below we use `%Q` syntax to create strings.
+  # This let's us worry less about escaping quotes.
+  # You can read more about Ruby's string literal syntax here:
+  # http://ruby-doc.org/core-2.3.0/doc/syntax/literals_rdoc.html#label-Strings
+
+  def test_sentence_empty
+    str = %Q()
```
kotp: Not appropriate here, you are just creating a string.  This should be str = '' or String.new

exercises/pangram/pangram_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    refute Pangram.is_pangram?(str)
+  end
+
+  def test_pangram_with_only_lower_case
+    skip
+    str = %Q(the quick brown fox jumps over the lazy dog)
```
kotp: Also not suitable here, simply create a string.

exercises/pangram/pangram_test.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def test_missing_character_x
+    skip
+    str = %Q(a quick movement of the enemy will jeopardize five gunboats)
+    refute Pangram.is_pangram?(str)
+  end
+
+  def test_pangram_with_mixed_case_and_punctuation
+    skip
+    str = %Q("Five quacking Zephyrs jolt my wax bed.")
```
kotp: Here it is appropriate.  Due to the actual use of double quotes.  This will preserve the actual double quote characters inside the string and requires no escaping.  However, likely more suitable with single quoted string, as we don't require interpolation.

exercises/pangram/pangram_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    assert Pangram.is_pangram?(str)
+  end
+
+  def test_pangram_with_non_ascii_characters
+    skip
+    str = %Q(Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich.)
```
kotp: Simply use quotes here as well.

kotp:
I agree with the cops, specifically, here we have no case where a string is using both single and double quotes in the string itself, and no string where interpolation is being used.

And so, take the cops advice, and use the syntax as shown, or simply use the 'simple' string in the example of a string that you want to include double quoted strings inside.

I prefer the use of the plain quoting in almost every case.  The exception is when you have a "String that may hold \" and ' characters inside".  Then this will use the `%q()` syntax, and when you have "a String that will do some #{calculations_and_interpolation} in conjunction with ' and \" characters" then you use the `%Q()` syntax.

Cohen-Carlisle:
@kotp The `%Q` is from the generator programmatically running each input from the `x-common` repo in an each loop.
I think we have miscommunicated around its use. Initially I had it using single quotes, but remarked this was being done blindly and could lead to errors later if single quotes are used in strings (from the `x-common repo`).
After some discussion, I thought it was preferred to use `%Q(...)` instead of `'...'` when programatically generating `pangram_tests.rb` as it was less likely to break on future inputs.
I actually agree that seeing the `%Q` here looks pretty odd. Should I revert those changes and go with only the changes in the first commit?

Cohen-Carlisle:
@kotp and I decided that going back to single quotes was a good solution. If single quotes become a problem we can fix that bridge if we come to it.

kytrinyx:
OK, thanks for taking the time to go back and forth on this so we could figure out how to best go about it.

kotp:
@Cohen-Carlisle was right to put the little party horn on there, as it was a pair programming party! :balloon:

Issue #280, opened 2016-04-07
Gemfile.lock


Cohen-Carlisle:
It seems odd to neither have it on the .gitignore nor in the repo.
Apologies if this has been discussed, but I did a quick search and turned up nothing.

kotp:
Thanks for bringing this to our attention.

I will check the history and see if this was discussed.

Since this is not a gem, perhaps it is something we should consider tracking.

Which way do you think it should go?

kotp:
I put it in the ignore for right now.  I am not solidly sold on it yet though.

This repository isn't deployed, it is only under development, and it isn't a gem.... It is driving me toward ignoring it.

It gets generated with Travis and when the `bin/local-status-check` executable is ran during development.

@kytrinyx what do you think? @kytrinyx ?

kytrinyx:
I don't have a strong preference either way, but I lean towards ignoring the lockfile since the dependencies are only needed for development (we've put the version of minitest in the require for the actual test suites so that people don't try to run it with 4.x).

Cohen-Carlisle:
Sounds reasonable. :+1:

kotp:
Thanks @Cohen-Carlisle and @kytrinyx

PR #281, opened 2016-04-07
Ignore only the projects ignored files

kotp:
Editor and Operating System related files are not "Project" related.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,4 @@
-*.swp
-*.beam
-.DS_Store
 .bundle
-.config
-.ruby-version
-tmp
 bin/configlet
 bin/configlet.exe
+Gemfile.lock


.gitignore
```diff
@@ -1,9 +1,4 @@
-*.swp
-*.beam
-.DS_Store
 .bundle
-.config
-.ruby-version
-tmp
 bin/configlet
 bin/configlet.exe
+Gemfile.lock
```
kotp: References #280

.gitignore
```diff
@@ -1,9 +1,4 @@
-*.swp
-*.beam
-.DS_Store
 .bundle
-.config
-.ruby-version
```
kotp: Removed this, as it deals with version managers, and we should not be concerned with what exact version they are running, just a minimum version as communicated in the Track readme.  We don't "enforce" it.

.gitignore
```diff
@@ -1,9 +1,4 @@
-*.swp
-*.beam
-.DS_Store
```
kotp: The above should be ignored, but in a "global" ignore file, usually located in the users home directory.

kytrinyx: Right, it *should* be, but often it is not, especially when dealing with newbies. Since a lot of newbies get involved here, I'd rather avoid having to ask them to remove files, squash commits, and push force to their branch.

On the other hand, the right thing to do is to coach them in the right way to do it, so maybe it's worth it.

.gitignore
kytrinyx: Ditto. We don't care, but a lot of people don't review their diffs, and accidentally commit things we don't care about.

kotp: This also goes to 'coaching', right?

kytrinyx: Yeah, it does.

kytrinyx:
It sounds like the way to go is a preference towards coaching. I think this is fine, as long as we're aware of it and have it as part of our mental checklist.

PR #285, opened 2016-04-09
Clock: Canonical test improvements

### LINKED ISSUE #282: clock: canonical test data has been improved
The JSON file containing canonical inputs/outputs for the Clock exercise has gotten new data.

There are two situations that the original data didn't account for:

- Sometimes people perform computation/mutation in the display method instead of in add. This means that you might have two copies of clock that are identical, and if you add 1440 minutes to one and 2880 minutes to the other, they display the same value but are not equal.
- Sometimes people only account for one adjustment in either direction, meaning that if you add 1,000,000 minutes, then the clock would not end up with a valid display time.

If this track has a generator for the Clock exercise, go ahead and regenerate it now. If it doesn't, then please verify the implementation of the test suite against the new data. If any cases are missing, they should be added.

See https://github.com/exercism/x-common/issues/166

kotp:
fixes #282
references exercism/x-common#166


exercises/clock/clock_test.rb
```diff
@@ -145,4 +145,81 @@ def test_wrap_around_backwards
...
+    clock1 = Clock.at(10, 37)
+    clock2 = Clock.at(10, 37) + days(1) + hours(1)
+    refute_equal clock1, clock2
+  end
+
+  def test_clocks_with_hour_overflow_by_several_days
+    skip
+    clock1 = Clock.at(3, 11)
+    clock2 = Clock.at(0, 11) + hours(99)
+    assert_equal clock1, clock2
```
kytrinyx: Is there a possibility here that people would handle overflow correctly in `+`, but not in `at`?

In other words, could this pass where equality with `Clock.at(0, 11+hours(99))` would fail?

kotp: With the current example implementation, no, because the `Clock.at(0, 11+hours(99))` will fail before you get to that point.  It would fail with an "Argument Error: argument out of range".

With a different implementation, definitely.

The tests as they stand works for the current example.rb file.  The example.rb is probably relying too much on Ruby built in time.  Coincidentally, it is implemented without dates, and so fits.

kytrinyx: The reason I added the tests in the first place was that in the Go track I was seeing bugs that the tests didn't catch; particularly clocks that were not equal.

I'd like to loosen up the restrictions on the example (which I think maybe you started working on) so that the tests catch this sort of thing if people take a different approach that we currently do in the example.

kotp:
Reworking `example.rb` and will continue that effort with a generator for these tests.

exercises/clock/clock_test.rb
kotp: Do we want to loosen it up or actually restrict use of Time, in the manner we did for Leap assignment?

kytrinyx: Let's just loosen it up. It makes for good conversations when people lean on Time.

kotp:
@kytrinyx Should we merge this, at least the tests are getting their information from x-common now.

We can continue on with the discussion about what the Clock should be doing when we add a time to it. Do we actually change the state of the clock like rotating that 'minus <---> plus' screw on the back of a wind up clock, or similar buttons on a digital clock, or is it simply meant to be a "What time would it be if I added this amount of time?"

lib/clock_cases.rb
```diff
@@ -0,0 +1,47 @@
...
+  def simple_test
+    "assert #{expected.inspect}, Clock.at(#{hour}, #{minute})#{add_to_clock}"
+  end
+
+  def add_to_clock
+    " + #{add}" if add
+  end
+
+  def skipped
+    index > 0 && 'skip' || '# skip'
```
Cohen-Carlisle: I don't know if I like the side effect of having a commented out skip.
Also the use of `&&` and `||` is bordering on overly-cute. You could replace both with a single ternary which is more readable, IMO.

lib/clock_cases.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+end
+
+require 'pp'
```
Cohen-Carlisle: This looks to be leftover from debugging?

kotp: Yes, I dreamt about this line actually last night.

lib/clock_cases.rb
```diff
@@ -0,0 +1,47 @@
+class ClockCase < OpenStruct
+  def name
+    'test_%s' % description.gsub(/[\(\) -]/, '_').gsub('=', 'is_equal_to')
+  end
+
+  def test_body
+    section == 'equal' &&
+      compare_clocks || simple_test
```
Cohen-Carlisle: Simpler to replace with ternary.

lib/clock_cases.rb
```diff
@@ -0,0 +1,47 @@
+class ClockCase < OpenStruct
+  def name
+    'test_%s' % description.gsub(/[\(\) -]/, '_').gsub('=', 'is_equal_to')
```
Cohen-Carlisle: FYI: You don't have to escape parens inside brackets of a regex.

Cohen-Carlisle:
Looks good aside from the `require` and minor stylistic differences.

exercises/clock/clock_test.rb
```diff
@@ -3,146 +3,300 @@
...
   def test_add_more_than_two_days
     skip
-    clock = Clock.at(1, 1) + days(2) + hours(10) + minutes(20)
-    assert_equal '11:21', clock.to_s
+    assert '11:21', Clock.at(1, 1) + 3500
+  end
+
+  def test_subtract_minutes
+    skip
+    assert '10:00', Clock.at(10, 3) + -3
```
Cohen-Carlisle: We aren't actually testing that we can subtract here (or with any subtraction)

exercises/clock/clock_test.rb
```diff
@@ -3,146 +3,300 @@
...
 
-  def test_subtrackt_more_than_two_hours_with_borrow
+  def test_subtract_more_than_two_hours_with_borrow
     skip
-    clock = Clock.at(6, 15) - hours(2) - minutes(40)
-    assert_equal '03:35', clock.to_s
+    assert '03:35', Clock.at(6, 15) + -160
+  end
+
+  def test_subtract_more_than_one_day__1500_min_is_equal_to_25_hrs_
```
Cohen-Carlisle: Method name has sequential underscores and an underscore at the end.

exercises/clock/clock_test.rb
Cohen-Carlisle: After checking on `x-common`, I retract my comment and think this is fine and within the spirit of the test.

exercises/clock/clock_test.rb
kotp: Yes, I need to open an x-common issue regarding comment like things in the tests.  I feel they aren't needed.  The fact that there are 25 hours in 1500 minutes is not helpful.  I may have already commented there, actually.

exercises/clock/clock_test.rb
```diff
@@ -3,146 +3,300 @@
...
+    assert_equal '00:20', (Clock.at(2, 20) + -3000).to_s
   end
 
-  def test_equivalent_clocks
+  def test_clocks_with_same_time
     skip
     clock1 = Clock.at(15, 37)
     clock2 = Clock.at(15, 37)
-    assert_equal clock1, clock2
+    assert clock1.to_s == clock2.to_s
```
Cohen-Carlisle: Should we be able to compare these directly, instead of their string comparisons?

Cohen-Carlisle:
I think I like it. Though its pretty late and my brain is hurting. I'd like to let this sit just a while longer and give it a final lookover tomorrow. This will also give @kytrinyx and/or @remcopeereboom a chance to look it over since there was a good bit of discussion in https://github.com/exercism/xruby/issues/282.

kotp:
Remember that the reason for this commit is mostly to get the x-common tests in play.  Changes in the example and how the tests drive the solution should be up to another action.

Cohen-Carlisle:
:+1: LGTM

PR #286, opened 2016-04-09
Add color option in tests

kotp:
ASSIGNMENT and ARGS documented, showing how to run a single test, and
how to activate pride, as one possible argument to minitest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@ Exercism Exercises in Ruby
 
 ## Setup
 
-You'll need a recent (1.9.3+) version of Ruby, but that's it. Minitest ships
+You'll need a recent (1.9.3+) version of Ruby, but that's it. Minitest
+ships
 with the language, so you're all set.
 
 ## Working on Test Suites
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,10 +28,35 @@ If you want to temporarily disable the skips while working on a test suite, you
...
-ruby -I../lib -rdisable_skip <filename_test.rb>
+```sh
+ruby -I../lib -rdisable_skip filename_test.rb
 ```
 
+In you would like to use the `make` tool to run a single test, while
+developing clock, for example, you can do something like this:
+
+```sh
+ASSIGNMENT=clock ARGS='-p' rerun make test-assignment
```
kytrinyx: Let's separate `rerun` out into its own section. That will make the README more scannable for the rerun case, while also making this description independent (people using rerun can figure it out, or we can describe it in the rerun section, but for people who aren't using it, they won't be confused by the addition).

README.md
```diff
@@ -28,10 +28,35 @@ If you want to temporarily disable the skips while working on a test suite, you
...
+
+Where, "ASSIGNMENT" is the variable that holds the name of the lesson,
+clock, and where I am using the `rerun` gem to monitor the folder
+structure for changes.
+
+ARGS is where you can put additional arguments, here I am demonstrating
+the `-p` argument which may give you 'pride' output.
+
+If you use zsh, you can use this function to make this process simple,
+shown here with pride asked for.
```
kytrinyx: I don't know that "with pride asked for" means. Is it an American idiom?

kytrinyx: Oh. `pride`. That wasn't clear--can we try rephrasing it for clarity?

README.md
kotp: I was thinking about removing it altogether, since you can use any other thing that you may use for continuous testing.

kytrinyx: Yeah, that's a better option.

kotp:
Updated the language and included other examples showing the options we can run the tests with.

README.md
```diff
@@ -24,14 +24,51 @@ Run the test with `ruby path/to/the_test.rb`.
...
+clock.
+
+ARGS is where you can put additional arguments, here I am demonstrating
+the `-p` argument which may give you 'pride' output.
 
-```ruby
-ruby -I../lib -rdisable_skip <filename_test.rb>
+If you use zsh, you can use the following function to make this process
+simple, shown here using the additional argument `-p` which may run your
+tests with color from the pride reporter with minitest.
```
kytrinyx: Since you show the command with two different examples of arguments later, I think this can be shortened. Maybe

```
If you use zsh, you can use the following function to make this process simple.
```

PR #287, opened 2016-04-11
set execution bit on bin/*

Cohen-Carlisle:


kotp:
Thanks @Cohen-Carlisle the bin folder by definition should contain executable files, I think.

PR #288, opened 2016-04-11
generate run-length-encoding

Cohen-Carlisle:
TODO: implement #decode in example.rb
      find place in config.json
      why isn't execution bit set on generated tests?


kotp:
The the tool in bin to set the executable bit.  Let me know if it doesn't do the right thing.

Cohen-Carlisle:
> implement #decode in example.rb

Working on this.
> find place in config.json

Any thoughts on this?
> why isn't execution bit set on generated tests?

Addressed by https://github.com/exercism/xruby/pull/289

Cohen-Carlisle:
Something seems incorrectly configured... had a similar issue on my pre-push checks... they are not finding `example.rb`.

Other than that, any thoughts on where this test should live in the track/config.json?

kotp:
You are missing the last part of the filen ame you are wanting to require... `require_relative 'run_length_encoding'` should fix that problem.

Cohen-Carlisle:
Updated.
Still wondering where a good place for this exercise is. I placed it where I thought it would fit, but it was kind of a wild guess :laughing:
If anyone has any thoughts let me know, I'll change it if need be and then squash.

kotp:
Generally speaking 'do' is one of the (around 42) reserved bare words in Ruby.  Should we be in the habit of using it?

Cohen-Carlisle:
Yeah, that is probably a bad idea. `do` blows up without an explicit receiver.
I used it to make [this line](https://github.com/exercism/xruby/pull/288/files#diff-7ce21b5a593c9baff7bfd826087b1ff34R42) <= 80 chars, as `encode` and `decode` were too big and splitting it across multiple lines would require some reworking of the generator files and I'm lazy :disappointed:

kotp:
The 80 is a guide, of course.  I also am not sure which line, your link is not helpful as for some reason it did not highlight the line as I thought it would.

I noticed there are other files that define a method do.  Hamming is used as the example currently for those that want to work on creating the generators.  I think it is the wrong thing to do (no pun intended.).  Until I understand the generators a little bit better, though, I have no alternative suggestion.

kotp:
Now that I have a little better understanding, I am using 'actual' in place of 'do'.

@kytrinyx I think that is a better representation of what we want to communicate.  It may also help with the stroop effect of being in ERB at that level.

kotp:
And now that it is all together, I am using something else. :)  But I did avoid 'do' as a method name.  And made the `erb` simpler to understand as well.

Cohen-Carlisle:
More and hopefully final changes.

kotp:
Looks good to me, version numbers are right, reads well, etc.  Thanks @Cohen-Carlisle!

PR #293, opened 2016-04-13
Update tests to follow x-common/sieve.json

DebbieGillespie:
Now that there is a common set of tests defined in the
`x-common/sieve.json` file, updating the ruby tests for sieve so they
implement all the tests that are defined. I had to modify the example so
that it would correctly handle the "no primes under two" case.


kotp:
Thank you @DebbieGillespie

PR #294, opened 2016-04-14
Extend pangram test suite

victorbrender:
Some implementations use the "/\w/" regex to match characters and add them to an array and then check that the size of the array is equal to 26. These implementations erroneously pass the current test suite.

### DIFF
--- a/exercises/pangram/pangram_test.rb
+++ b/exercises/pangram/pangram_test.rb
@@ -19,6 +19,16 @@ class PangramTest < Minitest::Test
     assert Pangram.is_pangram?(str)
   end
 
+  def test_pangram_with_underscore_character
+    str = 'the quick brown_fox jumps over the lazy dog'
+    assert Pangram.is_pangram?(str)
+  end
+
+  def test_pangram_with_sentence_containing_numbers
+    str = 'the qu1ck br0wn fox jumps over the lazy dog'
+    refute Pangram.is_pangram?(str)
+  end
+
   def test_missing_character_x
     skip
     str = 'a quick movement of the enemy will jeopardize five gunboats'


Cohen-Carlisle:
Thanks for pointing this out!

This exercise is generated from common metadata, so the ultimate fix is to update that metadata. I'd like to get the metadata fixed and then regenerate this exercise instead of doing a one off fix here. If you are up for it, you could open an issue or a PR on the [x-common](https://github.com/exercism/x-common) repo.

Should the test's name be a bit more clear about exactly what its testing? For example, would `test_pangram_with_underscore_word_separators` be better than `test_pangram_with_underscore_character`? I'm leaning towards a yes, but I'm not sure.

The name `test_pangram_with_sentence_containing_numbers` shouldn't say pangram as the input in this case isn't. Probably something along the lines of `test_letters_replaced_with_numbers`.

exercises/pangram/pangram_test.rb
```diff
@@ -19,6 +19,16 @@ def test_pangram_with_only_lower_case
     assert Pangram.is_pangram?(str)
   end
 
+  def test_pangram_with_underscore_character
+    str = 'the quick brown_fox jumps over the lazy dog'
+    assert Pangram.is_pangram?(str)
+  end
+
+  def test_pangram_with_sentence_containing_numbers
+    str = 'the qu1ck br0wn fox jumps over the lazy dog'
```
kotp: Avoid 0 and 1 as they are too close in some fonts to l and O.  When possible, use distinctively shaped numbers, rather than the possibly visually ambiguous ones.

This becomes aggravated especially when used in places where those ambiguities are replacements of letters that are so similar.

Cohen-Carlisle: The purpose of this test is to replace the letters with a corresponding number and make sure the solution does not think the string is a pangram. With this in mind, I think its ok. Especially if the method name were updated to something like `test_letters_replaced_with_numbers`.

At worst its an exercise in paying attention to detail, which is a good skill to have as a programmer.

kotp: Is that the purpose of the test?  

I don't think it is, and if so, it is not expressed clearly.

It appears that the purpose is that if there are 26 unique characters and yet it is missing the only letter this is missing, which happens to be an 'i', it should not be classified as a pangram.  

For intent, though, we would definitely need to ask @victorbrender since he is the only one that can confirm intent, as the author.

The visual distraction of similarly shaped numbers in the place of letters is simply a distraction.  Noting that the 'i' is the only thing missing is the exercise of detail attention that you speak of.  It happens here regardless of the visual distraction of the numbers similarity to the letters they are replacing.

kotp:
This is what I came up with that addresses test naming and the potential for misunderstanding the intent of the PR.

```ruby
def test_pangram_with_underscore_character
  str = 'the quick brown_fox jumps over the lazy dog'
  assert Pangram.is_pangram?(str)
end

def test_not_a_pangram_with_26_unique_characters
  str = '0123456789klmnOPQRstuvWXYz'
  refute Pangram.is_pangram?(str)
end

def test_pangram_with_string_containing_numbers_which_is_otherwise_valid
  str = '0123456789abcdefghijklmnOPQRstuvWXYz'
  assert Pangram.is_pangram?(str)
end
```

@victorbrender what do you think?  Does it address what your concerns are?  If so, can you update your PR?

Cohen-Carlisle:
I took the liberty of creating https://github.com/exercism/x-common/pull/228 to fix this for all languages, as well as address another gap in testing noted there.
@kotp @victorbrender input welcome

kotp:
also exercism/x-common#222

kytrinyx:
FYI @kotp and @Cohen-Carlisle I thought all track maintainers had commit on the x-common repository, but I failed at UX, and so everyone had "read" access. Great. On a public repo :) Anyway, this is fixed, and if y'all are happy with exercism/x-common#22, then either of you could go ahead and merge that.

kotp:
Thanks @victorbrender for the help in getting this worked out.  Please do let us know if we missed the mark.

PR #298, opened 2016-04-18
Giving the users the power of Skips

### LINKED ISSUE #297: Exercism Nitpickers POV - Remove skips?
I personally find the `skip`s a bit annoying as they need to be removed over and over.
I understand the idea is not to overwhelm the user with errors... it seems like a nice idea but gets old after a while.
If I only want to work on one case, I can just solve the first failure and rerun!
A middle ground could be to have them on the first `n` exercises and not after that.
Thoughts?

kotp:
We give the clients the nicety of not flooding them with errors and
failures when they fetch the lessons, but with this, we inform them how
they can control the skip behaviors, and how they can test individually
on demand.
fixes #297

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -7,4 +7,81 @@ $ ruby hamming_test.rb
 ```
 
 Only the first test will be executed, all the others have been made pending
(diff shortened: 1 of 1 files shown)


kytrinyx:
This is great!

Issue #258, opened 2016-02-28
Should sieve come before nth-prime?


Aqualon:
Solving the nth-prime task involves some kind of Prime-number generator, which has to be implemented in the sieve exercise at a later point. Wouldn't it make sense to first give sieve to users and then let them find out in nth-prime how to use the sieve to solve it?

kytrinyx:
Sure. There are other ways than `sieve` to write a prime number generator, but you make a great argument.

kotp:
Let's finish talking about this, and let someone get this in place.

I say we can move sieve to happen before nth-prime.  Perhaps between pangram and roman-numerals.

That will leave a couple of exercises between sieve and nth-prime, and break them up between "mathy" exercises.

To patch this is simple.  The hard part is figuring out the best place to put it.  So it is perfect for first-timers-only.  If someone needs help with this, it won't be with "coding" just with the process of contributions.

So what is needed to do now is simply OK the placement... and let someone grab this and do it.

kytrinyx:
I agree we should have a break between "mathy" exercises (same as with "lots-of-text-y" exercises).

I think somewhere between pangram and roman numerals sounds good.

kotp:
OK, so I think this is now set up to be an awesome first patch for someone!  It isn't high priority, and it is easy to do, and it has the idea of how to do this in this set of comments...

Who wants it? :balloon:

mostlybadfly:
Hello all, I'd love to do this.  I see that config.json seems to have the order of the exercises, is the best place to make the change?  I'll take a look in a couple of hours when I'm on lunch and can use my laptop :)

kotp:
@mostlybadfly great to see you here!

If you do have any questions don't hesitate to ask.  

I think you did find the place to make the change.

mostlybadfly:
@kotp thanks! submitted the PR.  Is there a list of all issues I can go through to find some beginner friendly ones? Or is it best to just peruse all the repos in the organization?

petertseng:
@mostlybadfly Ah, it sounds like you are looking for something organisation-wide? I can help out with that: https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+user%3Aexercism+label%3A%22good+first+patch%22

(Hmm @kytrinyx  should we throw a link to this search query somewhere where people can see?)

[...]

PR #302, opened 2016-04-20
Placed sieve between pangram and roman numerals.

mostlybadfly:
Hello! this is in reference to #258. I have changed config.json to place the sieve exercise between pangram and roman numerals.


kotp:
Looks good.

PR #304, opened 2016-04-26
Add Bowling exercise

bernardoamc:


exercises/bowling/example.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def open_frame?
+    !strike_frame? && !spare_frame?
+  end
+
+  def adjust_frames
+    if @frame_buffer.size == 3 && strike_frame?
+      @frame_scores << @frame_buffer.reduce(:+)
+      @frame_buffer.shift
+    elsif (@frame_buffer.size == 3 && spare_frame?)
```
repinel: You could remove the parenthesis 😉

bernardoamc: 👍 ❤️

exercises/bowling/example.rb
```diff
@@ -0,0 +1,66 @@
...
+  def score
+    raise GameInProgress unless @frame_scores.size == TOTAL_FRAMES
+    @frame_scores.reduce(:+)
+  end
+
+  private
+
+  def validate_roll(pins)
+    raise InvalidRoll if pins < MINIMUM_ROLL_SCORE
+    raise InvalidRoll if pins > MAXIMUM_ROLL_SCORE
```
kotp: Consider passing good messages to the error class when raised.
This could also benefit from the spaceship operator.  One good message to rule both cases.

bernardoamc: Makes total sense. 👍

bernardoamc:
@kotp There is [this](https://gist.github.com/bernardoamc/8f073ed34f8089fde79956f775be9ef1) alternative solution that extracts the concept of a `Frame` in its own class. Which solution do you prefer?

kotp:
I don't think it matters so much.  As long as both versions pass the reference test.  And that the test doesn't drive the way the solution comes about, if possible, it is all good.

The example.rb is actually just there for a proof of concept, that it is solvable, and doesn't have to be "perfect" in solving the problem.  So even the feedback here I think is just suggestive.

It is also why we don't "rubocop" the example.rb file.  The tests should be "good examples of coding" because they are public facing.  The examples are more for the designers use to proof the tests.

That being said.  How do you feel about both versions?  Maybe even better... let's get this up and running and then submit both as solutions for the masses to give feedback?

exercises/bowling/example.rb
```diff
@@ -0,0 +1,66 @@
...
+    elsif @frame_buffer.size == 3 && spare_frame?
+      @frame_scores << @frame_buffer.reduce(:+)
+      @frame_buffer.shift(2)
+    end
+
+    if @frame_scores.size < 10 && @frame_buffer.size == 2 && open_frame?
+      @frame_scores << @frame_buffer.reduce(:+)
+      @frame_buffer.shift(2)
+    end
+  end
```
kotp: This is example. so doesn't necessarily have to be exemplary.

But... If these two sections were named based on what they are doing or what the specific condition is that this happens in, it may be easier to read/understand.

```ruby
  def adjust_frames
    if @frame_buffer.size == 3 && (strike_frame? || spare_frame?)
      @frame_scores << @frame_buffer.reduce(:+)
      @frame_buffer.shift(strike_frame? ? 1 : 2)
    end

    if @frame_scores.size < 10 && @frame_buffer.size == 2 && open_frame?
      @frame_scores << @frame_buffer.reduce(:+)
      @frame_buffer.shift(2)
    end
  end
```

exercises/bowling/example.rb
bernardoamc: I added default messages in the constructor, what do you think?

bernardoamc:
@kotp I see, given what you said I think this solution is more intuitive and serves its purpose. That being said, it is always a good thing to identify hidden concepts in you code and make it explicit in the long run. Since this is a "closed problem" (will not evolve in complexity) I think this version is ok. 

I changed the `adjust_frame` method so it becomes more readable, what do you think?

exercises/bowling/example.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def validate_roll(pins)
+    raise InvalidRoll if pins < MINIMUM_ROLL_SCORE
+    raise InvalidRoll if pins > MAXIMUM_ROLL_SCORE
+
+    if second_roll_in_frame?
+      raise FramePinsExceeded if @frame_buffer[0] + pins > MAXIMUM_ROLL_SCORE
+    end
+
+    raise GameHasEnded if @frame_scores.size == TOTAL_FRAMES
```
kotp: This is the second time I have seen the `@frame_scores.size == TOTAL_FRAMES`so should this be named something like "game_over?"

exercises/bowling/example.rb
kotp: I like it.

This is how I refactored this method...

```ruby
  def validate_roll(pins)
    (MINIMUM_ROLL_SCORE..MAXIMUM_ROLL_SCORE).cover? pins or fail InvalidRoll
    if second_roll_in_frame?
      fail FramePinsExceeded if @frame_buffer[0] + pins > MAXIMUM_ROLL_SCORE
    end

    fail GameHasEnded if game_over?
  end
```

exercises/bowling/example.rb
```diff
@@ -0,0 +1,105 @@
...
+    @frame_buffer[0] == MAXIMUM_ROLL_SCORE
+  end
+
+  def spare_frame?
+    @frame_buffer[0] + @frame_buffer[1] == MAXIMUM_ROLL_SCORE
+  end
+
+  def open_frame?
+    !strike_frame? && !spare_frame?
+  end
```
kotp: Is this better?

```ruby
  def open_frame?
    not strike_frame? || spare_frame?
  end
```

exercises/bowling/example.rb
```diff
@@ -0,0 +1,105 @@
...
+  def adjust_frames
+    if strike_can_be_scored?
+      score_strike
+    elsif spare_can_be_scored?
+      score_spare
+    end
+
+    if open_frame_can_be_scored?
+      score_open_frame
+    end
```
kotp: This seems awkward...

```ruby
  def adjust_frames
    score_strike if strike_can_be_scored?
    score_spare if spare_can_be_scored?
    score_open_frame if open_frame_can_be_scored?
  end
```

exercises/bowling/example.rb
bernardoamc: Great idea. 👍

exercises/bowling/example.rb
bernardoamc: I think `!strike_frame? && !spare_frame?` reads better in this case.

exercises/bowling/example.rb
bernardoamc: After the method was refactored the name `adjust_frames` algo got strange. Maybe something like `adjust_frame_scores` using this implementation is better?

exercises/bowling/example.rb
bernardoamc: I like the use of `fail` here, but I prefer the `if` statements instead of the `.cover?`. If you really think this reads better, I have no problem in changing this. :)

bernardoamc:
@kotp Thank you for the suggestions, the code is way more explicit now. ❤️

exercises/bowling/example.rb
kotp: I have somewhat of a difficult time seeing the single sigil, it is so skinny. but yeah.

I think reading it in your form says "not strike frame and not spare frame", and in my form it says "not strike frame nor spare frame"

I am not sure how most people would say this in an instructional setting.

kotp:
If no one else has other comments, I think this is good to put into place.

exercises/bowling/example.rb
```diff
@@ -0,0 +1,90 @@
+class Game
+  VERSION = 1
+  RULES = { MIN: 0, MAX: 10 }.freeze
+  at_exit { private :score_frame, :validate }
```
kotp: Aren't there others that should be private?

kotp:
So this takes out the custom errors, and utilizes the RuntimeError messages to give better hints as to what is the idea of what is wrong in the journey through the tests.  Better than directing the custom errors in this exercise, I think.

Major rewrite, some very nice improvements.

Are the tests in logical order or do you have some rearranging to do yet?  If not, I :+!: we merge it in.

bernardoamc:
So, I removed the tests that forced the exercist to have methods that were not directly specified by the exercise `README` and the order seems pretty good now.

kotp:
:LGTM:

It can be merged in, and ready to roll. (yes, pun intended!)

@exercism/ruby have a quick look and vote?

exercises/bowling/bowling_test.rb
```diff
@@ -0,0 +1,233 @@
...
+    end
+  end
+  private :roll_n_times
+
+  # Problems in exercism evolve over time,
+  # as we find better ways to ask questions.
+  # The version number refers to the version of the problem you solved,
+  # not your solution.
+  #
+  # Define a constant named VERSION inside of Leap.
```
kytrinyx: This should be Game. But actually, since this isn't one of the first exercises, let's simplify this to one line:

```
# Don't forget to define a constant VERSION inside of Game. 
```

bernardoamc: Oops 😓   Agreed, going to change this. 👍

bernardoamc:
@kytrinyx Thanks for noticing the mistake, fixed.

kytrinyx:
This looks good to me. Thank you :heart:

kotp:
Awesome catch, @kytrinyx :blush:

PR #306, opened 2016-04-30
Add two-bucket exercise

repinel:


Cohen-Carlisle:
Hi @repinel, thanks for this PR!
Would you be interested in writing a generator that consumes [this metadata](https://github.com/exercism/x-common/blob/master/two-bucket.json) and generates the test file?
The metadata is there to keep problems consistent between all the different language tracks.
Another option that would still be a plus would be to manually modify the the test file to be in sync with that metadata.
Let us know if either of those sound like something you'd be interested in!

repinel:
@Cohen-Carlisle I added the generator and updated the PR. Thanks for bringing it up!

lib/two_bucket_cases.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def skipped?
+    index > 0
+  end
+end
+
+TwoBucketCases = proc do |data|
+  cases = []
+  JSON.parse(data)['two_bucket']['cases'].each_with_index do |row, i|
```
kytrinyx: I just realized that we could probably use `map.with_index` here instead of `each_with_index`, which would get rid of the `cases` temporary variable.

This is totally not a big deal, and I don't want to hold this back just for this tiny thing.

kytrinyx:
This looks good to me. Thanks, @repinel!

repinel:
@kytrinyx Thanks for the suggestion. I added the `map.with_index` to `lib/two_bucket_cases.rb`.

kytrinyx:
This looks great. I'm going to leave it for one of the others to merge, to give them a chance to catch things I might have missed.

Cohen-Carlisle:
I think it looks good, but I'll be honest, I haven't entirely wrapped my head around the exercise...
(and I kinda don't wanna spoil it for myself :laughing:)
Any input @kotp?

kotp:
Spent some time and looked at this on my machine... 

Any opinions on the actual placement for this exercise?  @repinel any reasoning for why it is in the `config.json` at this position?

repinel:
@kotp no especial reason

kotp:
Any better placement for this?

kytrinyx:
Good call, @kotp. I agree, closer to minesweeper than to ocr-numbers.

repinel:
I can change it 😉

repinel:
@kytrinyx @kotp I updated the `config.json`

kytrinyx:
Nice! I think this is good to go. @kotp want to do the honors?

kotp:
Take a :bow: @repinel Now to fetch it and solve it. :)

repinel:
Thanks!

kytrinyx:
Wait, let me deploy the x-api first!

kytrinyx:
... deployed.

Thank you :heart:

PR #307, opened 2016-05-01
add bracket-push tests generation

moofkit:
add generator for bracket-push problem to using [shared tests](https://github.com/exercism/x-common)

Should I change the VERSION from 1 to 2?


kytrinyx:
Yeah, I think this should be version 2 since it adds some tests which could potentially be backwards incompatible.

exercises/bracket-push/bracket_push_test.rb
```diff
@@ -1,43 +1,81 @@
...
+  def test_math_expression
+    skip
+    assert Brackets.paired?('(((185 + 223.85) * 15) - 543)/2')
+  end
+
+  def test_complex_latex_expression
+    skip
+    # rubocop:disable LineLength
+    assert Brackets.paired?('\left(\begin{array}{cc} \frac{1}{3} & x\\ \mathrm{e}^{x} &... x^2 \end{array}\right)')
+    # rubocop:enabled LineLength
```
Cohen-Carlisle: I don't think rubocop details should be present in our test files. We could assign it to a variable with a short name like `str` instead. Unfortunately, the string is so long it might still trigger rubocop. I hate to say it... but maybe we should shorten it up in xcommon if that's the case.

We might also consider adding a check on input length in x-common itself (cc: @kotp @kytrinyx)

kytrinyx: Uh. Yeah, I don't like having rubocop stuff in the test file.

I also don't want to shorten this in x-common. This is a very useful real-world use-case that easily has bugs, if I recall correctly.

How about disabling linelength for this file at the `.rubocop.yml` level?

Cohen-Carlisle: My thought was that we could still have a `LaTex` string but it would be shortened down so it would fit on a single line. Depends on how much we care about line length vs keeping everything in this example.

kytrinyx: I'm good with a shorter LaTeX string, provided that it covers the edge cases that it was added to cover.

moofkit:
update VERSION.
@kytrinyx @Cohen-Carlisle how's about that one?

exercises/bracket-push/example.tt
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env ruby
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'bracket_push'
+
+# Test data version:
+# <%= sha1 %>
+class BracketsTest < Minitest::Test<% test_cases.each do |test_case| %>
+  def <%= test_case.name %><% if test_case.skipped? %>
```
kotp: Should we avoid having in the 'view'?  See `xruby/clock/example.tt` along with `lib/clock_cases.rb` with the `skipped` method.

exercises/bracket-push/bracket_push_test.rb
kotp: There should be no reason for the rubocop directive nor for changing the latex, nor for having a length greater than 80 for this situation.  Two strings laid out side by side concatenate naturally.  Ignoring a line feed between should take care of the length.

exercises/bracket-push/bracket_push_test.rb
```diff
@@ -1,45 +1,94 @@
...
   end
 
-  def test_wrong_nested
+  def test_complex_latex_expression
     skip
-    refute Brackets.paired?('[({]})')
+    str = '\left(\begin{array}{cc} \frac{1}{3} & x\\ '\
+          '\mathrm{e}^{x} &... x^2 \end{array}\right)'
+    assert Brackets.paired?(str)
   end
```
kotp: Yes!  That! :100:

exercises/bracket-push/example.tt
```diff
@@ -0,0 +1,30 @@
...
+    skip<% end %><% if test_case.long_input? %>
+    str = '<%= test_case.splited_input[0] %>'\
+          '<%= test_case.splited_input[1] %>'<% else %>
+    str = '<%= test_case.input %>'<% end %><% if test_case.expected %>
```
kotp: Similar to 'skipped' solution for this.  Keep this simple if possible.

moofkit:
fix test template according @kotp recommendations. It's really looks simpler. Like it

exercises/bracket-push/example.tt
kotp: Can we work toward something like this:

```ruby
class ClockTest < Minitest::Test<% test_cases.each do |test_case| %>
  def <%= test_case.name %>
    <%= test_case.skipped %>
    <%= test_case.test_body %>
  end
<% end %>
```

works with this method in `clock_generator.rb`

```ruby
  def skipped
  end
```

That template erb file (.tt) is working with a JSON that has multiple test categories.  But the complexity has been removed from the template.

kotp:
Looks like my comments are about 180 degrees out of phase with your updates. :)  Shaping up well.

lib/bracket_push_cases.rb
```diff
@@ -0,0 +1,41 @@
...
+  def name
+    'test_%s' % description.gsub(/[ -]/, '_')
+  end
+
+  def skipped
+    index.zero? ? '# skip' : 'skip'
+  end
+
+  def test_body
+    long_input? ? splited_test : simple_test
```
kytrinyx: This is a purely English thing: it's called "split" not "splited". `split_test`

lib/bracket_push_cases.rb
```diff
@@ -0,0 +1,41 @@
...
+    input.length > 80
+  end
+
+  def simple_test
+    "#{assert_or_refute} Brackets.paired?('#{input}')"
+  end
+
+  def splited_test
+    "str = '#{splited_input[0]}'\\
+          '#{splited_input[1]}'
```
kytrinyx: ditto: `split` rather than `splited`

Cohen-Carlisle:
I'm late to the party but good job @moofkit! :tada:

PR #310, opened 2016-05-02
add poker

fabienalbi:
Hi !

I am working on translating this exercise :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -61,6 +61,7 @@
     "simple-cipher",
     "wordy",
     "allergies",
+    "poker",
     "kindergarten-garden",
     "linked-list",
     "pythagorean-triplet",
--- a/exercises/poker/example.rb
+++ b/exercises/poker/example.rb
@@ -1,0 +1,78 @@
+class Poker
+  VERSION = 1
+
+  def initialize(hands)
+    @hands = hands
+  end
+
+  def best_hand
+    best_score = @hands.map { |hand| hand_rank(hand) }.max
+    @hands.select { |hand| hand_rank(hand) == best_score }
+  end
+
+  private
+
+  def hand_rank(hand)
+    hand_parser(hand)
+    top_hand || low_hand
+  end
+
+  def hand_parser(hand)
+    @card_ranks = hand.map { |ele| '..23456789TJQKA'.index(ele[0]) }
+    @card_suits = hand.map { |ele| '.HSDC'.index(ele[1]) }
+    @card_split = @card_ranks.each_with_object(Hash.new(0)) { |r, h| h[r] += 1 }
+    @groups = @card_split.values.sort.reverse
+  end
+
+  def top_hand
+    straight_flush || square || full || flush || straight
+  end
+
+  def low_hand
+    three_of_a_kind || double_pair || pair || 0 + @card_ranks.max
+  end
+
+  def straight_flush
+    800 + @card_ranks.max if straight? && flush?
+  end
+
+  def square
+    700 + @card_split.key(4) if @groups == [4, 1]
+  end
+
+  def full
+    600 + @card_split.key(3) if @groups == [3, 2]
+  end
+
+  def flush
+    500 + @card_ranks.max if flush?
+  end
+
+  def straight
+    400 + @card_ranks.max if straight?
+  end
+
+  def three_of_a_kind
+    300 + @card_split.key(3) if @groups == [3, 1, 1]
+  end
+
+  def double_pair
+    if @groups == [2, 2, 1]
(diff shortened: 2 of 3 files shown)


bernardoamc:
Thanks @fabienalbi! Do you plan to add more test cases regarding the high card rule? For example `nothing vs nothing`, the winner being the one with the higher card.

Cohen-Carlisle:
Thanks @fabienalbi. Are you still in the process of working on this? I see that there is still a bit of work to do.

It looks like you still need to make the test file executable.
The example solution should be named `example.rb`.
And the exercise will need added at an appropriate place in the `config.json` file.

Thanks for your work so far!

fabienalbi:
@Cohen-Carlisle Hey ! Sorry yes, it's been a busy week but I am still working on it.
Thanks for the information :)

kotp:
In the bin folder you will find a couple of files in there that may be helpful for development on this, the one that I am specifically think of is the setup hooks executable there.

Look at it before you run it, of course, but it will set up a pre-push hook that will help to alert you before pushing your code that Travis CI may fail.

It just helps you to have your ducks in a row before pushing up.  There is also a way to push anyway.  If you prefer not to use the git hooks though, you can run the one that does a local check, manually.

fabienalbi:
Thank you @kotp ! I missed a step, I didn't know how to make the test executable :)

kotp:
744 is likely more appropriate than 755 for the mode changes.  I don't appreciate having the world be able to write to my files.

fabienalbi:
I am not sure how the 755 mode has been chosen and how to change it. Also, I don't know what this means really

kotp:
It is fine, it is just a note.  I or someone will look into this more at some point.  You are all good.

kytrinyx:
This looks good to me, with the reservation that I've not thought a whole lot about how poker works, so I don't have any juicy opinions about the implementation.

Side note: here's a great video of a talk that I love that has a section about implementing poker... http://confreaks.tv/videos/bathruby2015-a-lever-for-the-mind

bernardoamc:
Nice @fabienalbi! ❤️  

Should we use the highest card rule for cases like `pair x pair` when the pairs are the same? This rule is pretty tricky since you have to compare up to the last card in case of consecutive draws.

fabienalbi:
Very good point ! I haven't thought about this scenario... I am going to iron out the tests, check for every possibilities and adjust the example solution accordingly
Thanks for the video @kytrinyx :) It'll help !

fabienalbi:
There are a lot of edge cases neither I nor the python track thought about 😕 
The video, albeit showing useful leads, does not go into much details and use techniques I am not yet comfortable with. Also, it seems to me that it wouldn't work when comparing more than two hands.
The solution I came up with works fine for 99% of the cases but has to be fully rethink to reach 100%.
It seems like I am at a dead end for now. Sorry all

kotp:
@fabienalbi That just means, 'Good news, everyone!', you have reached the point where 'good enough' is known.

Now ,if everything else is in place, we have a life line.  We merge the the PR, and as it gets solved over time, the discussions help bring out the 'edge-cases' that are needed.  And with those discussions, the pieces to fill it out in the form of 'versions' of tests.

In other words, if you have gotten to 99%, you have likely exceeded the usable point, which is usually nearer the 80% mark.

Let's get this approved and out there!

fabienalbi:
All right, sounds good to me. I am looking forward to see the solutions that the users will come up with. Hopefully I'll quickly learn and find ways to iron out the exercise :)

kotp:
Let's add a commented skip in the first test.

And lets squash the commits.

fabienalbi:
All right, commented skip added and commits squashed. It seems like it's ready for review 👍

kotp:
Sent you a patch with suggestions @fabienalbi so should be good to go shortly.

fabienalbi:
@kotp I am not familiar with patches. I received the e-mail with diffs, should I implement them, squash the new commit and re-push ?

kotp:
Yes, you can view the diffs, effect them in the best way that you know how.  You can also e-mail me and if you would like we can work on it together, synchronously.

fabienalbi:
I wonder about the flatten part of the patch. How is it better/clearer to use flatten on the best_hand method rather than having the array brackets on the expected value ? Especially since it can't be used with the final test 'test_three_hand_with_tie' making the assertions syntaxed inconsistent ?

kotp:
@fabienalbi your feeling mirrors mine on that... it was strange to review... You are finger deep in that, and some brain matter maybe swimming in there too, though... do you have a suggestion on how to make that clearer?  I am interested as well.

PR #311, opened 2016-05-04
Add duck-type tests to sieve_test.rb

esquinas:
Added some test to introduce duck-typing and some error handling when bad arguments are passed.

### DIFF
--- a/exercises/sieve/example.rb
+++ b/exercises/sieve/example.rb
@@ -1,6 +1,11 @@
 class Sieve
   attr_reader :range
   def initialize(limit)
+    begin
+      limit = limit.to_i
+    rescue NoMethodError 
+      raise ArgumentError, "Limit not convertible to an integer (#{limit})."
+    end
     @range = (2..limit)
   end
 
--- a/exercises/sieve/sieve_test.rb
+++ b/exercises/sieve/sieve_test.rb
@@ -53,7 +53,7 @@ class SieveTest < Minitest::Test
   def test_limit_is_not_duck_type_valid
     skip
     expected = []
-    assert_equal expected, Sieve.new('I cannot be converted to an integer').primes
+    assert_equal expected, Sieve.new('Cannot convert to integer').primes
   end
 
   def test_limit_is_duck_type_valid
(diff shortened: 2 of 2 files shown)


exercises/sieve/example.rb
```diff
@@ -1,6 +1,11 @@
 class Sieve
   attr_reader :range
   def initialize(limit)
+    begin
+      limit = limit.to_i
+    rescue NoMethodError 
+      raise ArgumentError, "Limit not convertible to an integer (#{limit})."
```
kotp: Then use `Kernel#Integer(limit)`

exercises/sieve/example.rb
```diff
@@ -1,6 +1,11 @@
 class Sieve
   attr_reader :range
   def initialize(limit)
+    begin
+      limit = limit.to_i
```
kotp: `limit = Kernel#Integer(limit)`

Or simply `limit = Integer(limit)`

kotp:
I realize that this changes the testing a little bit.  Also, you have a check for 0, but not a negative number.  You have the string with the number at the beginning, which is fine for the relaxed `.to_i` method, but not OK with the more defensive `Integer` method.

Thanks for tackling this, please do squash your commits when they are ready for review.

kytrinyx:
Once they're ready for review, please also add a comment to the PR... otherwise we don't get notified, and the only way I know to look again is if someone triggers a notification :heart:

esquinas:
I am sorry I am not experienced with Git, I won't know what to do next. 
I am just happy I was able to make the automated tests pass.
Really, I think the Sieve exercise is hard enough as it is without adding the _duck-typing lesson_. Which only makes sense in this case with the `.to_int` method, btw.  
I will close this pull request, but please consider adding a new exercise just for teaching duck-typing or, maybe, making it a requirement in the tests of another exercise.  
Sorry again and thanks for your comments.
Cheers, 
An exercism.io fan. :+1:

kotp:
I think you have a point with the duck-typing part of this.  I will definitely look at this exercise though, as reviewing your changes brought up some questions.

Even if this is to remain closed, you have contributed in ways you may not have expected to, which is not uncommon!

PR #312, opened 2016-05-04
Add alphametics

repinel:
Working on this one.


lib/alphametics_cases.rb
```diff
@@ -0,0 +1,20 @@
+class AlphameticsCase < OpenStruct
+  def test_name
+    "test_#{description.gsub(' ', '_')}"
+  end
+
+  def do
```
kotp: Any way to stop using 'do' as a method name?  I need to change Hamming, I think this is where this is coming from.

repinel: Sure thing

kotp: Thanks.  It is only a problem due to the language keyword list.

kytrinyx:
For what it's worth, @repinel, @kotp suggested "work_load" in a different patch, which I like. So unless you have a better idea (we can always change the other ones), that's a good choice.

exercises/alphametics/example.rb
```diff
@@ -0,0 +1,37 @@
...
+      return letters_values if valid?(puzzle, letters_values)
+    end
+
+    nil
+  end
+
+  private
+
+  def format_to_ruby(puzzle)
+    puzzle.gsub(/\^/, '**') # format exponent: 2 ^ 3 to 2 ** 3
```
repinel: Rubocop is suggesting `tr` instead of `gsub`, but I can't see how.

@kytrinyx @kotp any suggestion?

kytrinyx: No, this seems like a bug in rubocop.

kotp: No, but also I would not put spaces around the `**`.  It looks strange, there probably shouldn't have been spaces around the original exponent symbol.  It isn't how we do it on paper, it should be the same, if possible in the language, in code.

repinel: Sorry, the complain was actually for `lib/alphametics_cases.rb`. Fixed.

exercises/alphametics/alphametics_test.rb
```diff
@@ -0,0 +1,42 @@
...
+class AlphameticsTest < Minitest::Test
+  def test_solve_short_puzzle
+    expect = { 'I' => 1, 'B' => 9, 'L' => 0 }
+    actual = Alphametics.new.solve('I + BB == ILL')
+    assert_equal(expect, actual)
+  end
+
+  def test_solve_long_puzzle
+    skip
+    expect = { 'S' => 9, 'E' => 5, 'N' => 6, 'D' => 7, 'M' => 1, 'O' => 0, 'R' => 8, 'Y' => 2 }
```
repinel: According to rubocop this line is too long. The hash was gathered from the `expected` in the JSON, and I have added some better formatting in `AlphameticsCase#expected_value`, but I was wondering about this issue. Suggestions?

kytrinyx: I usually put these on multiple lines:

```
expect = {
  'S' => 9,
  'E' => 5,
  # ...
}
```

kotp: Or even...

```ruby
expect =
{
  something
}
```

kotp:
Pulled this down, currently `bin/local-status-check` gets hung up on alphametics check for tests.

repinel:
> Pulled this down, currently bin/local-status-check gets hung up on alphametics check for tests.

@kotp Yes. The brute-force solution with permutation is really slow. I am still thinking about improvement in the algorithm.

kotp:
OK, thanks.  Finally finished after 140 seconds.

repinel:
I got it working in half of the previous time. Now 20 seconds. I'll look into a different approach.

kotp:
You must have a more powerful machine than I have... that is better than 33% of the original time.

kotp:
@kytrinyx is 20 seconds OK for a test that will be ran fairly often?  Thinking about pre-push commit hook here.  It does not just check for what has changed, it checks the entire suite.

Cohen-Carlisle:
According to http://x.exercism.io/v3/tracks/ruby/todo, lua is the only implementing language. https://github.com/exercism/xlua/pull/75 is the PR. No idea how fast their tests run or if we could borrow anything from them to speed up our tests.

repinel:
@Cohen-Carlisle I had checked the Lua's build and it runs in 8s. However, it uses `coroutine` to multithread.

I have been experimenting with multithreading in the MRI, but since the bottleneck is CPU, not IO, it hasn't helped much yet. I am open to suggestions.

kytrinyx:
I would probably get pretty fed up with a 20 second test. I am less patient than most, and get fed up with test runs that take 3 seconds, even. Still, if it's a choice between 20 seconds and not testing it, I'll take 20 seconds.

kotp:
It is over 2 minutes on my machine... so to me it is very irritating.

kytrinyx:
Can we minimise the tests without compromising accuracy/completeness? Is there more data than necessary, or redundant data?

bernardoamc:
> Can we minimise the tests without compromising accuracy/completeness? Is there more data than necessary, or redundant data?

We could remove `test_solve_long_puzzle`, since it is probably the one that is slowing the suite. It's a good test, but if we don't have alternatives I think it is the best approach.

kotp:
Or make the long puzzle shorter, for time.  The example only needs to show that it can solve the problem, it does not need to be "The Solution".  And the test only needs to cover 'enough'.

kytrinyx:
Or comment it out and add a comment that if you want a thorough test, here's one but it's slow.

kotp:
Not just a skip, but a comment through the test... good idea.  It does mean that it should be enabled whenever this exercise is touched while developing/maintaining, and disabled again when pushed.

repinel:
I have comment out the long puzzle test. It should be enabled again whenever the tests are generated from x-common.

@kytrinyx @kotp what do you think?

exercises/alphametics/example.rb
```diff
@@ -0,0 +1,41 @@
+class Alphametics
+  VERSION = 1
+
+  def solve(puzzle)
+    letters = Hash.new(0)
+    puzzle = format_to_ruby(puzzle)
```
kotp: What does this name mean?  Is it language specific?

kotp:
I ran `bin/generate` for this, and the version numbers are not incrementing.

exercises/alphametics/example.rb
repinel: The tests have cases like `R ^ 2`, so this would it to the equivalent Ruby expression `R ** 2`. Would `math_operators_to_ruby` be more meaningful?

kotp: I think it is fine either way.  It doesn't indicate that it should be evaluated by Ruby itself.  For instance, there is no `instance_eval(format_to_ruby(puzzle))` kind of thing being done.

I think it has some value to format it to a 'normalized ruby expression' though, it gives a little extra something in the exercise.

The method name `as_ruby_expression` may be good?  I don't know, naming is hard.

repinel: `as_ruby_expression` seems good. I didn't add close the `eval` because it would be an extra overhead for the loop.

repinel:
The version number issue should now be fixed.

exercises/alphametics/example.rb
kotp: I don't think the eval is a good idea, so yeah< I think that is good.

I just noted that there wasn't anything stating that the formulas needed to be directly evaluate-able.

repinel:
@kotp I replaced the Ruby `eval` with an infix calculation. Thanks for the feedback!

bernardoamc:
I really like this new version. 🎉

exercises/alphametics/example.rb
```diff
@@ -0,0 +1,97 @@
...
+    operators = []
+
+    expression.scan(/\d+|\+|\-|\*|\/|\^/).each do |token|
+      case token
+      when /^\d+$/
+        numbers.push(token.to_i)
+      when '+', '-', '*', '/', '^'
+        calculate_last(numbers, operators) if has_precedence?(operators, token)
+        operators.push(token)
+      else # ignored
```
bernardoamc: We can remove this `else` clause here.

exercises/alphametics/example.rb
```diff
@@ -0,0 +1,97 @@
...
+  def has_precedence?(operators, token)
+    return false if operators.empty?
+    prev_operator = operators.last
+
+    case token
+    when '+', '-'
+      prev_operator == '*' || prev_operator == '/' || prev_operator == '^'
+    when '*', '/'
+      prev_operator == '^'
+    else
```
bernardoamc: This one can also be removed since `nil` will be treated as `false`, but I really like this code since it is explicitly returning `false` (as I would expect from a `?` method). I would keep it.

Cohen-Carlisle:
I'm really out of the loop on this one but it looks like its ready to merge. If so, let's do it!

kotp:
@Cohen-Carlisle any idea on how to solve the testing time on this?  Minimal test to provide reasonable proof it is working, hopefully with a test run that is not 30 seconds running time?

If commenting the entire test out is acceptable, then I agree, bring it in.

Cohen-Carlisle:
Hmm, it does seem like including a test that has that kind of run time is problematic. Does it really add anything? My gut says to remove it as it will most likely lead to confusion and frustration. If we decide not to include it, we should also open an issue on x-common to survey if removal is appropriate.

On another note, the PR should be updated to use the new bookkeeping method.

repinel:
@Cohen-Carlisle I've updated the bookkeeping method. Thanks

kotp:
Thanks @repinel for the BookKeeping update.

I am for bringing it in, with the commented test that is long, if someone wants to merge this.

Cohen-Carlisle:
I'd like to at least get a comment on top of the long test saying this will be long running, if we decided to keep it. Other than that, 👍

repinel:
I added a comment to the test. Please let me know if it is ok. Thanks

kotp:
Let's monitor the exercise on the site and see how it runs?  Merged, thanks @repinel for your work (and patience) in getting this in play.  Thanks @Cohen-Carlisle @bernardoamc and @kytrinyx for helping out with this.

Cohen-Carlisle:
:tada:

repinel:
I'm also curious to see on the site. Thank you all! ❤️

PR #314, opened 2016-05-04
Remove use of method name do (it is a keyword)

### LINKED ISSUE #313: Shouldn't use keyword 'do' as a method name.
Change whatever the 'Golden Master' is for the creation of generated tests so that `def do` stops showing up in the new exercises.

kotp:
Also address the removal of the messages that were in the test data, but
are no longer there.

Clean up the template for skips.

fixes #313


kytrinyx:
Yeah, agreed. I like the change to add the `# skip` as well.

PR #317, opened 2016-05-11
Binary Search Tree requires enumerator

remcopeereboom:
In the Ruby core classes `#each` returns an `Enumerator` to itself if it has not gottten a block to work on.

I added a test that requires this and updated the example to work with the new tests.

Added a version constant for the tests.

### DIFF
--- a/binary-search-tree/example.rb
+++ b/binary-search-tree/example.rb
@@ -1,5 +1,5 @@
 class Bst
-  VERSION = 2
+  VERSION = 1
   
   attr_reader :data, :left, :right
   def initialize(data)
--- a/binary-search-tree/example.rb
+++ b/binary-search-tree/example.rb
@@ -1,4 +1,6 @@
 class Bst
+  VERSION = 2
+  
   attr_reader :data, :left, :right
   def initialize(data)
     @data = data
--- a/binary-search-tree/example.rb
+++ b/binary-search-tree/example.rb
@@ -2,6 +2,7 @@ class Bst
   attr_reader :data, :left, :right
   def initialize(data)
     @data = data
+    @size = 1
   end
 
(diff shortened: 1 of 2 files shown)


binary-search-tree/example.rb
```diff
@@ -1,7 +1,10 @@
 class Bst
+  VERSION = 2
```
kotp: Version 1 would suffice, prior version was 0, as it was not tracked for versions.

remcopeereboom:
Decremented version number to use zero-based indexing :)

kotp:
This pull review is really out of date, it is 'pre-exercise folder' out of date.

remcopeereboom:
That seems to be an issue with exercism. I only forked minutes earlier. I remember this having occured before. Is there a way to rebase right on github?

kotp:
I just did a `git rebase origin/master` while on the pull request branch... it did have white space errors but otherwise was simple to rebase.

The white space errors will still need to be addressed, other than that, this should end up being a 'clean' operation.

remcopeereboom:
> I just did a git rebase origin/master while on the pull request branch... it did have white space errors but otherwise was simple to rebase.

Thanks, but that's not what I meant. I meant is there a way to do the rebase while on github.com, i.e. without needing to download the repository to my computer. I often only fork, change from within github, and then pull request. Typically the time is so short that there is no need to rebase. It would be so convenient if I could also do the rebasing online so I can work on computers that don't have `git` installed.

remcopeereboom:
I'll resubmit.

kytrinyx:
I don't think there is a way to rebase from GitHub.

remcopeereboom:
No I didn't think so, but thanks for mentioning it!

PR #318, opened 2016-05-12
Updated tests and example for Bst

remcopeereboom:
`Bst#each` should return an enumerator if it doesn't get a block.
Added a test to ensure that it does and that the returned enumerator
behaves properly.

Added a version constant for the tests.

Updated `example.rb` to pass the updated tests.


kytrinyx:
Did this get merged somewhere else? It looks like a really good change.

kytrinyx:
Ah, nevermind. I found the discussion here: https://github.com/exercism/xruby/pull/319

PR #319, opened 2016-05-12
Bst each should return enumerator

remcopeereboom:
Ruby's core collections all have `#each` return an enumerator when the method doesn't get a block. Bst should do the same, so I updated the tests and example.rb to reflect this.

### DIFF
--- a/exercises/binary-search-tree/binary_search_tree_test.rb
+++ b/exercises/binary-search-tree/binary_search_tree_test.rb
@@ -86,4 +86,31 @@ class BstTest < Minitest::Test
     four.insert 5
     assert_equal [1, 2, 3, 4, 5, 6, 7], record_all_data(four)
   end
+
+  def test_each_returns_enumerator_if_no_block
+    skip
+
+    tree = Bst.new 4
+    [2, 1, 3, 6, 7, 5].each { |x| tree.insert x }
+    each_enumerator = tree.each
+
+    assert_kind_of Enumerator, each_enumerator
+
+    (1..7).each { |x| assert_equal(x, each_enumerator.next) }
+
+    assert_raises(StopIteration) { each_enumerator.next }
+  end
+
+  # Problems in exercism evolve over time,
+  # as we find better ways to ask questions.
+  # The version number refers to the version of the problem you solved,
+  # not your solution.
+  #
+  # Define a constant named VERSION inside of Bst.
+  # If you are curious, read more about constants on RubyDoc:
+  # http://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/constants.html
+  def test_bookkeeping
+    skip
+    assert_equal 1, Bst::VERSION
+  end
 end
--- a/exercises/binary-search-tree/example.rb
+++ b/exercises/binary-search-tree/example.rb
@@ -1,7 +1,10 @@
 class Bst
+  VERSION = 1
+
   attr_reader :data, :left, :right
   def initialize(data)
     @data = data
+    @size = 1
   end
 
   def insert(value)
@@ -10,11 +13,15 @@ class Bst
     else
       insert_right(value)
     end
+
+    @size += 1
   end
 
   def each(&block)
+    return enum_for(:each) { @size } unless block_given?
+
     left && left.each(&block)
-    block.call(data)
+    yield
     right && right.each(&block)
   end
 
--- a/exercises/binary-search-tree/example.rb
+++ b/exercises/binary-search-tree/example.rb
@@ -21,7 +21,7 @@ class Bst
     return enum_for(:each) { @size } unless block_given?
 
     left && left.each(&block)
-    yield
+    yield data
     right && right.each(&block)
   end
 


kotp:
If you are enforcing this in the test, are you removing the discussion from the solution?

Up until your test, I don't see any tests that check for "type".

If it must be tested for, I am glad it is the last test.  But in having them solve for the return, the discussion of "why" may be lost on the website.

You also seem to be testing Ruby itself, in lines 99 and 101, on a built in class, that it will raise StopIteration, as spec'ed in the language itself.  If Ruby breaks, hopefully we find out about it, but not through this test.

I think the entire test could be expressed as:

```ruby
  def test_each_returns_enumerator_if_no_block
    skip
    assert_kind_of Enumerator, Bst.new(4).each
  end
```

remcopeereboom:
> If you are enforcing this in the test, are you removing the discussion from the solution?

Maybe, but I find that virtually nobody on exercism is aware of the idiom of returning an enumerator and as a result almost nobody is mentioning it in the discussions. My reasoning is that adding it to the tests is at least going to raise that awareness.

> Up until your test, I don't see any tests that check for "type".

Yeah I wasn't keen on the check for `Enumerator`. I took some time browsing through the mini test assertions for some alternatives, but I couldn't find anything decent. I considered that you might only want to check for the duck type and leave out the test for `assert_kind_of`, but I was worried that people might not understand the concept of enumerators.

> You also seem [...]

The problem of doing:

````ruby
def test_each_returns_enumerator_if_no_block
  skip
  assert_kind_of Enumerator, Bst.new(4).each
end
````

is that it's possble to return an arbitrary enumerator. The only way to check if it's the right enumerator is to see if returns the same values and raises the `StopIteration` exception at the right point. I've actually been bitten by this when I thought it was so trivial and not worth testing. Then you add up quickly typing the wrong method name in `enum_for(:method)` and boom!

**Comment by kotp** [comment]
Checking that it returns an enumerator and checking that it returns the right enumerator are two different tests though.  

But you aren't testing Bst's each in that test, you are testing Array's .each in that test.  This is why it was so trivial to toss that test out, it was testing the wrong object.

I think then, that it is your calling to give feedback on using and returning each when the review happens.  By doing this, you also give that same knowledge to others, and they start to do the same thing in their feedback.  

Rather than a test to enforce implementation details, I think we need to think about how to introduce the feedback.

With the test, perhaps they implement to pass the test, but still don't get that information that you want to impart, as to the why.

**Comment by remcopeereboom** [comment]
> But you aren't testing Bst's each in that test, you are testing Array's .each in that test. This is why it was so trivial to toss that test out, it was testing the wrong object.

I don't think I understand you. How so? The example doesn't even use an array.

> Rather than a test to enforce implementation details, I think we need to think about how to introduce the feedback.

I feel very strongly that returning an enumerator is not an implementation detail. It is the intended observable result of the method.

But if you feel equally strongly about this, I can close the pull.

**Comment by kotp** [comment]
`(1..7).each { |x| assert_equal(x, each_enumerator.next) }` is what I am talking about as to the testing the Array.

I think it is fine being open for a while, I need to think about how this will effect discussions in the code submissions, if having it there accomplishes the goal you have in mind, which is to make the 'best practice' known.

If it is accepted, but doesn't accomplish your goal, then it is counter productive to take it in.  That is the only concern I have for it.

However, with that said, it is useful to have the example.rb show this in the code, even perhaps with a comment.  It isn't something that normal exercist reviewers see, but it is there for us to see, as a reminder when we go to touch the exercise again.

**Comment by kytrinyx** [comment]
OK, so here's a completely different direction... what if Rikki automatically commented about it if the bst each wasn't following the idiom?

**Comment by kotp** [comment]
It is good, because it leaves it open to make the solution as they would, and remembers to always mention it so we don't have to.

And I like that. :)

**Comment by remcopeereboom** [comment]
I like putting it in the rikki since it brings up a discussion point, but I still feel that having an enumerator is the method's observable behaviour. It's part of the expected functionality. _Not_ returning an enumerator is so non-standard that I would expect a client to specify it as required behaviour.

If you do decide to put it in the rikki than I suggest making it specific to `#each`, not just `Bst#each` (although I'm not aware of any other exercism requiring an `#each` method).

**Comment by kytrinyx** [comment]
> If you do decide to put it in the rikki than I suggest making it specific to #each, not just Bst#each

Agreed. Even if other exercises don't require an each, if someone does implement one it would be the perfect time to bring up the idiom.

File: exercises/binary-search-tree/binary_search_tree_test.rb [hunk 1]
```diff
@@ -86,4 +86,31 @@ def test_iterating_over_complex_tree
...
+
+    tree = Bst.new 4
+    [2, 1, 3, 6, 7, 5].each { |x| tree.insert x }
+    each_enumerator = tree.each
+
+    assert_kind_of Enumerator, each_enumerator
+
+    (1..7).each { |x| assert_equal(x, each_enumerator.next) }
+
+    assert_raises(StopIteration) { each_enumerator.next }
```
Cohen-Carlisle: What if instead of lines 99 and 101 we simply do:
```ruby
assert_equal %w(1 2 3 4 5 6 7), each_enumerator.map(&:to_s)
```

remcopeereboom: @Cohen-Carlisle 
Interesting thought, but it's easy to slip up and have `#map` be called on the wrong object as `#each` typically returns the object itself when it does get a block. I haven't added a test for returning `self` as there are occasions where it requires choosing between exposing internals or adding extra level of indirection in the iteration. If you do return `self`, than `#map` doesn't actually test that its an enumerator.

Cohen-Carlisle:
What's the status of this PR? If @remcopeereboom still wants to merge it let's get it back on track!

kotp:
@kytrinyx are we waiting on Riki issue exercism/rikki#16 or should we merge it now?

kytrinyx:
No, rikki-'s going to take its sweet old time, I'm afraid. Let's merge this.

kotp:
Thanks @remcopeereboom for the patience and hard work on this.

PR #323, opened 2016-05-13
add perfect-numbers

chintas1:
Will be working on translating to ruby

### DIFF
--- a/exercises/perfect-numbers/example.rb
+++ b/exercises/perfect-numbers/example.rb
@@ -1,0 +1,25 @@
+class PerfectNumber
+  VERSION = 1
+  def self.classify(num)
+    raise 'not a natural number' if num < 0
+    aliquot_sum = aliquot_sum(num)
+    aliquot_sum == num ? "perfect" : aliquot_sum < num ? "deficient" : 'abundant'
+  end
+
+  def self.aliquot_sum(num)
+    get_divisors(num).reduce(:+)
+  end
+
+  private
+
+  def self.get_divisors(num)
+    divisors = [0]
+    (1..Math.sqrt(num)).each do |n|
+      if num % n == 0
+        divisors << n
+        divisors << num/n unless (n == 1) || (num/n) == n
+      end
+    end
+    divisors
+  end
+end
--- a/exercises/perfect-numbers/perfect_numbers_test.rb
+++ b/exercises/perfect-numbers/perfect_numbers_test.rb
@@ -1,0 +1,27 @@
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'example'
+
+class PerfectNumberTest < Minitest::Test
+  def test_initialize_perfect_number
+    assert_raises RuntimeError do
+      PerfectNumber.classify(-1)
+    end
+  end
+
+  def test_classify_deficient
+    assert_equal 'deficient', PerfectNumber.classify(13)
+  end
+
+  def test_classify_perfect
+    assert_equal 'perfect', PerfectNumber.classify(28)
+  end
+
+  def test_classify_abundant
+    assert_equal 'abundant', PerfectNumber.classify(12)
+  end
+
+  def test_bookkeeping
+    assert_equal 1, PerfectNumber::VERSION
+  end
+end
--- a/config.json
+++ b/config.json
@@ -74,7 +74,8 @@
     "custom-set",
     "minesweeper",
     "scale-generator",
-    "protein-translation"
+    "protein-translation",
+    "perfect-numbers"
   ],
   "deprecated": [
     "octal",


chintas1:
ready for review

exercises/perfect-numbers/example.rb
```diff
@@ -0,0 +1,39 @@
...
+    factors = get_divisors
+    factors.reduce(:+)
+  end
+
+  
+  private
+
+  def get_divisors
+    return [0] if num <= 1
+    divisors = []
```
bernardoamc: Can we remove the line above and just start divisors with `[0]`?

exercises/perfect-numbers/example.rb
```diff
@@ -0,0 +1,39 @@
...
+    @num = num
+  end
+
+
+  def classify
+    aliquot_sum == num ? "perfect" : aliquot_sum < num ? "deficient" : 'abundant'
+  end
+
+
+  def aliquot_sum
```
bernardoamc: We could do just `divisors.reduce(:+)`, what do you think?

exercises/perfect-numbers/perfect_numbers_test.rb
```diff
@@ -0,0 +1,44 @@
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'example'
+
+class PerfectNumberTest < Minitest::Test
+  def test_aliquot_sum_zero
+    assert_equal 0, PerfectNumber.new(1).aliquot_sum
```
bernardoamc: I don't know if we should force people to implement a public method called `aliquot_sum`. The exercise specifies that we need to classify numbers, how to do it should be open to the user. Should we test `aliquot_sum` or just a few cases of `classify`?

exercises/perfect-numbers/example.rb
```diff
@@ -0,0 +1,39 @@
+class PerfectNumber
+  VERSION = 1
+  attr_accessor :num
+  
+  def initialize(num)
+    raise "#{num} is not a natural number." if num < 0
```
bernardoamc: Should we add a test for this? :)

chintas1: Hey Bernardo. Was having an initialize and an instance variable a good idea on my part? I think if I wanted to write tests for this, I would have enforce the users to also call their instance variable num which doesn't seem correct.

exercises/perfect-numbers/perfect_numbers_test.rb
chintas1: I looked at the other implementations for reference and was a bit confused on this point. The lua variant has it similar to this where it tests both aliquot_sum and classify. Would you suggest just testing cases for classify and not aliquot_sum?

bernardoamc: Yes, but it's just an idea, I don't think a method called `aliquot_sum` (in the public API) is necessary to solve this problem. :)

exercises/perfect-numbers/example.rb
bernardoamc: Usually I write my exception messages without relying on variables (the exception being if it is a custom Exception). In this case we could use a generic message like  `a natural number is expected.` or something similar. What do you think?

exercises/perfect-numbers/perfect_numbers_test.rb
chintas1: Thanks Bernardo that makes sense since the core functionality should be the classify. This is my first time contributing to open source so I tried to follow the lua example very strictly. Thanks for the feedback I'll start making the changes now :).

exercises/perfect-numbers/example.rb
chintas1: Oh yeah I totally agree! I was confused at first because I thought you meant I should write a test to make sure num was getting set in initialize

bernardoamc: I see! Sorry for not being clear enough! I was suggesting that we could write a test to see if an exception will be raised in case we initialize this class with something other than a natural number.

bernardoamc:
Thank you for this PR @chintas1 ! ❤️

exercises/perfect-numbers/perfect_numbers_test.rb
bernardoamc: No problem, you did an awesome job! 👍

exercises/perfect-numbers/perfect_numbers_test.rb
```diff
@@ -3,23 +3,10 @@
...
-    assert_equal 15, PerfectNumber.new(16).aliquot_sum
-  end
-
-  def test_aliquot_sum_large_num
-    skip
-    assert_equal 229, PerfectNumber.new(1115).aliquot_sum
+  def test_initialize_perfect_number
+    assert_raises RuntimeError do
+      PerfectNumber.new(-1)
+    end
```
kotp: `assert_raises RuntimeError { PefectNumber.new(-1) }`

kotp:
@chintas1 The tests strongly suggests that we define a class, due to the method `new` being called on the constant.

Can we re-write the test to try to avoid dictating this, especially as, in the tests, there is no need to have more than one PerfectNumber object, and there seems to be no call for state.  In fact, this problem can be solved in a purely functional way.

exercises/perfect-numbers/perfect_numbers_test.rb
chintas1: I'm having trouble getting the assert_raises to work when I try to implement it in this format. NoMethodError: undefined method `RuntimeError' is the error I get.

kotp: It is due to the stronger adhesion of the block.  Use parenthesis around the argument `RuntimeError` to compensate for this.

kotp:
This should end up being fine, as it is.  Take note of the in-line comment regarding the stronger binding nature of the `{ }` format of a block, compared to `do..end` but don't worry too much about it.  It was just a nicety more than anything.

PR #325, opened 2016-05-13
[wip] add change

eessex:
working on this

### DIFF
--- a/exercises/.byebug_history
+++ b/exercises/.byebug_history
@@ -1,0 +1,18 @@
+exit
+target
+exit
+target
+coins
+exit
+coins.inject([]) {|mem, var| mem.any? {|c| c%var == 0} ? mem : mem+[var]}
+coins.reject { |coin| coin > target }
+coins
+exit
+mem
+exit
+key
+hash
+n
+key
+hash
+optimal_change
--- a/config.json
+++ b/config.json
@@ -74,7 +74,8 @@
     "custom-set",
     "minesweeper",
     "scale-generator",
-    "protein-translation"
+    "protein-translation",
+    "change"
   ],
   "deprecated": [
     "octal",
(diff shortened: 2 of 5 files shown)


exercises/.byebug_history
```diff
@@ -0,0 +1,18 @@
+exit
```
kytrinyx: What does this file do?

eessex: That can be deleted, it was automatically created by a debugging gem. Thanks!

kotp:
@essex the ignore should be done in your `.gitignore` file, in your home folder.

Things like editor, operating system files, and various tools, that are there because of individual developers, should be addressed at that level, rather than the project level.

There are a few `.gitignore` entries that are not quite exemplory in exercism that I would love to see cleaned up.

kytrinyx:
> There are a few .gitignore entries that are not quite exemplory in exercism that I would love to see cleaned up.

Yeah, agreed. We should clean those up and make it part of our process here to help people get their development machine configured correctly.

w1zeman1p:
Re `.gitignore` it might be good to start with these ones? https://github.com/github/gitignore They have more languages than are supported by exercism, but i'm not sure they overlap directly.

kytrinyx:
@w1zeman1p Interesting. Yeah, we could at the very least link to that directory in the "launch checklist" when we create a new repository, and ask people to make sure that the gitignore makes sense for the language.

kytrinyx:
@eessex did you get stuck on this PR? Is there anything we can help out with?

Insti:
Thanks for deciding to help us add exercises to Exercism.
If you are, is there anything you need help with?
If you're not, use the 'close request' button on Github so someone else can work on adding the problem.

If we don't hear back from you, we will close it ourselves one week from now. 
(On or after the 30th July 2016)

PR #332, opened 2016-05-13
add list-ops

rhoprhh:


rhoprhh:
ready for review

rhoprhh:
@invasivelionfish collaborated with me on building these tests

exercises/list-ops/example.rb
```diff
@@ -0,0 +1,67 @@
+class ListOps
```
bernardoamc: Hi @rhoprhh! Since you are adding these methods in your own class it is ok to have methods named `map` and `concat` inside it. Any motive to use your own names?

exercises/list-ops/example.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def self.mapper(arr)
+    answer = []
+    arr.each do |element|
+      answer << (element + 1)
+    end
+    answer
+  end
+
+  def self.filterer(arr)
```
bernardoamc: I think methods like this could receive a block without relying on a fixed condition like `element % 2 == 1`. It would be something like `self.filter(arr, &block)`. With this you can call your block inside the `each` method with `block.call(element)`.

exercises/list-ops/example.rb
rhoprhh: The markdown document said to not use existing functions. I just wanted to make sure it was very obvious that you were writing your own methods and not just using array enumerators to do it.

Here is the entire .md document:
"In functional languages list operations like length, map, and reduce are very common. Implement a series of basic list operations, without using existing functions."

exercises/list-ops/example.rb
rhoprhh: That is definitely a possible way to do this, but this seemed very much like a beginner lab and we were trying to use very simple and explicit answers.

bernardoamc: Yeah, this is a problem indeed, but the way `filterer` was implemented is not exactly like `filter`, right? I think we could implement simpler methods (the ones that don't expect a block) if this is a problem for beginners. So `length`, `reverse`, `concat`, `pop` are all good methods.

exercises/list-ops/example.rb
bernardoamc: Since this file is an example, I don't think using the "real" names is a problem. But no strong feelings about this, just my opinion. :)

exercises/list-ops/example.rb
rhoprhh: The markdown said not to use those methods. Our original example had used all 4 of those, but we felt we needed to re-write all of them to fit what was wanted better.
If I was supposed to use those methods, then it was because the markdown file wasn't very clear since it was only 1 sentence.

kotp:
I will review this in a few hours, looks like it is at that stage.  Let me know if it should not yet have the review tag on it.

exercises/list-ops/example.rb
kytrinyx: I lean slightly towards the direction of preferring the real names, because people already know what they do. It's obvious when looking at the code that these aren't just using the standard library implementations, so I think that would be totally fine.

exercises/list-ops/list_ops_test.rb
```diff
@@ -0,0 +1,107 @@
...
+require_relative 'list_ops'
+
+class ListOpsTest < Minitest::Test
+  def test_count_empty
+    assert_equal 0, ListOps.arrays([])
+  end
+
+  def test_count_normal
+    skip
+    assert_equal 5, ListOps.arrays(Array.new(5))
```
bernardoamc: This method name is a little misleading since I'm expecting multiple arrays instead of elements.
So maybe we could call it `elements` instead, or even `length`?

exercises/list-ops/list_ops_test.rb
```diff
@@ -0,0 +1,107 @@
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'list_ops'
+
+class ListOpsTest < Minitest::Test
+  def test_count_empty
+    assert_equal 0, ListOps.arrays([])
+  end
+
+  def test_count_normal
```
bernardoamc: I think `test_count_normal` is not expressive enough in this situation, what is `normal` in this context?  Should we call it `test_count_elements`?

exercises/list-ops/list_ops_test.rb
```diff
@@ -0,0 +1,107 @@
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'list_ops'
+
+class ListOpsTest < Minitest::Test
+  def test_count_empty
```
bernardoamc: So, people trying to solve this exercise will not be able to see the `example.rb` file, so the tests should be the guide here. So maybe we could call this `test_count_empty_array_elements`?  This way people know exactly what is expected just by reading the title.

exercises/list-ops/list_ops_test.rb
```diff
@@ -0,0 +1,107 @@
...
+    assert_equal 0, ListOps.arrays([])
+  end
+
+  def test_count_normal
+    skip
+    assert_equal 5, ListOps.arrays(Array.new(5))
+  end
+
+  def test_count_gigantic
```
bernardoamc: I think this test is not adding/testing a new functionality, maybe it can be removed? Was this a problem during implementation?

exercises/list-ops/list_ops_test.rb
```diff
@@ -0,0 +1,107 @@
...
+    skip
+    assert_equal 1_000_000, ListOps.arrays(Array.new(1_000_000))
+  end
+
+  def test_reverse_empty
+    skip
+    assert_equal [], ListOps.reverser([])
+  end
+
+  def test_reverse_normal
```
bernardoamc: I think we have the same situation here regarding the word `normal`, what is it in this context? There are a few tests like this, maybe we could change them so the title becomes more expressive?

kotp:
The overall test time on my (admittedly slow) computer is 5.77 seconds.  On a one time run.  Is this normal, or is this because of the 'gigantic' tests?

exercises/list-ops/list_ops_test.rb
```diff
@@ -0,0 +1,107 @@
...
+    skip
+    assert_equal [5, 4, 3, 2, 1], ListOps.reverser([1, 2, 3, 4, 5])
+  end
+
+  def test_reverse_gigantic
```
bernardoamc: So are we testing for performance with these `gigantic` tests? Maybe we could remove them if they are not adding new functionality.

kytrinyx: Agreed. In general here's my philosophy with respect to test data:

- as small as possible
- as boring as possible

If I'm testing "more than one", then I use 2, unless there's something about a longer list that will potentially have different behavior that could be missed with two (a middle element, some sort of overlap, etc).

I want the test data to express as succinctly as possible what we are actually testing, so I avoid "real-ish" data, because then the person is left guessing why that data was used. Incidental features in the "real-ish" data can act as red herrings, and people make up mental models of what the problem is about that might not be accurate, and that are definitely not necessary.

kotp: @rhoprhh Just checking in to see if there are any changes coming in from your side...  Thanks for all the work so far!  Looking forward to getting these in!

Insti:
Has there been any progress on this? 
Does anyone want to address the issues raised by @bernardoamc's last review?

If not I suggest we should just merge it and have a less-than-perfect problem out there for people to try and then suggest improvements to.

kytrinyx:
> If not I suggest we should just merge it and have a less-than-perfect problem out there for people to try and then suggest improvements to.

Yeah, let's do it.

PR #342, opened 2016-05-13
[WIP]  add sublist (Closes #335)

### LINKED ISSUE #335: Sublist
Liana and Steve working on sublist

lianaluci:


exercises/sublist/example.rb
```diff
@@ -0,0 +1,18 @@
+
+def sublist(set_a, set_b)
```
devonestes: There are actually great methods around doing this in the Ruby standard library - maybe you could find some inspiration there for a more easily readable example implementation? http://ruby-doc.org/stdlib-2.3.1/libdoc/set/rdoc/Set.html

bernardoamc: I think using the `Set` here doesn't solve the problem entirely because of the following example in the [sublist.md](https://github.com/exercism/x-common/blob/master/sublist.md) file.

`A = [1, 2, 4], B = [1, 2, 3, 4, 5], A is not a superlist of, sublist of or equal to B`

If we use the `Set`:

```ruby
a = Set.new([1, 2, 4])
b = Set.new([1, 2, 3, 4, 5])
a.subset?(b)
true # But as the problem specified, it should not be considered a subset.
```

exercises/sublist/example.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  return 'Equal' if set_a == set_b
+  return 'A is a subset' if subset(set_a, set_b)
+  return 'A is a superset' if subset(set_b, set_a)
+
+  'There are no subsets'
+end
+
+def subset(set_1, set_2)
```
bernardoamc: Just for the sake  of curiosity, I think an alternative implementation could be:

```ruby
def subset(set_1, set_2)
  return false if set_2.size < set_1.size

  size_diff = set_2.size - set_1.size

  (0..size_diff).each do |start|
    return true if set_2.slice(start, set_1.size) == set_1
  end

  false
end
```

bernardoamc:
Thanks @lianaluci 🎉

exercises/sublist/sublist_test.rb
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env ruby
+
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'sublist'
+
+class Sublist < Minitest::Test
```
bernardoamc: There is a JSON file with common tests that other languages have implemented for this problem, you can look it [here](https://github.com/exercism/x-common/blob/master/sublist.json) and create similar tests and expectations. :)

exercises/sublist/example.rb
kytrinyx: Agreed, sets are unordered and the elements are unique, whereas lists are ordered and can have multiples of elements.

exercises/sublist/sublist_test.rb
kytrinyx: I would name the test class `SublistTest` rather than Sublist, for a couple of reasons:

1. It follows the convention that we use elsewhere.
2. It won't risk conflicting with the implementation.

exercises/sublist/example.rb
devonestes: Ahhh, I didn't fully understand the problem then. Thanks for the clarification.

Insti:
Thanks for deciding to help us add exercises to Exercism.
If you are, is there anything you need help with?
If you're not, use the 'close request' button on Github so someone else can work on adding the problem.

If we don't hear back from you, we will close it ourselves one week from now. 
(on or after 31 July 2016)

Insti:
Thanks for the effort you've put into this.
Since you seem to have stopped working on it before it was completed,
we're closing this request without merging in the changes.

If you decide you want to continue and complete it, you can re-open this request (or start a new one) and we'll be happy to assist you.

PR #345, opened 2016-05-14
Robot-Name: Better error message on test failure.

Insti:
Use refute_equal (rather than assert !=) in order to return a more helpful error message on test failure.


kotp:
Thanks @Insti

PR #346, opened 2016-05-15
add rail-fence-cipher

raysapida:
I'm working on tests and a reference solution for the rail-fence-cipher exercise.

### DIFF
--- a/exercises/rail-fence-cipher/example.rb
+++ b/exercises/rail-fence-cipher/example.rb
@@ -1,0 +1,73 @@
+# rubocop:enable all
+class RailFenceCipher
+  VERSION = 1
+
+  def self.encode(message, rails)
+    return message if message.empty? || rails == 1
+    empty_fences = create_empty_fences_array(message, rails)
+    zig_zag_array = populate_fences_array(message, rails, empty_fences)
+    zig_zag_array.flatten.join
+  end
+
+  def self.decode(message, rails)
+    return message if message.empty? || rails == 1
+    empty = create_empty_fences_array(message, rails)
+    mark_fences = populate_fences_array('X' * message.length, rails, empty)
+    line_lengths = find_the_lengths_of_the_lines(message, rails)
+    lines = split_the_encrypted_message_by_rows(message, line_lengths)
+    result = switch_marked_cells_with_string(mark_fences, lines, empty)
+    result.transpose.flatten.join
+  end
+
+  def self.create_empty_fences_array(message, rails)
+    Array.new(rails) { Array.new(message.length) }
+  end
+
+  def self.populate_fences_array(message, rails, fences)
+    row = 0
+    row_direction = 1
+    message.length.times do |column|
+      fences[row][column] = message[column]
+      row_direction = 1 if row == 0
+      row_direction = -1 if row == (rails - 1)
+      row += row_direction
+    end
+    fences
+  end
+
+  def self.find_the_lengths_of_the_lines(message, rails)
+    line_lengths = Hash.new(0)
+    row = 0
+    row_direction = 1
+    (0..(message.length - 1)).each do
+      line_lengths[row] += 1
+      row_direction = 1 if row == 0
+      row_direction = -1 if row == (rails - 1)
+      row += row_direction
+    end
+    line_lengths
+  end
+
+  def self.split_the_encrypted_message_by_rows(message, line_lengths)
+    lines = []
+    line_lengths.each do |_row_num, line_length|
+      lines << message.slice!(0...line_length)
+    end
+    lines
+  end
+
+  def self.switch_marked_cells_with_string(mark_fences, lines, fences)
+    mark_fences.map.with_index do |row, row_number|
+      row.map.with_index do |cell, col_number|
+        if cell == 'X'
+          fences[row_number][col_number] = lines[row_number].slice!(0)
+        end
+      end
+    end
+  end
+
+  private_class_method :create_empty_fences_array,
+                       :populate_fences_array,
+                       :find_the_lengths_of_the_lines,
+                       :split_the_encrypted_message_by_rows
+end
--- a/exercises/rail-fence-cipher/rail_fence_cipher_test.rb
+++ b/exercises/rail-fence-cipher/rail_fence_cipher_test.rb
@@ -1,0 +1,67 @@
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'rail_fence_cipher'
+
+# rubocop:enable all
+class RailFenceCipherTest < Minitest::Test
+  def test_encode_with_empty_string
+    assert_equal '', RailFenceCipher.encode('', 4)
+  end
+
+  def test_encode_with_one_rail
+    skip
+    assert_equal 'One rail, only one rail',
+                 RailFenceCipher.encode('One rail, only one rail', 1)
+  end
+
+  def test_encode_with_two_rails
+    skip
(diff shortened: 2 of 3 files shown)


raysapida:
I think the reference solution and tests are ready for review. 

For the reference solution, I wasn't sure if it was ok to use the syntax below to make the class methods besides `encode` and `decode` private. The rubocop gem was recommending another method called `private_class_method` and I could also change it to that instead. 

```ruby
class << self
    private
    ..
end
```

exercises/rail-fence-cipher/example.rb
```diff
@@ -0,0 +1,72 @@
...
+    private
+
+    def create_empty_fences_array(message, rails)
+      Array.new(rails) { Array.new(message.length) { nil } }
+    end
+
+    def populate_fences_array(message, rails, fences)
+      row = 0
+      row_direction = 1
+      (0..(message.length - 1)).each do |column|
```
bernardoamc: We could use `...` here, meaning that you want to stop right before the `message.length`.

Another possibility is to use the `times` method, in this case it would be  `message.length.times do |column|`

exercises/rail-fence-cipher/example.rb
```diff
@@ -0,0 +1,72 @@
...
+    lines = split_the_encrypted_message_by_rows(message, line_lengths)
+    result = switch_marked_cells_with_string(mark_fences, lines, empty)
+    result.transpose.flatten.join
+  end
+
+  class << self
+    private
+
+    def create_empty_fences_array(message, rails)
+      Array.new(rails) { Array.new(message.length) { nil } }
```
bernardoamc: The default of `Array.new(number)` is generating an array with nils, so we could also do: 
`Array.new(rails) { Array.new(message.length) }

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    assert_equal 'WECRLTEERDSOEEFEAOCAIVDEN',
+                 RailFenceCipher.encode('WEAREDISCOVEREDFLEEATONCE', 3)
+  end
+
+  def test_encode_with_ending_in_the_middle
+    skip
+    assert_equal 'ESXIEECSR', RailFenceCipher.encode('EXERCISES', 4)
+  end
+
+  def test_encode_with_a_more_diverse_text
```
bernardoamc: I think this test is not really exercising a new feature or edge case, could this be removed?

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
```diff
@@ -0,0 +1,94 @@
...
+                 RailFenceCipher.decode('XXXXXXXXXOOOOOOOOO', 2)
+  end
+
+  def test_decode_with_three_rails
+    skip
+    assert_equal 'THEDEVILISINTHEDETAILS',
+                 RailFenceCipher.decode('TEITELHDVLSNHDTISEIIEA', 3)
+  end
+
+  def test_decode_with_a_more_diverse_text
```
bernardoamc: I don't know exactly if adding a new rail exercises a new aspect of the code, if it doesn't we could trim a few test cases. What do you think?

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
```diff
@@ -0,0 +1,94 @@
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'rail_fence_cipher'
+
+# rubocop:enable all
+class RailFenceCipherTest < Minitest::Test
+  def test_encode_with_empty_string
+    assert_equal '', RailFenceCipher.encode('', 4)
```
bernardoamc: I like the use of a class method here, this gives people the opportunity to use a `module` to solve the exercise if they want to. 👍

bernardoamc:
Nice @raysapida ! 🎉 

I think it would be good to use `private_class_method`, but this is just my preference! :)

exercises/rail-fence-cipher/example.rb
raysapida: For me, it looks like `message.length.times` would be the most readable. Thank you for suggesting it!

exercises/rail-fence-cipher/example.rb
raysapida: Ohh , I see. Would it be better to be explicit that it's populating with nil or is it a common assumption that the default is nil?

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
raysapida: I think it could be removed. I was adding most of the tests cases from different implementations and they do have a lot of overlap. In those situations, would it be better to have just enough tests to cover the feature and edge cases?

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
raysapida: While I was going through the challenge, it seemed like adding a new rail helped me make sure that I wasn't depending on the wrong pattern.  I first tried to populate the fences without marking them and I found out that the logic I was using wouldn't work past the third rail.

exercises/rail-fence-cipher/example.rb
bernardoamc: That's a good question, I think we could say it is a common assumption. :)

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
bernardoamc: I think so, when I see a test I always wonder what feature/edge case it is covering. If it is not doing this I usually remove it since it is more code for people to read that is not helping them to understand the logic.

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
bernardoamc: Yeah, tests are awesome during development, but you have found that the same logic applies to all cases now, right? Maybe this could even keep people wondering if they have to implement something different for texts with 4 or more fences, for example.

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
kotp: A name such as "a more diverse..." that can not be clear in what it is testing, is likely a hint it is not necessary, as it can not be articulated what specifically it is testing for.

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
raysapida: Yeah, it does all apply to all of those cases now. And I could see what you meant earlier that having a lot of tests wouldn't necessarily help make the code easier to understand if they're covering the same things.

exercises/rail-fence-cipher/rail_fence_cipher_test.rb
kotp: Interestingly, this doesn't only give people the opportunity to use a module, but it does free up the solution to be pretty open.

kotp:
Looks good to me, I think this is ready for merge.

kytrinyx:
I agree. @raysapida would you rebase this onto master to resolve the conflicts? (We can also do that for you, just let us know).

raysapida:
Hi @kytrinyx and @kotp . Would rebasing it onto master be similar to this?

kotp:
@raysapida That would be the way to do it. :cake:

PR #351, opened 2016-05-20
Config for Gitter Room Chat on page

kotp:
This is tied to exercism/exercism.io#2893


kytrinyx:
Yeah, I think that's a good implementation.

PR #353, opened 2016-05-24
[nth-prime] Remove constraint on Prime class usage.

### LINKED ISSUE #352: nth-prime test suite throwing error ThreadError:deadlock; recursive locking
The following line in the Test suite causes the error down below. 
In particular it is the re-writing of the `:new` method that is causing the problem:

```{ruby}
class Prime
  [:each, :new, :prime?, :take].each do |m|
    define_method(m) { |*_| fail ERROR_MESSAGE }
  end
end
```

```
  1) Error:
TestPrimes#test_first:
ThreadError: deadlock; recursive locking
    /Users/davidacevedo/.rbenv/versions/2.3.0/lib/ruby/2.3.0/singleton.rb:140:in `synchronize'
    /Users/davidacevedo/.rbenv/versions/2.3.0/lib/ruby/2.3.0/singleton.rb:140:in `instance'
    /Users/davidacevedo/.rbenv/versions/2.3.0/lib/ruby/2.3.0/singleton.rb:142:in `block in instance'
    /Users/davidacevedo/.rbenv/versions/2.3.0/lib/ruby/2.3.0/singleton.rb:140:in `synchronize'
    /Users/davidacevedo/.rbenv/versions/2.3.0/lib/ruby/2.3.0/singleton.rb:140:in `instance'
    nth_prime_test.rb:30:in `test_first'
```

bernardoamc:
Fix #352 

This constraint was throwing an error in new Ruby versions, quoting @Cohen-Carlisle :

> I favor removal of the anti cheat as impacting a non cheating user is worse than not catching a cheating user.

### DIFF
--- a/exercises/nth-prime/nth_prime_test.rb
+++ b/exercises/nth-prime/nth_prime_test.rb
@@ -1,27 +1,7 @@
 #!/usr/bin/env ruby
-# frozen_string_literal: true
 
 gem 'minitest', '>= 5.0.0'
 require 'minitest/autorun'
-
-require 'prime'
-ERROR_MESSAGE = <<-MSG.freeze
-Using Ruby's Prime class is probably the best way to do this in a
-'real' application; but this is an exercise, not a real application,
-so you're expected to implement this yourself.
-MSG
-
-class Prime
-  [:each, :new, :prime?, :take].each do |m|
-    define_method(m) { |*_| fail ERROR_MESSAGE }
-  end
-end
-
-class Integer
-  [:prime?, :each_prime].each do |m|
-    define_method(m) { |*_| fail ERROR_MESSAGE }
-  end
-end
 
 require_relative 'nth_prime'
 


bernardoamc:
Also, I don't see a problem in people using the Prime class actually. Maybe they are learning the language and found out that Ruby provides it, which is a nice thing.

bernardoamc:
I was talking with @kotp about this and explained that the bug disappears by removing the `:new` symbol from [here](https://github.com/exercism/xruby/pull/353/files#diff-e1866a7bd9c2ecd0055814df3d2b47aeL15). Should we consider this as an option instead of removing everything?

Cohen-Carlisle:
I'm a bit biased as I don't like the anti cheat to begin with... Personally, I'd go with remove them all.

bernardoamc:
Another option would be changing the expected class/module to be something like `NthPrime` and dropping a line like this in the test file: `Prime = 'This class is off limits.'`

kytrinyx:
I think I'm slowly deciding that the anti-cheat is less than ideal. We don't want to artificially constrain people, and we do want to provide rich opportunities for conversation. I'm still on the fence about it all, but I'm +1 "looks good to me" on this change.

kotp:
Bringing this in as is, as eyes have been on it, and it looks good.

PR #354, opened 2016-05-24
BookKeeping moved to its own responsibility

kotp:
This will avoid unnecessary errors when solutions aren't the more common
directions.


exercises/hamming/example.tt
```diff
@@ -17,11 +17,11 @@ class HammingTest < Minitest::Test<% test_cases.each do |test_case| %>
   # The version number refers to the version of the problem you solved,
   # not your solution.
   #
-  # Define a constant named VERSION inside of Hamming.
+  # Define a constant named VERSION inside of BookKeeping.
```
kytrinyx: Nothing has mentioned something called `BookKeeping` until this moment, and it might feel like it came out of nowhere (_book-keeping? What book-keeping? Am I supposed to know what that means?_).

Since this is the first (non-hello-world) exercise, I suggest that we be a little bit more verbose here. Maybe something like:

```
# Create a class or module named BookKeeping which defines a constant named VERSION.
# You can do this in the same file as the rest of your solution.
```

kytrinyx:
I like where this is going. I have one comment about how we introduce it the first time it's mentioned.

lib/bookkeeping.md
```diff
@@ -0,0 +1,13 @@
...
+  # as we find better ways to ask questions.
+  # The version number refers to the version of the problem you solved,
+  # not your solution.
+  #
+  # Define a constant named VERSION inside of BookKeeping module.
+  #  In your file, it will look like this:
+  # module BookKeeping
+  #   VERSION = 1 # Where the version number matches the one in the test.
+  # end
+  #
```
kotp: Addresses the comment by @kytrinyx but not in place until we generate.

lib/bookkeeping.md
```diff
@@ -0,0 +1,13 @@
+  # Problems in exercism evolve over time,
+  # as we find better ways to ask questions.
+  # The version number refers to the version of the problem you solved,
+  # not your solution.
+  #
+  # Define a constant named VERSION inside of BookKeeping module.
```
kytrinyx: inside of _a_ BookKeeping module?

lib/bookkeeping.md
```diff
@@ -0,0 +1,13 @@
...
+  #  In your file, it will look like this:
```
kytrinyx: The indentation is odd here.

lib/bookkeeping.md
```diff
@@ -0,0 +1,13 @@
...
+  # module BookKeeping
```
kytrinyx: Let's have a blank line before showing the code example.

lib/bookkeeping.md
kotp: Yes, yes, it is.  It matches where it is being read into.

kytrinyx: I don't understand, would you elaborate please?

kotp:
Steps after this is merged:

1. Proof the `bookkeeping.md` file for any changes, make appropriate PR or quick change, as appropriate.
2. Run the generators in `bin/generate-*`
3. Submit resulting changes in PR for next generation of tests with active bookkeeping changes.

lib/bookkeeping.md
kotp: The `bookkeeping.md` file holds the information that will be read into a
template file, creating the test file.

The alignment in that file should match where it is being read.

As a generated file, for example with Hamming, it will end up doing the
right thing, like this:

```ruby

#!/usr/bin/env ruby
gem 'minitest', '>= 5.0.0'
require 'minitest/autorun'
require_relative 'hamming'

# Test data version:
# deb225e Implement canonical dataset for scrabble-score problem (#255)

class HammingTest < Minitest::Test
  def test_identical_strands
    # skip
    assert_equal 0, Hamming.compute('A', 'A')
  end

  def test_disallow_second_strand_longer
    skip
    assert_raises(ArgumentError) { Hamming.compute('ATA', 'AGTG') }
  end

  # Problems in exercism evolve over time, as we find better ways to ask
  # questions.
  # The version number refers to the version of the problem you solved,
  # not your solution.
  #
  # Define a constant named VERSION inside of the top level BookKeeping
  # module.
  #  In your file, it will look like this:
  #
  # module BookKeeping
  #   VERSION = 1 # Where the version number matches the one in the test.
  # end
  #
  # If you are curious, read more about constants on RubyDoc:


  def test_bookkeeping
    skip
    assert_equal 3, BookKeeping::VERSION
  end
end
```

The file above shows only the first two and last definition, but the
alignment is correct.

kytrinyx:
This looks good to me.

PR #362, opened 2016-05-31
Suggest adding BookKeeping near end of the file

### LINKED ISSUE #360: Discussion: Location of BookKeeping
Is there a way to encourage people to put the BookKeeping module at the bottom of their code rather than at the top? It's a lot easier for reviewers if the first thing they see is the important part of the code.

kotp:
fixes #360


kytrinyx:
Thanks!

PR #364, opened 2016-06-01
Add flatten-array

dougtebay:
I'll be working on this.


dougtebay:
I got the following error message after running bin/local_status_check:
`exercises/flatten-array/flatten_array_test.rb#test_0001_should have the execution bit set [bin/executable-tests-check:7]: Execution bit not set for exercises/flatten-array/flatten_array_test.rb`
I compared my test with others but wasn't able to find any difference between them. I also had a look at the code in bin/local_status_check and tried to read up on execution bits, but with no luck. All other tests are passing.

kotp:
You should be able to do `chmod 744 exercises/flatten-array/flatten_array_test.rb` while in the project root.  Them commit that change.  The test should then pass.

dougtebay:
That did it -- thanks @kotp !

kotp:
This exercise is very simple in Ruby, should it be placed where it is in `config.json`?  Or @kytrinyx should we mention to try not to use built in methods in `x-common/flatten-array.md`?

kotp:
@dougtebay would you like to use the common data to generate the tests?  Let me know.

dougtebay:
@kotp I did use the common data to generate the tests. Regarding the example, I went with the most straight-forward solution to test the test suite but assumed you would probably ask people to avoid those methods when writing their own solutions. In that case, solutions will most likely involve recursion, making the exercise more difficult. I based my placement of it within the config file on where it was ranked for other languages.

kytrinyx:
We could add a `HINTS.md` that suggests that it might be interesting to try to solve it without the usual enumerable methods? I don't know. Or just stick it somewhere towards the beginning (around raindrops?)

kotp:
@kytrinyx in other exercises, we do have a note in the .md that says "If your language provides ..." which may be suitable, generic and in one place?

@dougtebay I see that you based the tests from the common data, but would you like to generate them directly from there?  It requires a little more work, especially the first time around, just to learn how things fit together.  I wasn't sure you had actually generated it dynamically from the data, or had manually done it.  If you have generated it, the generator files can go in this PR as well.  (You may have just forgotten to place them with this, or it may be in another PR, I haven't looked closely yet.)

dougtebay:
@kotp Oh, I misunderstood. I based my tests on the common data but didn't realize there was a way to generate them automatically. Would there be any benefit to generating them, at this point? And if so, how would I go about it? They required some minor editing--in terms of line length, replacing null with nil, etc.--to get the status check to pass, so I'm not sure if the generator would take care of that?

kotp:
Yes, the generator can take care of that.  It is something that you write that takes care of that though.

You can start learning about this a bit more in the [contributing](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#updating-a-generated-test-suite) guide.

Feel free to chat with us in the [gitter room(s)](https://gitter.im/exercism/dev) as well.

kytrinyx:
>  I'm not sure if the generator would take care of that?

Kind of... we hand-code the specific bits of a generator to tailor it to each exercise, but a lot of the generic generator-code already exists, so it's pretty straight forward.

I suggest that we merge this, since it's what we want, and then @dougtebay can take a stab at creating the generator if they feel like it.

>  Would there be any benefit to generating them, at this point?

Yeah, since we have so many languages, a lot of different languages implement the same problems, and then different languages discover different interesting edge cases, and we update the common data. If we have a generator it's really easy to update the ruby implementation when the common data changes.

dougtebay:
@kytrinyx Sounds great to me. I've taken a look at the documentation on generators and would definitely like to try writing one in the near future.

kytrinyx:
Thanks!

PR #369, opened 2016-06-06
Leap: Add failure messages to tests

### LINKED ISSUE #366: Leap: No test failure messages.
When tests fail they give no indication of why they failed.

All the failure message strings are set to ''

Best solution: Add informative failure messages.

Easiest solution: Remove the empty strings so it at least gives the message that there is a failure with no message.

Insti:
Regenerated the leap tests to add in the test failure error messages.

Since the tests themselves have not changed the BookKeeping version number has not been incremented.

Actually fixes #366 (#367 fixed the generator but did not include the updated test file.)

### DIFF
--- a/exercises/leap/example.tt
+++ b/exercises/leap/example.tt
@@ -10,8 +10,8 @@ class Date
     throw "Implement this yourself instead of using Ruby's implementation."
   end
 
-  alias_method :gregorian_leap?, :leap?
-  alias_method :julian_leap?, :leap?
+  alias gregorian_leap? leap?
+  alias julian_leap? leap?
 end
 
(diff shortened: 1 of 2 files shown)


Insti:
Looks like Travis is using a different rubocop than I am, I'll see if I can sort mine out.

kytrinyx:
You may need to say `bundle exec`.

Cohen-Carlisle:
:+1: from me

kotp:
I think we have been talking about putting the BookKeeping module on the bottom of the file, just to keep it out of the way.

Insti:
@kotp ? The BookKeeping module *is* at the bottom of the file.

kotp:
Thanks @Insti  we really appreciate the time you spent and the work done!

PR #370, opened 2016-06-06
Generator: Add message if x-common doesnt exist.

Insti:
Now gives a helpful error message if you run a test generator without
already having the x-common repository set up in the correct location.

(Also: Extracted methods from generate to appease Rubocop.)

### DIFF
--- a/lib/generator.rb
+++ b/lib/generator.rb
@@ -5,6 +5,8 @@ require 'json'
 require 'ostruct'
 
 class Generator
+  METADATA_REPOSITORY = 'x-common'.freeze
+
   attr_reader :name, :cases
   def initialize(name, cases)
     @name = name
@@ -12,7 +14,7 @@ class Generator
   end
 
   def metadata_dir
-    File.expand_path(File.join('..', '..', '..', 'x-common'), __FILE__)
+    File.expand_path(File.join('..', '..', '..', METADATA_REPOSITORY), __FILE__)
   end
 
   def data
@@ -35,13 +37,43 @@ class Generator
     cases.call(data)
   end
 
+  def metadata_repository_missing_message
+    <<-EOM.gsub(/^ {6}/, '')
+
+      '#{METADATA_REPOSITORY}' repository not found.
+      Try running the command:
+        git clone https://github.com/exercism/#{METADATA_REPOSITORY}.git "#{metadata_dir}"
+
+    EOM
+  end
+
   def generate
+    check_metadata_repository_exists
+    generate_test_file
+    increment_version
+    increment_version_in_example
+  end
+
+  def check_metadata_repository_exists
+    unless File.directory?(metadata_dir)
+      STDERR.puts metadata_repository_missing_message
+      fail Errno::ENOENT.new(metadata_dir)
+    end
+  end
+
+  def generate_test_file
     File.open(path_to("#{name.gsub(/[ -]/, '_')}_test.rb"), 'w') do |f|
       f.write ERB.new(File.read(path_to('example.tt'))).result binding
     end
+  end
+
+  def increment_version
     File.open(path_to('.version'), 'w') do |f|
       f.write version + 1
     end
+  end
+
+  def increment_version_in_example
     contents = File.read(path_to('example.rb'))
     File.open(path_to('example.rb'), 'w') do |f|
       f.write contents.gsub("VERSION = #{version}", "VERSION = #{version + 1}")


lib/generator.rb
```diff
@@ -35,13 +37,43 @@ def test_cases
...
   end
 
+  def metadata_repository_missing_message
+    <<EOM
+
+'#{METADATA_REPOSITORY}' repository not found.
+Try running the command:
+  git clone https://github.com/exercism/#{METADATA_REPOSITORY}.git ../#{METADATA_REPOSITORY}
+
+EOM
```
kotp: What do you think about using either the `~EOM` or at least the `-EOM` form of the heredoc?  

The squigly heredoc is relatively new, so may not be correct to use yet.

lib/generator.rb
```diff
@@ -35,13 +37,43 @@ def test_cases
...
+  end
+
   def generate
+    check_metadata_repository_exists
+    generate_test_file
+    increment_version
+    increment_version_in_example
+  end
```
kotp: I asked about maybe giving a choice to inhibit version increment with an environment variable.  It might look something like this:

```ruby
  def generate
    check_metadata_repository_exists
    generate_test_file
    unless ENV['NOINCREMENT']
      increment_version 
      increment_version_in_example
    end
  end
```

kotp: Maybe not related to what you are doing though...

lib/generator.rb
```diff
@@ -35,13 +37,43 @@ def test_cases
     cases.call(data)
   end
 
+  def metadata_repository_missing_message
+    <<EOM
+
+'#{METADATA_REPOSITORY}' repository not found.
+Try running the command:
+  git clone https://github.com/exercism/#{METADATA_REPOSITORY}.git ../#{METADATA_REPOSITORY}
```
kotp: That is a pretty specific command, when there are more than one commands that could work.  And they may not be running in the folder you are thinking.  It may risk some wrong moves here.

Would a better communication be to go to the `xruby/README.md` file?

lib/generator.rb
Insti: That's a new feature which should have it's own issue/PR.

lib/generator.rb
Insti: There are multiple ways that this could work, but I like that this one gives an immediate resolution to the current problem. Just copy/paste that line and you're good to go again.

Good spot on the relative path issue, I'll replace that with an absolute path.

lib/generator.rb
kotp: Yep, that is what I thought too.

lib/generator.rb
kotp: I just haven't used the https form for at least 5 or 6 years... so it is not something that I specifically would copy and paste.

I suspect that at this point, if they are getting to this file, they had to have written the corresponding `bin/generate` file, and so are probably not aware of what is required, so the documentation would be a good reference.

The purpose of the pull request, I think, was if the file doesn't exist, not the entire repository, so it is only one aspect, maybe, of what could be wrong?

kotp:
Are we ready to squash these commits?

Insti:
Squashed.

lib/generator.rb
```diff
@@ -35,13 +37,43 @@ def test_cases
     cases.call(data)
   end
 
+  def metadata_repository_missing_message
+    <<-EOM.gsub(/^ {6}/, '')
+
+      '#{METADATA_REPOSITORY}' repository not found.
+      Try running the command:
+        git clone https://github.com/exercism/#{METADATA_REPOSITORY}.git "#{metadata_dir}"
```
Cohen-Carlisle: Should we provide an SSH command as well?

Insti: Good suggestion, but I prefer keeping the message short. Anyone who has a ssh key set up should know how to change the repository URL.

kotp: We could peek at their .git/config file and find out and report accordingly.  How their 'origin' reference is formed, we will know what their preference is.

It is just an extra push in being personally helpful, and is some information we really can peek at and know.

Some importance to being able to "copy and paste" a command was communicated, so if this is a valued, then there is value to being inclusive, not excluding those that  prefer another form.

Insti: > We could peek at their .git/config file and find out and report accordingly. How their 'origin' reference is formed, we will know what their preference is.

That's a lot of functionality that is not really the responsibility of the generate class.
If someone wants to make another module with this functionality I'd be happy to use it, but it seems like a lot of work for a single line of an error message which hopefully no-one will see anyway.

> there is value to being inclusive, not excluding those that prefer another form.

It's not excluding anybody.

The proposed change is an improvement over the original mystery error message:
```
[snip..]/xruby/lib/generator.rb:19:in `read': No such file or directory @ rb_sysopen - [snip..]/x-common/leap.json (Errno::ENOENT)
[snip stack trace]
```
And provides some immediately actionable clues about what might need to be done to fix the error. The user still has the option to type a different command into their command line, modify the suggested line in any way, or to seek out other information.

Is this a case of "The Perfect being the enemy of the Good"?

kotp: The two stated concerns I was hearing was "short message" and "copy and paste".  The coding change addresses both values.  If they are not worth the code, then this is likely good enough, and that was the intended (though apparently not well communicated) message.

Here is what it may look like though, if the two expressed values are worth the effort...

```ruby
  def metadata_repository_missing_message
    <<-EOM.gsub(/^ {6}/, '')

      '#{METADATA_REPOSITORY}' repository not found.
      Try running the command:
        git clone #{preferred_protocol}#{METADATA_REPOSITORY}.git "#{metadata_dir}"

    EOM
  end

  def preferred_protocol
    { true => 'git@github.com:exercism/',
      false => 'https://github.com/exercism/'
    }[IO.read('.git/config').include?('git@github')] 
# The above line is sloppy, as it does not check for 
# the precise scope of 'origin', and as 'origin' is a 
# common, though arbitrary, name and does not 
# cover 100%.
  end
```

There still may be some advantage to determine set "PROJECT_ROOT" so that it can be used for both the .git/ and the ../x-common folders.

If this is incorporated, then more people can simply copy and paste, which I think is worth a little bit of extra oomph.

If it isn't, then just consider this my "Looks good to me" vote.

Insti:
I think we should merge it as is. 
If anyone then wants to add a protocol detection feature they can do it in another PR.

kotp:
Then I think we have at least two votes for coming in.

Don't let it be misunderstood, I really do appreciate this change, even as it sits, and the conversation.

Cohen-Carlisle:
Brought it home. Thanks, @Insti.

Issue #371, opened 2016-06-07
Gemfile does not specify Ruby version


kotp:
The current version of Ruby for the website is Ruby 2.2.2, should we expect the code to work on a minimum version of Ruby and if so, should we "enforce" this via the `Gemfile`? 

The benefit:

* Better information from Rubocop
* A good place for exercise designers to know what to design and test against

Cohen-Carlisle:
If we include something like this, I think we should do it like this:
```
raise 'Ruby version must be at least 2.1' unless  RUBY_VERSION.to_f >= 2.1
```
and support any Ruby that has not reached end of life.

kytrinyx:
Enforcing this is a bit overkill, to my mind. If older versions of Ruby work, then fine. We could put something in the README to say that we are not explicitly supporting any Ruby versions that have reached end of life.

kotp:
I see, nice placement of 'not' there. :+1: .

What versions of Ruby are we using while maintaining this track?  I find that I am often using Ruby 2.3.1.

Insti:
2.1.5 or 2.2.2 depending on which machine I'm on.

kotp:
It is already being enforced with the Travis CI, though at version 2.1.2,  I want a way to have the local checks use that information as well.  I ran into where the documentation states I can indicate a minimum, but that is where I hit the snag.  

Tests, ran at the CI level, to help decide if a push will pass Travis, before we get the late notification.  Right now, that doesn't necessarily happen, even if you run the local status check tool.

Cohen-Carlisle:
Should we close this issue along with #372, @kotp?

PR #372, opened 2016-06-08
Ensures designers ruby at version 2.1.2 or higher

kotp:
#371 discussion about helping ensure we are reporting on older versions of Ruby for exercise developers.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,5 @@
 source 'https://rubygems.org'
+ruby '>= 2.1.2'
 
 gem 'minitest'
 gem 'rubocop', '0.36.0'


Cohen-Carlisle:
I think the string is being interpreted literally, as in, ruby is now looking for a ruby version called `>= 2.1.2` instead of a ruby version that is greater than or equal to `2.1.2`

Is there a reason for `2.1.2` over say, `2.1.0`?

kotp:
Seems like Travis CI is set to 2.1.x these two should match, so I am lowering the minimum Ruby version to match that rather than what is required for repository `exercism-io`.

Strange, the specification is failing in Travis CI, though Bundle version specification in Gemfile is correctly formatted.  A Travis bug?  Checking for that possibility.

kotp:
The version string is technically correct for Gemfile, it seems Travis is comparing it in a manner than doesn't account for that in the Gemfile.  I think I asked Travis team the same thing, it appears to check it 'stringally' rather than 'versionally'.

kotp:
And I would say "Yes, there is some reason."  But I do not know what the reason is.  Could be an end of life point, would likely have to check the commit history for that file, if it was annotated for the change, or it was a 'convenience' change at that point.

kotp:
Closing due to lack of interest, and it is possible to do this locally, away from the repository.

kytrinyx:
Thanks, @kotp, I think that was the right choice.

PR #373, opened 2016-06-09
Fix 'require_relative' directive in perfect_numbers_test.rb

gchan:
`require_relative 'example'` is in `perfect_numbers_test.rb`.

It should be `require_relative 'perfect_numbers'`.


kotp:
Thank you @gchan

PR #374, opened 2016-06-09
Fix Gigasecond test failure messages.

Insti:
bin/generate-gigasecond was generating empty test failure messages.

In this case the default messages are fine so I have removed from the template the .msg method calls which were returning the empty strings.

Regenerated the gigasecond_test.rb file WITHOUT incrementing the version number.


kotp:
Thanks @Insti

Issue #299, opened 2016-04-19
grade-school tests force one type of solution?


Cohen-Carlisle:
The tests are calling to_h on the user defined class `School`, which heavily suggests using a hash.

I think thinking through how to structure the metadata might help us develop an API into `School` that is more agnostic of the underlying implementation.

Plus, x-common needs metadata for that, so 2 :bird:, 1 :moyai:.

kytrinyx:
I'd love to see this improved to not guide people to a specific solution.

Insti:
#375 is currently a 1 :bird: 1 :moyai: solution which covers the test portion but not the metadata yet.

`students_by_grade` still returns a hash as that seemed like the most sensible data structure to return.

The [Readme](https://github.com/exercism/x-common/blob/411222fe3202860b8866e4d6ae465e1810a5643d/grade-school.md) seems to suggest a more literal style which would return strings, I'm not sure if the output should be changed to match that. (My preference would be to return proper data structures that something else could use to create the strings.)

Insti:
See also: #389, #390

Cohen-Carlisle:
Closing this as referenced PRs have solved this.

PR #375, opened 2016-06-10
Grade school: Make tests less implemetation dependent.

Insti:
The `School` object is now only tested via its public query methods `grades` and `students_by_grade`

### DIFF
--- a/exercises/grade-school/grade_school_test.rb
+++ b/exercises/grade-school/grade_school_test.rb
@@ -4,85 +4,104 @@ require 'minitest/autorun'
 require_relative 'grade_school'
 
 class SchoolTest < Minitest::Test
-  attr_reader :school
-
-  def setup
-    @school = School.new
-  end
-
-  def test_an_empty_school
-    assert_equal({}, school.to_h)
+  def test_empty_grade
+    subject = School.new
+    expected = []
+    assert_equal expected, subject.grade(1)
   end
 
   def test_add_student
-    skip
-    school.add('Aimee', 2)
-    assert_equal({ 2 => ['Aimee'] }, school.to_h)
-  end
-
-  def test_add_more_students_in_same_class
-    skip
-    school.add('Blair', 2)
-    school.add('James', 2)
-    school.add('Paul', 2)
-    assert_equal({ 2 => %w(Blair James Paul) }, school.to_h)
+    subject = School.new
+    assert subject.add('Aimee', 2)
+    expected = ['Aimee']
+    assert_equal expected, subject.grade(2)
   end
 
   def test_add_students_to_different_grades
     skip
-    school.add('Chelsea', 3)
-    school.add('Logan', 7)
-    assert_equal({ 3 => ['Chelsea'], 7 => ['Logan'] }, school.to_h)
+    subject = School.new
+    subject.add('Aimee', 3)
+    subject.add('Beemee', 7)
+    assert_equal ['Aimee'], subject.grade(3)
+    assert_equal ['Beemee'], subject.grade(7)
   end
 
-  def test_get_students_in_a_grade
+  def test_grade_with_multiple_students
     skip
-    school.add('Bradley', 5)
-    school.add('Franklin', 5)
-    school.add('Jeff', 1)
-    assert_equal %w(Bradley Franklin), school.grade(5)
+    subject  = School.new
+    grade    = rand(6)
+    students = %w(Aimee Beemee Ceemee)
+    students.each { |student| subject.add(student, grade) }
+    assert_equal students, subject.grade(grade)
   end
 
-  def test_get_students_sorted_in_a_grade
+  def test_grade_with_multiple_students_sorts_correctly
     skip
-    school.add('Franklin', 5)
-    school.add('Bradley', 5)
-    school.add('Jeff', 1)
-    assert_equal %w(Bradley Franklin), school.grade(5)
+    subject = School.new
+    grade    = rand(6)
+    students = %w(Beemee Aimee Ceemee)
+    students.each { |student| subject.add(student, grade) }
+    expected = students.sort
+    assert_equal expected, subject.grade(grade)
   end
 
-  def test_get_students_in_a_non_existant_grade
-    skip
-    assert_equal [], school.grade(1)
+  def test_empty_students_by_grade
+    subject = School.new
+    expected = {}
+    assert_equal expected, subject.students_by_grade
   end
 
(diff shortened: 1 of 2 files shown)


Insti:
Fixes #299

kotp:
:eyes: :on:

exercises/grade-school/grade_school_test.rb
```diff
@@ -4,85 +4,104 @@
...
-  def test_an_empty_school
-    assert_equal({}, school.to_h)
+  def test_empty_grade
+    subject = School.new
+    expected = []
+    assert_equal expected, subject.grade(1)
   end
 
   def test_add_student
-    skip
```
Cohen-Carlisle: looks like this is now missing its skip

exercises/grade-school/grade_school_test.rb
```diff
@@ -4,85 +4,104 @@
...
   end
 
-  def test_get_students_sorted_in_a_grade
-    skip
-    school.add('Franklin', 5)
-    school.add('Bradley', 5)
-    school.add('Jeff', 1)
-    assert_equal %w(Bradley Franklin), school.grade(5)
+  def test_empty_students_by_grade
+    subject = School.new
```
Cohen-Carlisle: also missing skip here

exercises/grade-school/grade_school_test.rb
```diff
@@ -4,85 +4,104 @@
...
   end
 
-  def test_add_students_to_different_grades
+  def test_grade_with_multiple_students
     skip
-    school.add('Chelsea', 3)
-    school.add('Logan', 7)
-    assert_equal({ 3 => ['Chelsea'], 7 => ['Logan'] }, school.to_h)
+    subject  = School.new
+    grade    = rand(6)
```
Cohen-Carlisle: This will allow `0`, which is a bit odd. Any particular reason to have random grades?

exercises/grade-school/grade_school_test.rb
```diff
@@ -4,85 +4,104 @@
...
-  def test_sort_school # rubocop:disable Metrics/MethodLength
+  def test_students_by_grade_sorted
     skip
-    [
-      ['Jennifer', 4], ['Kareem', 6],
-      ['Christopher', 4], ['Kyle', 3]
-    ].each do |name, grade|
-      school.add(name, grade)
+    subject = School.new
+    everyone.each do |(grade, students)|
```
Cohen-Carlisle: You don't technically need parens here.

exercises/grade-school/grade_school_test.rb
```diff
@@ -4,85 +4,104 @@
...
-  attr_reader :school
-
-  def setup
-    @school = School.new
-  end
-
-  def test_an_empty_school
-    assert_equal({}, school.to_h)
+  def test_empty_grade
+    subject = School.new
```
Cohen-Carlisle: Any particular reason to go with `subject`? It seems less intuitive to me than `school` without even considering the class is named `School`. I think it could be distracting to the user.

Cohen-Carlisle:
@Insti Thanks for this! I really like the direction its going. I left a first pass of comments on the diff.

exercises/grade-school/grade_school_test.rb
kotp: I think a "list" in this context should be a list of students, such as

```text
"Adam\n
Betty\n
Charlie\n
Denise\n"
```
rather than

```ruby
[
  'Adam",
  "Betty",
  "Charlie",
  "Denise",
]
```

Then the implementation is left open, we only test for the result of the query, rather than the structure.

Cohen-Carlisle: I have to disagree with this. The most straightforward way to represent an ordered list in Ruby is with an array. Using strings with delimiters is more work and less useful to the consumer of the query.

Plus, a hash is being used later for `students_by_grade`. I definitely don't want to try to consume that as a string instead of a hash. And if it uses a ruby data structure, why not this?

exercises/grade-school/grade_school_test.rb
Insti: Selecting a random grade is a unit testing technique. Since the exact grade being used doesn't matter for the test, just pick one randomly. This should make it clear (to anyone familiar with the technique anyway) that the exact grade is unimportant in this test.

The problem doesn't specify any limits on grade + given practically anything will work as a hash key I wasn't too worried about filtering out `0`.

exercises/grade-school/grade_school_test.rb
Insti: > Any particular reason to go with `subject`?

This is another testing habit, I always call the object under test `subject` which makes it clear which object is being tested. It just happens here that `subject` is a possible domain specific word so changing it to `school` for this test file is probably a good idea.

Insti: Discussion of the appropriate return values should probably move to #299 rather than continuing here.

exercises/grade-school/grade_school_test.rb
Cohen-Carlisle: Perhaps I'm being overly concerned with simplicity and/or code newbies, but I could see this being confusing.
 - A grade of `0` could be confusing as it doesn't have an obvious meaning.
 - The randomness could also be confusing as failure messages change on reruns.

exercises/grade-school/grade_school_test.rb
kotp: The most straight forward way in Ruby, yes.  I am not referring to Ruby, though, I am referring to the task at hand, from a "readme" point of view.  "Ordered list" from a mathematics view, from a programming view, or from a "Teacher in a school" view?  I most often see them on a piece of paper, as written list of names.  How they got that way, I don't concern myself, as a teacher, just that the report is correct.  This removes the worry of "how was it implemented" but when we test directly against implementation, it can make things more restricted.  The changes here are driven from "are we enforcing a single implementation" and so this is just me rambling with that in the back of my mind.

We have done it before, using to_s t show what is expected in other exercisms.  This frees up the implementation, because you are basing the test on a final report, rather than the specifics of a return in implementation.

The reference to a hash is being used later for students by grade, true.  But it doesn't have to be that. 

It is just me rambling, just reflecting on what I see.  Not that actions need to be taken on it.  It is likely good enough.

Cohen-Carlisle:
:+1: from me.

kotp:
Thanks for the hard work and thanks @Cohen-Carlisle for the feedback.

PR #377, opened 2016-06-16
gigasecond and space-age are 2nd and 3rd exercise

Cohen-Carlisle:
`gigasecond` and `space-age` are extremely simple exercises that should be directly after `hello_world`

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,8 +6,9 @@
   "gitter": "xruby",
   "problems": [
     "hello-world",
-    "hamming",
     "gigasecond",
+    "space-age",
+    "hamming",
     "rna-transcription",
     "raindrops",
     "difference-of-squares",
@@ -34,7 +35,6 @@
     "trinary",
     "beer-song",
     "bowling",
-    "space-age",
     "anagram",
     "binary-search-tree",
     "crypto-square",


Insti:
I think it risks being a little boring if gigasecond and space-age are next to each other, can we put something that exercises some different skills in between them?

Insti:
Implementing space age also includes some interesting metaprogramming options which I think makes it a better candidate for being further down the list, so as to not confuse new programmers with metaprogrammed solutions.

Cohen-Carlisle:
> I think it risks being a little boring if gigasecond and space-age are next to each other, can we put something that exercises some different skills in between them?

This is a good point. I made this mostly from a newbie perspective. Putting these simple exercises upfront is optimal for them as they don't need to learn a lot of new concepts at once. However, it does risk boring a more advanced user. Though honestly I think 2 very simple exercises after `hello-world` is probably fine and not overly "boring".

> Implementing space age also includes some interesting metaprogramming options

I'm a big fan of keeping the code simple, so I didn't even really think of metaprogrammatic solutions. In fact short of something like defining them based on a hash of planets and seconds, I'm not really sure what you would do. Of course, this maybe keeps it more interesting for more advanced users.

> I think [the metaprogramming option] makes it a better candidate for being further down the list, so as to not confuse new programmers with metaprogrammatic solutions.

Hmm, I don't think I share this concern. I don't see newer programmers even considering metaprogrammatic solutions and I think they are less likely to browse solutions. And if they do find metaprogrammatic solutions, I think it would be a good learning experience. Either they can learn a bit of metaprogramming or learn that they aren't ready for all language features yet :smile:

kotp:
The third point is, in my opinion, comes from having reviews and the discussion started on the site.

I don't feel strongly about the move, either way.

Does this end up being a "no op"?

Cohen-Carlisle:
Time to tally the thumbs?

kytrinyx:
I've seen a bunch of interesting metaprogramming approaches to the space age problem. I would support putting it a bit further down in the list.

kotp:
@kytrinyx further down the list than the 2nd position?  Or further down than it currently is?  Or where it currently is is good, so this becomes a "won't merge"?

kytrinyx:
Oh, good questions. I meant further down than the 2nd position. I actually think that the current position is pretty good, but I'm open to discussing that.

Cohen-Carlisle:
![:-1:] (http://i.imgur.com/SiN92.gif)

PR #380, opened 2016-06-19
Trinary: Test for 4, as invalid trinary value

kotp:
Solutions that check for decimals in "normalization" should be stricter
than simply checking for decimals.

### DIFF
--- a/exercises/trinary/trinary_test.rb
+++ b/exercises/trinary/trinary_test.rb
@@ -57,4 +57,28 @@ class TrinaryTest < Minitest::Test
     skip
     assert_equal 0, Trinary.new("Invalid\n201\nString").to_decimal
   end
+
+  def test_number_out_of_range
+    skip
+    assert_equal 0, Trinary.new('4').to_decimal
+  end
+
+  # Problems in exercism evolve over time, as we find better ways to ask
+  # questions.
+  # The version number refers to the version of the problem you solved,
+  # not your solution.
+  #
+  # Define a constant named VERSION inside of the top level BookKeeping
+  # module.
+  #  In your file, it will look like this:
+  #
+  # module BookKeeping
+  #   VERSION = 1 # Where the version number matches the one in the test.
+  # end
+  #
+  # If you are curious, read more about constants on RubyDoc:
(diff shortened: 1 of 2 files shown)


Insti:
Looks fine.

Probably unrelated issue: Shouldn't Trinary.new('4') raise an error rather than something that allows to_decimal to be called on it?

Update: Ok I've read the README, I still think it should raise an error though.

exercises/trinary/trinary_test.rb
```diff
@@ -52,4 +52,28 @@ def test_invalid_trinary_with_digits_is_decimal_0
     skip
```
Insti: Another extra thing: 
Should this function be called: `test_invalid_trinary_with_invalid_digits_is_decimal_0`

kotp: Definitely an incomplete title.

kotp:
I think it is consistent behavior that '4' results in 0 as well.

* Readme
* Other tests, i.e., 'carrot' -> 0, '0a1b2c' -> 0
* Ruby's own default coercion: i.e., `String#to_i`, `String#to_f`, etc.

Insti:
I agree it's consistent.

kotp:
Brought in @Insti change for the multiline strings, as it is appropriate.  /cc @exercism/xruby

exercises/trinary/trinary_test.rb
```diff
@@ -52,4 +52,33 @@ def test_invalid_trinary_with_digits_is_decimal_0
...
+  #  In your file, it will look like this:
+  #
+  # module BookKeeping
+  #   VERSION = 1 # Where the version number matches the one in the test.
+  # end
+  #
+  # If you are curious, read more about constants on RubyDoc:
+  # http://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/constants.htm
+  def test_bookkeeping
+    assert_equal 1, BookKeeping::VERSION
```
Insti: Would version 2 be more appropriate? Or is there a convention that the unversioned version is always 0?

kotp: I have been using the no version indicated is implied 0, and once we start keeping track, it becomes version 1.

Cohen-Carlisle:
:+1:

PR #382, opened 2016-06-20
Trinary: Test with multiline strings.

Insti:
Many trinary solutions use a regex: /^[012]+$/ to check for trinary
validity which can be 'fooled' by strings which contain newlines.

This adds a test for that case.

(It also fixes the example solution which had that bug.)


kotp:
Brought it in under #380

PR #383, opened 2016-06-22
Generated Custom Set

### LINKED ISSUE #348: Changes to Custom Set tests
We recently rewrote the test suite for Custom Set. Since this track implements Custom Set, please [take a look at the new custom_set.json file](https://github.com/exercism/x-common/blob/master/custom-set.json) and see if your track should update its implementation or tests.

The new test suite reorders tests so that students can get to green quickly. It also reduces the number of tests so that students can focus on solving the interesting edge cases.

More details on this change are available in [the pull request](https://github.com/exercism/x-common/pull/232)

### LINKED ISSUE #365: Updated tests for the Custom Set problem
In order to reduce the amount of code required to pass incremental tests (assuming that users pass tests starting from the top), the order of the tests was modified slightly.

Since this track implements Custom Set, please take a look at the new custom-set.json file and see if your track should update its tests.

If you do need to update your tests, please refer to this issue in your PR. That helps us see which tracks still need to update their tests.

If your track is already up to date, go ahead and close this issue.

More details on this change are available in https://github.com/exercism/x-common/pull/257.

kotp:
The custom-set exercise is generated using data from x-common
repository.

fixes #365
fixes #348
references exercism/x-common#257


exercises/custom-set/custom_set_test.rb
```diff
@@ -3,114 +3,303 @@
 require 'minitest/autorun'
 require_relative 'custom_set'
 
+# rubocop:disable Metrics/LineLength
```
Cohen-Carlisle: Hmm, I don't like this. The offending lines seem overly verbose. Would it be advisable to shorten them up in the x-common json?

lib/custom_set_cases.rb
```diff
@@ -0,0 +1,89 @@
+class CustomSetCase < OpenStruct
+  def name
+    'test_%s' % description.gsub(/ |-/, '_')
+  end
+
+  def test_body
+    instance_eval section
```
Cohen-Carlisle: Yikes, an eval! What is `section` here?

lib/custom_set_cases.rb
```diff
@@ -0,0 +1,89 @@
...
+  def skipped
+    index.zero? ? '# skip' : 'skip'
+  end
+end
+
+CustomSetCases = proc do |data|
+  i = 0
+  json = JSON.parse(data)
+  cases = []
+  p json.keys
```
Cohen-Carlisle: Leftover debugging code.

lib/custom_set_cases.rb
```diff
@@ -0,0 +1,89 @@
...
+  (json.keys - ['#']).each do |section|
+    json[section]['cases'].each do |row|
+      row = row.merge(row.merge('index' => i, 'section' => section))
```
Cohen-Carlisle: Whoa. Shadowing the original block variable with a double merged version....
It looks like the outermost merge is unnecessary... and how about using `row_merged` or something over `row` here and after?

exercises/custom-set/custom_set_test.rb
kotp: Yes, it does make sense to attack this in the common file.  I did not like it when I did it either.  I am not addressing it here though, as it is not the 'source'.  The source is the linked issues, I think I have linked them in the PR.

lib/custom_set_cases.rb
kotp: An instance_eval, but yes.  Section is the hash being generated from the JSON.  Should use parse here.  It was definitely a 'get it done' and 'dirty'.

Insti: Could you use `send` instead of `instance_eval` ?
This makes it clearer that you're just calling one of the methods of `CustomSetCase` and you no longer have to worry too much if you're ever going to get something nasty in through `section`

lib/custom_set_cases.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def test_body
+    instance_eval section
+  end
+
+  def union
+    "set1 = CustomSet.new #{set1}
+    set2 = CustomSet.new #{set2}
+    expected = CustomSet.new(#{expected})
+    #{assert_or_refute}_equal set1.union(set2), expected"
```
Insti: I like how you're assigning expected to a variable, but it appears that you have expected and actual inverted which will cause confusing test failure messages.
(A lot of the test templates in this file have this issue.)  

The [signature for assert_equal](http://docs.ruby-lang.org/en/2.1.0/MiniTest/Assertions.html#method-i-assert_equal) is: `assert_equal(exp, act, msg = nil)`

exercises/custom-set/custom_set_test.rb
Insti: This file is generated and the lines are only long because the test names are verbose so I have no problem with disabling the LineLength metric here.

lib/custom_set_cases.rb
```diff
@@ -0,0 +1,85 @@
+class CustomSetCase < OpenStruct
+  def name
+    'test_%s' % description.gsub(/ |-/, '_')
+  end
+
+  def test_body
+    send section
```
kotp: I was trying to attack this vector from the other end... This is much simpler... Thanks @Insti !

lib/custom_set_cases.rb
kotp: I took this from Clock cases, and will need to have that cleaned up as well.  Looks much better now, I hope.

lib/custom_set_cases.rb
```diff
@@ -0,0 +1,85 @@
...
+  def skipped
+    index.zero? ? '# skip' : 'skip'
+  end
+end
+
+CustomSetCases = proc do |data|
+  json = JSON.parse(data)
+  cases = []
+  (json.keys - ['#']).each do |section|
+    json[section]['cases'].each_with_index do |row, i|
```
Cohen-Carlisle: I believe this `each_with_index` should be simply an `each` while the above line's `each` should be the `each_with_index`. This is currently leaving the first exercise of each type (diff, union, etc.) unskipped.

Cohen-Carlisle:
Noticed one more thing about skips but other than that I think it looks good.

kotp:
Manual index had to go back in, if that is the only thing with the skips, I think we are ready to merge.

lib/custom_set_cases.rb
kotp: @Cohen-Carlisle Thanks for keeping an eye on this...

exercises/custom-set/custom_set_test.rb
```diff
@@ -3,114 +3,303 @@
...
+    # skip
+    set = CustomSet.new []
+    assert_equal CustomSet.new, set
+  end
+
+  def test_sets_with_elements_are_not_empty
+    skip
+    set = CustomSet.new [1]
+    refute_equal CustomSet.new, set
+  end
```
Cohen-Carlisle: Shouldn't the empty tests use the `empty?` method? As written, these are really testing equality and not emptiness.

exercises/custom-set/custom_set_test.rb
```diff
@@ -3,114 +3,303 @@
...
+    set2 = CustomSet.new []
+    assert set2.subset? set1
+  end
+
+  def test_empty_set_is_a_subset_of_non_empty_set
+    skip
+    set1 = CustomSet.new []
+    set2 = CustomSet.new [1]
+    assert set2.subset? set1
+  end
```
Cohen-Carlisle: These subset tests seem backwards to me. I read `set2.subset? set1` as Is set2 at subset of set1?
Looking at Ruby's `Set` class, I see it also handles it this way.

exercises/custom-set/custom_set_test.rb
```diff
@@ -3,114 +3,303 @@
...
+    assert set1.disjoint? set2
+  end
 
-    assert_equal CustomSet.new([1, 2.0, 3]),
-                 CustomSet.new([1, 2.0, 3]).difference(CustomSet.new([2, 4]))
+  def test_empty_sets_are_equal
+    skip
+    set1 = CustomSet.new []
+    set2 = CustomSet.new []
+    assert_equal set1, set2
```
Cohen-Carlisle: This is the right place to test equality. Note that the first test (about emptiness) is actually more complex than this as its a duplicate except it requires no args to `new` to be treated the same as an empty array to new.

exercises/custom-set/custom_set_test.rb
```diff
@@ -3,114 +3,303 @@
...
-  def test_put
+  def test_add_to_non_empty_set
     skip
-    assert_equal CustomSet.new([1, 2, 3, 4]),
-                 CustomSet.new([1, 2, 4]).put(3)
+    set = CustomSet.new [1, 2, 4]
+    element = 3
+    expected = CustomSet.new [1, 2, 3, 4]
+    assert_equal expected, set.put(element)
+  end
```
Cohen-Carlisle: Something about this seems a little awkward to me. As I read each line I don't understand why its important. Perhaps something like?
```
set1 = CustomSet.new [1, 2, 3, 4]
set2 = CustomSet.new [1, 2, 4]
assert_equal set1, set2.put(3)
```

exercises/custom-set/custom_set_test.rb
```diff
@@ -3,114 +3,303 @@
...
+    set2 = CustomSet.new [3, 2, 5]
+    expected = CustomSet.new [2, 3]
+    assert_equal expected, set2.intersection(set1)
+  end
+
+  def test_difference_of_two_empty_sets_is_an_empty_set
+    skip
+    set1 = CustomSet.new []
+    set2 = CustomSet.new []
+    expected = CustomSet.new([])
```
Cohen-Carlisle: As a matter of style we haven't used parens for `CustomSet.new` until here.

Cohen-Carlisle:
Looked more closely and had more comments 😛

exercises/custom-set/custom_set_test.rb
kotp: This comment is probably more directed to the x-common repository for that .json file?

exercises/custom-set/custom_set_test.rb
kotp: Also from the point of view of the common data?

exercises/custom-set/custom_set_test.rb
kotp: Noted, I will look in the cases and see if I added it in the method.  I wonder why Rubocop wouldn't have picked this up.

exercises/custom-set/custom_set_test.rb
Cohen-Carlisle: Not necessarily, you are still consuming the data. For instance, I was thinking the above code (revised slightly) would correspond to the generator code:
```ruby
  def add
    "set = CustomSet.new #{set}
    expected = CustomSet.new #{expected}
    #{assert_or_refute}_equal expected, set.put(#{element})"
  end
```
So actually, I think the thing I didn't like was declaring `element` separately.

exercises/custom-set/custom_set_test.rb
Cohen-Carlisle: The problem isn't here, it was really a follow up comment to:
> Shouldn't the empty tests use the `empty?` method? As written, these are really testing equality and not emptiness.

So then I followed up with:
> This is the right place to test equality. Note that the first test (about emptiness) is actually more complex than this as its a duplicate except it requires no args to new to be treated the same as an empty array to `new`.

exercises/custom-set/custom_set_test.rb
kotp: The test assert_empty? requires that the object responds to empty? and there is no spec for that currently.

Cohen-Carlisle: I believe this should be that very spec. The metadata for these exercises is:
```json
   "empty": {
     "description": "Returns true if the set contains no elements",
     "cases": [
        "..."
     ]
   },
```
This is not the place to rely on equality. Equality has its own section later on.

Cohen-Carlisle:
There seems to be some confusion over whether to use `empty?`. In my mind, there is no question. The metadata intends for the first exercises to have an `empty?` method first and much later to implement an equality method.

kotp:
If I am following the leads properly, it seems that there no common data that dictates a test that forces a method to be defined strictly named "empty?".  That is an implementation that may be wise, but at the moment I am not seeing that it is forced.  The same with equality.

The metadata that I am looking for is specifically 'example.rb', and 'custom-set.json', and nothing else.  The restriction is "refactoring" level, so no changes in behavior, hopefully.  The goal for this PR is simply to bring in the tests from x-common repository.  Definitely not to make improvements in the testing at this time.

I may not be looking at it from the proper point of view though.

If I have accomplished this, with this goal in mind, then I think we can bring it in, and then address the "we should have the implementation have `empty?` and `equals` defined in the track" for this problem set.

I think you are correct about the equality and the empty being defined... and an "empty?" method being desirable.  Especially since "empty" is defined, and so a query partner for that method makes even more sense, because of it.

Insti:
There should **not** be an `empty?` method.

There is an `empty` section in `custom-set.json` but there are also a lot of other sections in that file that act as descriptions of what is tested within that section and not specifications of query methods.

The tests that @kotp has implemented accurately reflect the intent of the common data and the problem description.

Cohen-Carlisle:
I strongly disagree. Looking at the PR that put the current ordering into place (https://github.com/exercism/x-common/pull/257), we see the comment:
> The goal of this change is to ensure that no tests that require set equality in an assertion precede the tests for set equality. This allows each set of tests to introduce as little new functionality as possible (assuming that tests are completed from top to bottom).

Insti:
I was wrong.

Having re-read the referenced thread and reviewed the json file I now agree with @Cohen-Carlisle 
and believe an `empty?` method **is** required.

exercises/custom-set/custom_set_test.rb
```diff
@@ -3,114 +3,299 @@
...
+    skip
+    set = CustomSet.new [1]
+    refute_equal CustomSet.new, set
+  end
+
+  def test_nothing_is_contained_in_an_empty_set
+    skip
+    set = CustomSet.new []
+    element = 1
+    refute set.member? element
```
Insti: `member?` should be `contains?` 

`member` is not referenced in the custom-set.json but `contains` is.

lib/custom_set_cases.rb
```diff
@@ -0,0 +1,86 @@
...
+    "set1 = CustomSet.new #{set1}
+    set2 = CustomSet.new #{set2}
+    expected = CustomSet.new #{expected}
+    #{assert_or_refute}_equal expected, set2.intersection(set1)"
+  end
+
+  def add
+    "set = CustomSet.new #{set}
+    expected = CustomSet.new #{expected}
+    #{assert_or_refute}_equal expected, set.put(#{element})"
```
Insti: `put` should be `add`

exercises/custom-set/custom_set_test.rb
Cohen-Carlisle: I would actually make an argument for `include?` as its the most common duck typed method for this kind of thing in Ruby. I don't think we have to stick to the description exactly, especially when ruby has a clear pattern for this kind of query.

kotp: The names in section don't necessarily have to match the methods, they are section headers for common-data.  But I changed it in the example code anyway.

lib/custom_set_cases.rb
kotp: Changed.

Cohen-Carlisle:
Two of my comments seem to have been buried after unrelated diff changes.

1) The `subset?` tests are backwards. For example, we have a test that reads like this when variables are in-lined:
```ruby
    assert CustomSet.new([1]).subset? CustomSet.new([])
```
Read in English, this reads:
> assert that a set with the element `1` is a subset of an empty set

Which is actually backwards. [Ruby's standard library](http://ruby-doc.org/stdlib-2.3.1/libdoc/set/rdoc/Set.html#method-i-subset-3F) agrees.

2) I favor the use of `include?` over `contains?` as `include?` is the ruby standard for this. Illustrated in [Enumerable](http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-include-3F) and [Array](http://ruby-doc.org/core-2.3.1/Array.html#method-i-include-3F).

Cohen-Carlisle:
Also, https://github.com/exercism/xruby/issues/386 applies here. Not sure if we want to hold it up over that or fix it later.

kotp:
I think it was taken care of, if this is the code that you are thinking of.

```ruby
  def test_empty_set_is_a_subset_of_non_empty_set
    skip
    set1 = CustomSet.new []
    set2 = CustomSet.new [1]
    assert set2.subset? set1
  end

  def test_non_empty_set_is_not_a_subset_of_empty_set
    skip
    set1 = CustomSet.new [1]
    set2 = CustomSet.new []
    refute set2.subset? set1
  end
```

Insti:
1) @kotp I'm confused.
The code you have quoted looks wrong to me.
In `test_empty_set_is_a_subset_of_non_empty_set` `set2` is not a subset of `set1`
I think this is what @Cohen-Carlisle is also saying.

2) Using `include?` seems reasonable, I don't have particularly strong opinions on method naming in this case. It is nice to know that matching the names in the json is not required.

kotp:
When you use a named test in minitest it can use a specific method.  When you use equals, then you use whatever you want to compare together.  This is why using them will sometimes expose that a certain method is not used, due to internals, at least that is what I am seeing when I apply suggested changes with specific method names.

In the code that appears wrong, feel free to clone the PR branch and play with it, and propose a solution... because I am not seeing what you guys are seeing, or I am not properly recreating the situation.  I have changed it over and over again, and get feedback based on the actual runs.

Cohen-Carlisle:
I'm unfortunately pretty busy right now and will be for a while.

I was trying to say that the logic around the receiver and argument should be swapped. Instead of `Set[1].subset? Set[] # => true`, it should be `Set[].subset? Set[1] # => true`.

In other words, it should be implemented as "is the receiver a subset of the argument?", while its currently "is the argument a subset of the receiver?"

kotp:
I had changed it I think twice now... Going from

```ruby
  def test_empty_set_is_a_subset_of_non_empty_set
    skip
    set1 = CustomSet.new []
    set2 = CustomSet.new [1]
    assert set2.subset? set1
  end

  def test_non_empty_set_is_not_a_subset_of_empty_set
    skip
    set1 = CustomSet.new [1]
    set2 = CustomSet.new []
    refute set2.subset? set1
  end
```
to
```ruby
  def test_empty_set_is_a_subset_of_non_empty_set
    skip
    set1 = CustomSet.new []
    set2 = CustomSet.new [1]
    assert set1.subset? set2
  end

  def test_non_empty_set_is_not_a_subset_of_empty_set
    skip
    set1 = CustomSet.new [1]
    set2 = CustomSet.new []
    refute set1.subset? set2
  end
```

Insti:
The second one looks correct, but does not match the code in the currently committed `custom_set_test.rb` file.
Do you have a version that has not been merged into this PR?

kotp:
I do not have a version that has not been merged in.  Would have to dig through the outdated ones to see.  But I am pretty sure I had done this a few times and we keep going back and forth.

kotp:
Oh, "when I get a comment that it is incorrect" may be comments from the tests, not from a human.  Perhaps the example.rb is not implemented correctly/backwards.

Insti:
The `example.rb` implementation is incorrect.

exercises/custom-set/example.rb
```diff
@@ -36,8 +40,9 @@ def intersection(other)
   def member?(datum)
     data.any? { |node| node.datum.eql?(datum) }
   end
+  alias include? member?
```
kotp: Enumerable aliases `member?` to `include?` while internally, it uses `member`.  Standard library `Set` has Ruby method `include?` which it aliases to `member?`.

Insti:
I think the most recent change fixed most of the problems we had with this PR.
I vote we merge it and then keep an eye on it to see if anyone reports any problems.

kotp:
Thanks @Cohen-Carlisle and @Insti the goal of at least having it generated is now complete.  Now we can focus on any changes that come out, based on common data.

PR #389, opened 2016-07-01
Keys order doesn't matter in Hash equality

srabuini:
http://ruby-doc.org/core-2.3.1/Hash.html#method-i-3D-3D

So, I would add an assertion to ensure that Hashes are equal plus keys are in the same order.

Have a look please and let me know,

Thanks!

### DIFF
--- a/exercises/grade-school/example.rb
+++ b/exercises/grade-school/example.rb
@@ -1,5 +1,5 @@
 module BookKeeping
-  VERSION = 2
+  VERSION = 3
 end
 
 class School
--- a/exercises/grade-school/grade_school_test.rb
+++ b/exercises/grade-school/grade_school_test.rb
@@ -71,6 +71,7 @@ class SchoolTest < Minitest::Test
     end
     expected = everyone_sorted
     assert_equal expected, school.students_by_grade
+    assert_equal expected.keys, school.students_by_grade.keys
   end
 
   def everyone
@@ -104,6 +105,6 @@ class SchoolTest < Minitest::Test
 
   def test_bookkeeping
     skip
-    assert_equal 2, BookKeeping::VERSION
+    assert_equal 3, BookKeeping::VERSION
   end
 end


kotp:
Can you update the example.rb file and incrememt the version for the bookkeeping?

srabuini:
Done!

kotp:
Looks good to me,  Anyone else?

Insti:
Well spotted @srabuini, but I don't like adding this test.
**Because** key order doesn't matter in hash equality.

If it is imperative that the key ordering is important, a different (more complicated) data structure should be returned by `students_by_grade`

tejasbubane:
I agree with @Insti. We should not rely on the fact that a hash is being returned and make our tests implementation-specific. How about `assert_equal expected, school.grade(grade)` for each key (as grade) in the `everyone` hash?

Insti:
I think we need to decide if `students_by_grade` should be returning the hash as it does currently, (In which case the `test_students_by_grade_sorted` test should be merged with `test_students_by_grade` and the sort criteria ignored) OR `students_by_grade` should return an array of hashes that looks something like this:
```
    [ 
      {grade: 1, students: %w(Effmee Geemee)},
      {grade: 2, students: %w(Aimee Beemee Ceemee)},
      {grade: 3, students: %w(Deemee Eeemee)},
    ]

```

Reading the grade school description: https://github.com/exercism/x-common/blob/master/grade-school.md would seem to suggest that the array of hashes is the desired solution.

TODO: Check what other languages do.

kotp:
I have been wrestling with that too.  It does not necessary force a hash solution, but does force a solution that is able to have `keys` called on it.

It is also not currently an exercise that has common test data.  Would be interesting to see how the tracks that use this are testing.

Insti:
> but does force a solution that is able to have `keys` called on it.

I don't understand why this would be a requirement.

kotp:
I think you are right, since keys were called on expected, not on the actual.

tejasbubane:
Also, I think (with the existing tests), the name `school.grade(1)` is a little confusing since it accepts grade as an argument and returns a list of students. Should be something like `school.students(grade)`. [Example code uses `level` and `grade` interchangeably](https://github.com/exercism/xruby/blob/master/exercises/grade-school/example.rb#L11-L15).

srabuini:
I agree. What do you think about this:

```
assert_equal expected.keys, school.students_by_grade.map { |k, _v| k }
```

Insti:
I has a look at how some of the [other languages that implement this problem](http://x.exercism.io/v3/problems/grade-school/todo) handle the result of `students_by_grade`

* ruby - (unordered map)
* cpp - (unordered map) 
* javascript - (unordered map) object/hash
* java - (unordered map) 
* python - (ordered list) - But adds them in sorted order anyway, and subverts it's own test for sortedness. 
* fsharp - (ordered list) 
* elixir - (ordered list)
* elm    - (ordered list) - of tuples

It seems to be about 50/50 among the languages I looked at.

Insti:
Since I prefer the ordered list of hashes solution I've created a new PR #390 which implements this.

Insti:
Anyone want to comment on pull request #390?
Is it :+1: or :-1:, does it need more work?

kotp:
@srabuini Thanks for the work on this, and bring attention to this.

PR #395, opened 2016-07-15
Accumulate: Test case no block is passed

duffuniverse:
It is typical to call [#to_enum](http://ruby-doc.org/core-2.3.1/Object.html#method-i-enum_for) when defining methods for a generic Enumerable, in case no block is passed.

exercises/accumulate/example.rb
```ruby
...
return to_enum(__method__) unless block_given? # __method__ is :accumulate here
...
```


Insti:
Thanks @duffuniverse, I think this is a good test and would be useful in an idiomatic Ruby solution, 
but given that the [README](https://github.com/exercism/x-common/blob/master/accumulate.md) lists the restrictions:

> Keep your hands off that collect/map/fmap/whatchamacallit functionality
> provided by your standard library!
> Solve this one yourself using other basic tools instead.

I'm not sure that this is appropriate behaviour to be testing for in this case. I'd file Enumerable under "whatchamacallit functionality" and I think it's OK (and in fact preferred) to leave the no-block behaviour undefined for this problem.

What do other people think?

duffuniverse:
@Insti I see

kotp:
Should the return be formalized in a test? That not given a block, it returns something?

kytrinyx:
What about adding a `HINTS.md` file that suggests some things to think about, without mandating an approach? This would get included in the README, and people might use that as a jumping-off point for research or discussions.

Insti:
@duffuniverse, did you want to turn your patch and first comment here into a `HINTS.md` file in `exercises/accumulate` that we can use when generating the README.md file?

duffuniverse:
Sure

kotp:
You would need to write it... push it up with this PR and feedback will be given.

It could have a level 2 heading and perhaps look like this:

```mardkown
## Hints:

1) How would you handle, if at all, a block not being given?
2) How does the language built in methods handle this?
```
Or whateever you feel would go in there.  As @kytrinyx mentioned, she would consider this an addendum to the Readme for the track.

Insti:
There doesn't seem to be an already existing `HINTS.md` file anywhere, but it is just a markdown file.
It will be automatically incorporated into the README.md file by the Exercism API

Here's a rough draft:
filename: `exercises/accumulate/HINTS.md`
```
## Advanced

In Ruby it is typical to call #to_enum when defining methods for a generic Enumerable, in case no block is passed.

Here is an additional test you could add:

  def test_no_block_is_passed
    skip
    result = [1, 2, 3].accumulate
    assert_instance_of Enumerator, result
  end
```

duffuniverse:
Thank you

Insti:
Looks good to me :+1:

HINTS.md
```diff
@@ -0,0 +1,24 @@
...
+
+For example:
+
+```ruby
+module Enumerable
+  def some_method
+    return to_enum(__method__) unless block_given? # __method__ is :some_method here
+    ...
+  end
+end
```
kotp: Should the example be less 'concrete'?  A location where to find out about this kind of thing may be long term more productive.

The "How to do it" is part of the exercise.

duffuniverse: No problem

kotp:
This should be rebased (perhaps interactively, did not review that closely)... but when this is done, the things to consider are:

```
Applying: Test case no block is passed
Using index info to reconstruct a base tree...
M       exercises/accumulate/accumulate_test.rb              # be aware of white space...
<stdin>:13: trailing whitespace.
    assert_instance_of Enumerator, result 
warning: 1 line adds whitespace errors.
Falling back to patching base and 3-way merge...             # This is likely due to the merge from master
Auto-merging exercises/accumulate/accumulate_test.rb # But it should not be a problem
Applying: Implement behaviour for case no block is passed
Applying: Add HINTS.md file
```

duffuniverse:
Rubocop didn't detect any offense

Insti:
The branch looks OK to me?

duffuniverse:
:+1:

kotp:
Nope, Rubocop did not, git did directly... It is probably my git settings letting me know.
It's all good.  Just was an "be aware" kind of message.  No problems at all.

Insti:
Thanks for all your help @duffuniverse, I'm glad we could find a way to include the information in the exercise even if it wasn't directly in the test cases.

duffuniverse:
I'm glad too. Have a good day!

matthew-campbell:
The current test allows an invalid implementation of `to_enum`. 

See: http://exercism.io/submissions/8203db3518d04ecdaa7678b0a8c01989

kytrinyx:
@matthew-campbell Would you please open an issue about this describing the problem and (potentially) what you see as a solution? This issue is 18 months old, and there's no reason to believe that anyone is still tracking it.

PR #401, opened 2016-07-22
Move food-chain further down the order. (Closes #399)

### LINKED ISSUE #399: Move food-chain exercise to later?
I noticed that about half of the people that complete couple solutions before food-chain do not complete food-chain.  I was wondering if it would be better to move that exercise to later in the order.

Insti:
Based on solution numbers, many of the people who were completing the problems earlier in the list were stopping when hitting the food-chain problem. This patch moves the food-chain problem much later in the order.

There was no scientific method used in determining the new positioning, I just read through the list of problems and found a place that sounded about right.

If anyone has any arguments about the new placement or wishes to do any scientific analysis of problem positioning feel free, but barring that I suggest we merge this change and if the chosen position turns out to be inappropriate in the future to we can always accept another PR to move it somewhere else.

See also: issue #399


Insti:
Question: Will moving a problem later cause `exercism fetch` to skip problems between the old an new positions?

kotp:
No, I have not seen that problem happen, I sometimes solve them out of order and so far so good.

kotp:
I am OK with this, I see it moved about 30 exercises down.

kytrinyx:
> Question: Will moving a problem later cause exercism fetch to skip problems between the old an new positions?

Ditto on @kotp's observation. I made sure to implement "next" in a flexible way so that it lets you solve things out of order, and also lets us reorder things without affecting people's experience.

PR #403, opened 2016-07-23
[WIP] add all-your-base

sebarmano:
I'll will be working on this exercise.

### DIFF
--- a/config.json
+++ b/config.json
@@ -83,7 +83,8 @@
     "perfect-numbers",
     "connect",
     "list-ops",
-    "diamond"
+    "diamond",
+    "all-your-base"
   ],
   "deprecated": [
     "octal",
--- a/exercises/all-your-base/all_your_base_test.rb
+++ b/exercises/all-your-base/all_your_base_test.rb
@@ -1,0 +1,303 @@
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'all_your_base'
+
+class AllYourBaseTest < Minitest::Test
+  def test_single_bit_to_one_decimal
+    digits = [1]
+    input_base = 2
+    output_base = 10
+    expected = [1]
+
+    converted = BaseConverter.convert(input_base, digits, output_base)
+
+    assert_equal expected, converted,
+                 "Input base: #{input_base}, output base #{output_base}. " \
+                 "Expected #{expected} but got #{converted}."
+  end
+
+  def test_binary_to_single_decimal
+    skip
+    digits = [1, 0, 1]
+    input_base = 2
+    output_base = 10
+    expected = [5]
+
+    converted = BaseConverter.convert(input_base, digits, output_base)
+
+    assert_equal expected, converted,
+                 "Input base: #{input_base}, output base #{output_base}. " \
+                 "Expected #{expected} but got #{converted}."
+  end
+
+  def test_single_decimal_to_binary
+    skip
+    digits = [5]
+    input_base = 10
+    output_base = 2
+    expected = [1, 0, 1]
+
+    converted = BaseConverter.convert(input_base, digits, output_base)
+
+    assert_equal expected, converted,
+                 "Input base: #{input_base}, output base #{output_base}. " \
+                 "Expected #{expected} but got #{converted}."
+  end
+
+  def test_binary_to_multiple_decimal
+    skip
+    digits = [1, 0, 1, 0, 1, 0]
+    input_base = 2
+    output_base = 10
+    expected = [4, 2]
+
+    converted = BaseConverter.convert(input_base, digits, output_base)
+
+    assert_equal expected, converted,
+                 "Input base: #{input_base}, output base #{output_base}. " \
+                 "Expected #{expected} but got #{converted}."
+  end
+
+  def test_decimal_to_binary
+    skip
+    digits = [4, 2]
+    input_base = 10
+    output_base = 2
+    expected = [1, 0, 1, 0, 1, 0]
+
+    converted = BaseConverter.convert(input_base, digits, output_base)
+
+    assert_equal expected, converted,
+                 "Input base: #{input_base}, output base #{output_base}. " \
+                 "Expected #{expected} but got #{converted}."
+  end
+
+  def test_trinary_to_hexadecimal
+    skip
+    digits = [1, 1, 2, 0]
+    input_base = 3
+    output_base = 16
+    expected = [2, 10]
+
+    converted = BaseConverter.convert(input_base, digits, output_base)
+
+    assert_equal expected, converted,
+                 "Input base: #{input_base}, output base #{output_base}. " \
+                 "Expected #{expected} but got #{converted}."
+  end
+
+  def test_hexadecimal_to_trinary
+    skip
+    digits = [2, 10]
+    input_base = 16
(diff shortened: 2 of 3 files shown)


Insti:
How is it going?

sebarmano:
Good! I'll commit my first working solution soon, so we can discuss about it. I'm sure it can be improved.

Insti:
Great. I look forward to seeing it. 
It's OK to post a work-in-progress too.

config.json
```diff
@@ -66,6 +66,7 @@
     "wordy",
     "allergies",
     "poker",
+    "all-your-base",
```
Insti: Put new problems last in config list, once they've been solved a few times we can move them to a more appropriate place in the list.
(We probably should document this somewhere.)

sebarmano:
I think it's ready for review. :)

Insti:
Looking good. Did you manually create the test file or did you write a generator to generate it from the x-common all-your-base.json file?

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -0,0 +1,114 @@
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'all_your_base'
+
+class AllYourBaseTest < Minitest::Test
+  def test_single_bit_to_one_decimal
+    assert_equal [1], BaseConverter.new.convert(2, [1], 10)
```
Insti: It would be good if this was a class method rather than an instance method.
`BaseConverter.convert(2, [1], 10)`

config.json
sebarmano: Sure. I think the confusion comes from this post https://www.sitepoint.com/exorcise-your-newbie-demons-by-contributing-to-exercism/. In the "Updating the configuration" section it asks to "stick it anywhere in the second half". I think that document this would be great.

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  def test_binary_to_single_decimal
+    skip
+    assert_equal [5], BaseConverter.new.convert(2, [1, 0, 1], 10)
```
Insti: It would be nice to know what the numbers were.
I'd create 4 named local variables: `expected`, `input_base`, `digits`, `output_base`
So all your assertions become:
assert_equal expected, BaseConverter.convert( input_base, digits, output_base )

sebarmano:
I manually created the test file from the json file. I also (forgot to say this) decided on the special cases as I thought made more sense. Basically I took the approach of "what would happen if writing this in a calculator". I that way, for example, entering [0, 0] is valid and the same as [0] (and, in turn, 0 in any base is 0). Hope this makes sense...

exercises/all-your-base/all_your_base_test.rb
Insti: I'd also create a helper method or refinement to string so I could do something like:
`digits = "101".to_digits`
(Maybe that's not an improvement..)

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  def test_single_decimal_to_binary
+    skip
+    assert_equal [1, 0, 1], BaseConverter.new.convert(10, [5], 2)
```
Insti: I'd also like to see useful error messages, this will show me something like 
```
Expected: [1,0,1] 
Actual: [5]
```
Which doesn't help me work out what the problem is.
adding a 3rd `message` argument to `assert_equal` lets you give that information as part of the failure message
for example:
`input base: 10, digits: [5], output_base: 2. Expected [1,0,1] but got [5]`

exercises/all-your-base/example.rb
```diff
@@ -0,0 +1,32 @@
...
+class BaseConverter
+  def convert(base_from, number_array, base_to)
+    return if number_array.any?{|number| number < 0 || number >= base_from}
+    return if base_from <= 1 || base_to <= 1
+    return [] unless number_array.any?
+    number_in_base_ten = convert_to_base_ten(number_array, base_from)
+    convert_from_base_10(number_in_base_ten, base_to)
+  end
+
+  def convert_to_base_ten(number_array, base)
```
Insti: I think you forgot to remove `_ten` here.

exercises/all-your-base/example.rb
```diff
@@ -0,0 +1,32 @@
+module AllYourBase
```
Insti: This should be called `BookKeeping`

exercises/all-your-base/example.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def convert_to_base_ten(number_array, base)
+    total = 0
+    number_array.reverse.each_with_index do |number, index|
+      total += number * base ** index
+    end
+    total
+  end
+
+  def convert_from_base_10(number, base_to)
```
Insti: Remove `_10` here.

Insti:
Looks good, :heart: It's certainly in a good enough state to merge in as it is, but it would be even better if you're able to address some of the points I've raised. (And also feel free to disagree if you don't think they'd be an improvement.)

I've not looked too closely at your example solution, that's what the Exercism site is for :).

config.json
Insti: Ok, Thanks. That document is out of date already. (Which is not at all your fault! Thanks for following instructions.)

exercises/all-your-base/all_your_base_test.rb
sebarmano: Do you mean that it should be another method in the BaseConverter class? Would it return the array of values as an input to the converter?

Insti: It would be a method in the `AllYourBaseTest` class that would return an array of values as an input to the converter. 
On reflection **I don't think it's a good idea here**, since you're already getting the input arrays from the json file.
I'm just lazy and would rather type "101" than "[1, 0, 1]"

Insti:
> I also decided on the special cases as I thought made more sense

You should create a pull request to add them to the `all-your-base.json` file in https://github.com/exercism/x-common

Sometime in the future, someone will create a generator to generate the test file based on the common json file, and this might cause your cases to be lost.

exercises/all-your-base/example.rb
sebarmano: Actually, I'm going from base "a" to base "b" through base 10, that's why the names of the methods are those ones. I'm sure there is a better solution without doing this. I'll see if I can find it and change the example.

sebarmano:
I think it's ready for another review. If it's good I'll rebase and squash to leave just one commit. Sorry for all the time it took me to finally finish this!

Insti:
Looks good to me :+1:

sebarmano:
I think it's ready to merge. I've rebased and squashed all commits into one. Let me know if I am missing something. :)

Insti:
Looks good :+1: I'll wait and see if @kotp has any comments, otherwise I'll merge it in over the weekend.

kotp:
Nothing so far, but I think I can review this later this evening.

Insti:
Ok. If you're happy you can merge it in.

Insti:
Thanks for all your work on this @sebarmano :heart:

PR #394, opened 2016-07-14
Add test generator for "binary" exercise

tommyschaefer:
I recently came across #381, and saw that @kytrinyx suggesting converting all existing exercises to use generated tests as a first step.

I was wondering if this is still a goal? If so, I would love to help work on that! 😄

To start this off, I've added a generator for the "binary" example.

Thank you so much for your time, and I look forward to your feedback!

PR #405, opened 2016-07-28
Re-generate binary tests with split assertions

### LINKED ISSUE #404: binary: Re-generate test file.
The binary-generator will need to be re-run once https://github.com/exercism/x-common/issues/303 has been completed.

(Synchronisation is hard.)

tommyschaefer:
Resolves #404 

With changes proposed at exercism/x-common#303 and based on test data proposed at exercism/x-common#305.

This PR includes two changes:
* Re-generate tests from exercism/x-common
* Remove compound assertion logic from BinaryCases since all tests with multiple assertions have been split into separate tests.

Thank you so much for your time and feedback!

### DIFF
--- a/exercises/binary/binary_test.rb
+++ b/exercises/binary/binary_test.rb
@@ -5,7 +5,7 @@ require 'minitest/autorun'
 require_relative 'binary'
 
 # Test data version:
-# dd43e66
+# 01092b0
 class BinaryTest < Minitest::Test
   def test_binary_0_is_decimal_0
     # skip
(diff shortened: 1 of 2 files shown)


tommyschaefer:
Hmm... Looks like the build is failing because one of the test names causes a line in the test file to exceed 80 cols. I'll see if I can come up with a shorter name and regenerate! 😄

Insti:
Does the version number need to increment?

tommyschaefer:
I think based on @kytrinyx's comment on #394:

>If solutions that pass the ungenerated test file would also pass the generated one, I wouldn't change the version.

Probably not?

The only change made to the original assertions is that we now test that `01201` is invalid instead of `012`. Do you think that warrants a version bump?

Insti:
Seems fine without a version bump then. Good job.

Insti:
Waiting on: https://github.com/exercism/x-common/pull/305

PR #411, opened 2016-08-01
[WIP] nth-prime: Add test file generator.

cacqw7:
The Nth Prime problem was missing `BookKeeping` module.


kotp:
It is OK, because this exercise is not yet generated, and so does not need the BookKeeping information yet.

Would you be up to writing the generator while you are here and interested?

cacqw7:
Sure 👍

cacqw7:
@kotp Just pushed up code for the generator. 

I also [created a PR on x-common](https://github.com/exercism/x-common/pull/332) with the test case `json`.

Let me know if I've missed something!

exercises/nth-prime/nth_prime_test.rb
```diff
@@ -31,4 +31,26 @@ def test_weird_case
...
+  # module BookKeeping
+  #   VERSION = 1 # Where the version number matches the one in the test.
+  # end
+  #
+  # If you are curious, read more about constants on RubyDoc:
+  # http://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/constants.html
+
+  def test_bookkeeping
+    skip
+    assert_equal 2, BookKeeping::VERSION
```
kotp: This should be version 1.

exercises/nth-prime/example.rb
```diff
@@ -1,3 +1,7 @@
+module BookKeeping
+  VERSION = 2
```
kotp: Should be version 1.  (The ungenerated one is version 0).

kotp:
First look, this looks pretty good.  Version number needs to be set back to 1.

lib/nth_prime_cases.rb
```diff
@@ -0,0 +1,19 @@
+class NthPrimeCase < OpenStruct
+  def name
+    'test_%s' % description.downcase.gsub(/[ -]/, '_')
+  end
+
+  def do
```
Insti: I really don't like that there is a method called `do`, 
I see it a lot, is there documentation that uses an example with `do` in it? 

I'd prefer it be called `actual` to fit the `assert_equal expected, actual` pattern but am open to other suggestions for better names.

kotp: This is true, it mirrors a keyword in Ruby, so is a good feeling.

Let's avoid it if we can.

cacqw7: There's no documentation on `do`. I had seen it in a few other files when I was working on this.

I'll change it to `actual`.

cacqw7:
I rolled back the version and updated it to reflect the changes to the [x-common PR](https://github.com/exercism/x-common/pull/332)

exercises/nth-prime/example.tt
```diff
@@ -0,0 +1,22 @@
...
+# Test data version:
+# <%= sha1 %>
+# Rubocop directives
+# rubocop:disable Style/NumericLiterals
+#
+class NthPrimeTest < Minitest::Test<% test_cases.each do |test_case| %>
+  def <%= test_case.name %><% if test_case.skipped? %>
+    skip<% end %>
+    assert_equal <%= test_case.expected %>, <%= test_case.actual %>
+  end
```
Insti: It would be good to remove the conditional from the template.
See how https://github.com/exercism/xruby/blob/master/lib/rna_transcription_cases.rb handles this.

Insti:
Waiting on https://github.com/exercism/x-common/pull/332 which needs to be merged first.

exercises/nth-prime/nth_prime_test.rb
```diff
@@ -22,13 +25,35 @@ def test_sixth_prime
 
   def test_big_prime
     skip
-    assert_equal 104_743, Prime.nth(10_001)
+    assert_equal 104743, Prime.nth(10001)
```
Insti: If you wanted to be nice, you could make a helper method that formatted the numbers from the json into underscored versions in the test file.

(This is not at all a requirement, just an optional extra if you feel like it.)

exercises/nth-prime/nth_prime_test.rb
```diff
@@ -22,13 +25,35 @@ def test_sixth_prime
...
   end
 
-  def test_weird_case
+  def test_there_is_no_zeroth_prime
     skip
     assert_raises ArgumentError do
       Prime.nth(0)
     end
   end
```
Insti: Where is this code generated?

cacqw7: The `json` file generates the `false` case. After it generated that I changed it to raise the exception.

I wasn't sure whether to leave it `false` or change it to this.

Insti: I think raising an ArgumentError is ok.

The important thing is that the generator generates a test file that doesn't need further modification, so if/when the .json file changes in future, we can just re-generate the test file and not worry about any special cases lurking in there that need manual tweaking.

cacqw7: I updated [example.tt](https://github.com/cacqw7/xruby/blob/9b1b74bf000466805dd6d2559fd928802a084a9b/exercises/nth-prime/example.tt#L13) to generate the exception test.

Cohen-Carlisle:
This looks ready to merge to me. 👍

Insti:
Still waiting on exercism/x-common#332. Perhaps you could prod that along.

Cohen-Carlisle:
The x-common PR has been merged. Is this PR in sync with the master branch of x-common and ready to be merged?

cacqw7:
@Cohen-Carlisle Sorry for the late reply. I think this PR is good to go with the recent update to the `data` method in [`generator.rb`](https://github.com/exercism/xruby/blob/cc9d7e88a7d5b35e15413f34acc2f406e72b354a/lib/generator.rb#L21)

Insti:
Thanks a lot @cacqw7 :heart:

PR #350, opened 2016-05-17
Add diamond

cloudleo:
Working on Diamond

PR #413, opened 2016-08-02
adding Diamond solution and minitest #350

cloudleo:


Insti:
you seem to have submitted diamond_mess again.
You need to try the "clean" one

PR #416, opened 2016-08-12
Add a more helpful test failure message

srabuini:
As you can see in the README file, you must ignore arithmetical
precedence rules and calculate as you move forward each step.

### DIFF
--- a/exercises/wordy/wordy_test.rb
+++ b/exercises/wordy/wordy_test.rb
@@ -71,7 +71,8 @@ class WordProblemTest < Minitest::Test
   def test_add_then_multiply
     skip
     question = 'What is -3 plus 7 multiplied by -2?'
-    assert_equal(-8, WordProblem.new(question).answer)
+    message = 'You should ignore order of precedence. -3 + 7 * -2 = -8, not -17'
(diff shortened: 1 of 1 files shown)


exercises/wordy/wordy_test.rb
```diff
@@ -71,7 +71,8 @@ def test_multiply_twice
   def test_add_then_multiply
     skip
     question = 'What is -3 plus 7 multiplied by -2?'
-    assert_equal(-8, WordProblem.new(question).answer)
+    message = 'You should ignore order of precedence. 3 + 2 * 3 = 15, not 9.'
```
Insti: It would be nice if the error message matched the question: "-3 + 7 * -2 = -8, not -17"

Insti:
Awesome, thanks for adding this :heart:

exercises/wordy/wordy_test.rb
srabuini: You're right. I thought that a different example could add value. But it might be confusing. I'll change it.

Insti:
Great. looks good. :+1:

PR #417, opened 2016-08-12
Fix typo in SETUP.md

awolfson:
Add space to command between flag and argument so it is clearer what the command does.


kytrinyx:
Thanks! I think this makes sense. (Leaving the final choice to the track maintainers).

Insti:
More clarity is always good :+1:

Insti:
Thanks @awolfson.

PR #418, opened 2016-08-17
Possible changes in scale-generator

srabuini:
Maybe there's a reason to write all flats with `%` and all sharps with `[]`. Otherwise, I think would be better to unify criteria. Idem about scale names.

Please have a look and let me know.

Thanks!.


Insti:
Looks good, :+1: for consistency.
String to Symbol is a breaking change so we should add a BookKeeping::VERSION = 1

Insti:
Looks good. :+1:
@kotp?

kotp:
This does look good.

Insti:
Thanks @srabuini :heart:

PR #420, opened 2016-08-21
Fix tests and example [list-ops]

srabuini:
Hey,

I've found a couple of inconsistencies here. Some `assert` should be `assert_equal`, and fixing that, I've realized that some tests doesn't pass anymore. So I'm proposing those changes.

Please have a look and let me know.

Thanks!

### DIFF
--- a/exercises/list-ops/example.rb
+++ b/exercises/list-ops/example.rb
@@ -1,5 +1,4 @@
 class ListOps
-  VERSION = 1
   ## Do not use the functions map, concat, count, length, etc.
   ## and other built in ruby methods to solve  these
   ##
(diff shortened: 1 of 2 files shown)


Cohen-Carlisle:
That stuff was definitely pretty broken 😨
Thanks, @srabuini, these changes look good to me.

Insti:
Who merged this? Oh, it was me :cry: 
Why did I do that? I was cleaning up abandoned pull requests, and this one looked like it had enough work on it that just rejecting it would be a waste.

[I wrote](https://github.com/exercism/xruby/pull/332#issuecomment-234627267):
> Does anyone want to address the issues raised by @bernardoamc's last review?
> If not I suggest we should just merge it and have a less-than-perfect problem out there for people to try and then suggest improvements to.

Wow, this turned out to be more broken than I expected, but at least my plan worked.  Many thanks @srabuini

Insti:
Merging this now due to the brokenness of the original version.
Any tweaks needed can be made in another PR.

PR #421, opened 2016-08-27
Fix class name on Ruby's Pangram exercise

yogodoshi:
:)

### DIFF
--- a/exercises/pangram/pangram_test.rb
+++ b/exercises/pangram/pangram_test.rb
@@ -79,6 +79,6 @@ class PangramTest < Minitest::Test
 
   def test_bookkeeping
     skip
-    assert_equal 2, BookKeeping::VERSION
+    assert_equal 2, Pangram::VERSION
   end
 end


Cohen-Carlisle:
Hi, @yogodoshi! Thanks for your interest in exercism and ruby! The change to `BookKeeping` is actually intentional; we didn't want to force users to pollute the solution class or module with test artifacts. The intention is also reflected in the comments above that test:
```
  # Problems in exercism evolve over time, as we find better ways to ask
  # questions.
  # The version number refers to the version of the problem you solved,
  # not your solution.
  #
  # Define a constant named VERSION inside of the top level BookKeeping
  # module.
  #  In your file, it will look like this:
  #
  # module BookKeeping
  #   VERSION = 1 # Where the version number matches the one in the test.
  # end
  #
  # If you are curious, read more about constants on RubyDoc:
```
I'm going to close this PR in light of the above, but again, thanks for looking to contribute!

PR #423, opened 2016-08-28
[WIP] add change (exercise)

laushinka:


Insti:
@laushinka Thanks for deciding to help us add exercises to Exercism.
If you're not, use the 'close request' button on Github so someone else can work on adding the problem.  If we don't hear back from you, we will close it ourselves one week from now. 
(On or after the 24th October 2016)

Sorry for not giving you feedback sooner, but if you are still interested in working on it there are a few things that need fixing before we can merge it in:

- [ ] `change.rb` needs to be renamed to `example.rb`
- [ ] a "change" section needs to be added to the end of the `problems` array in `config.json`
- [ ] `assert_equal` arguments should be ordered as 'expected', 'actual'
- [ ] version should be 1

PR #437, opened 2016-09-18
Triangle: Update tests to match canonical-data.json.

durrellchamorro:
As discussed at exercism/x-common#311 and exercism/x-common#375 here are my proposed updates to the tests which reflect the canonical-data.json.

### DIFF
--- a/exercises/triangle/triangle_test.rb
+++ b/exercises/triangle/triangle_test.rb
@@ -33,9 +33,19 @@ class TriangleTest < Minitest::Test
     assert_equal :isosceles, Triangle.new(10, 10, 2).kind
   end
 
+  def test_isosceles_triangles_have_unequal_side_larger_than_equal_sides
+    skip
+    assert_equal :isosceles, Triangle.new(4, 7, 4).kind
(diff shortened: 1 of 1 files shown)


kotp:
@durrellchamorro are you interested in doing the generator for the test so that it can be done once, and regenerated if that json file changes?

durrellchamorro:
I just looked at generator.rb and some of those cases.rb files. It's really neat how you generate the tests this way. I didn't realize that's how it is done and I'm not exactly sure how it all works, so I'll pass on writing the generator for this test. If you don't want to merge this PR, I understand.

kotp:
It is definitely not that I don't want to merge this in, I just know that it will come back because it is not using that data directly, so this will (eventually) need to be revisited.  With that said, the work is done in the form already, and so... it should come in.

:eyes:  @exercism/ruby

kotp:
If by no one objects, I will consider no objections to merging.

Insti:
Correct manually created tests are better than incorrect tests. Merge :+1:

PR #440, opened 2016-10-02
attribute the vienna.rb icon Floor Dress(?) let us use

catb0t:
a step in fixing https://github.com/exercism/exercism.io/issues/3112.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,3 +164,6 @@ work, please see the [contributing guide](https://github.com/exercism/x-api/blob
 The MIT License (MIT)
 
 Copyright (c) 2014 Katrina Owen, _@kytrinyx.com
+
+## Ruby icon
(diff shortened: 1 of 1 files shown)


Insti:
Thanks for this @catb0t.

Providing more context in the PR message would be helpful, it took me a long time to work out what you were doing and where the line you added to the readme came from.

Something like: 

Moving the image credit from [the exercism.io ATTRIBUTION.md file](https://github.com/exercism/exercism.io/blob/master/ATTRIBUTION.md) into the track Readme.md since the images now live in the individual track repositories.

catb0t:
Oh, sorry about that, the description for the pulls I created this morning are abnormally short because I was expecting @kytrinyx to merge most of them (which she did).

kytrinyx:
Sorry, both of you! We've been on a shared rampage, @catb0t and I.

This looks good to me.

Issue #412, opened 2016-08-01
Raindrop: Common data has changed.


kotp:
exercism/x-common#314 has updated the common data.

Generator needs to be adjusted.

kotp:
Thank you @Insti

Insti:
@bmulvihill did most of the work.

PR #441, opened 2016-10-04
Update raindrops generator

bmulvihill:
Based on https://github.com/exercism/xruby/issues/412

The generator no longer relies on the description, instead using the number

The generator metadata directory also needed to be updated based on the x-common layout.


Insti:
Awesome, Thanks :heart:
Someone will review it in the next few days.

lib/raindrop_cases.rb
```diff
@@ -1,6 +1,6 @@
 class RaindropsCase < OpenStruct
   def name
-    'test_%s' % description.downcase.gsub(/[ -]/, '_')
+    'test_%s' % number
```
Insti: Where does `number` come from?

lib/generator.rb
```diff
@@ -14,11 +14,11 @@ def initialize(name, cases)
   end
 
   def metadata_dir
-    File.expand_path(File.join('..', '..', '..', METADATA_REPOSITORY), __FILE__)
+    File.expand_path(File.join('..', '..', '..', METADATA_REPOSITORY, 'exercises', "#{ name }"), __FILE__)
```
Insti: No need to have redundant string interpolation here, 
```
File.expand_path(File.join('..', '..', '..', METADATA_REPOSITORY, 'exercises', name), __FILE__)
```
should work just as well.

Insti:
Ideally the generator.rb changes would be part of separate PR, since they're not raindrops specific. If you can make another PR for that it would be good.

lib/raindrop_cases.rb
Insti: Found it, It's one of the keys in the JSON.

Insti:
All the raindrops specific stuff looks good :+1: 

Just move the `generator.rb` changes to a new PR and fix that string interpolation and we're good to go.

(Please let me know if you need help with this.)

bmulvihill:
@Insti Ok, I will open another PR with the generator stuff

Insti:
The updated raindrops files should be part of this PR.

Create the generator PR and we can merge that in, then add the regenerated raindrops_test file back into this PR.

(Or you can do it locally by just not checking in the changed generator.rb file to this branch.)

Does that make sense?

Insti:
Waiting on the generator update PR to be merged.

Insti:
Thanks again @bmulvihill !
If you're doing Hacktoberfest you're half way to a T-shirt.

Issue #439, opened 2016-10-02
Queen-Attack missing test cases


remcopeereboom:
This might be a duplicate of #396, but queen attack is missing test cases for testing that queens are actually placed on the board. These test cases _are_ in the [x-common tests](https://github.com/exercism/x-common/blob/master/exercises/queen-attack/canonical-data.json).

Insti:
Thanks. We should still have correct test cases even if they are manually generated.

What needs to happen:

Either 
a) Write a generator 
OR
b) Manually fix the [queen_attack_test.rb](https://github.com/exercism/xruby/blob/master/exercises/queen-attack/queen_attack_test.rb) file.

remcopeereboom:
I'll see if I can submit a fix to the .rb file later today.

I'm not familiar with the protocol for the generators. I'll have a look at the alphametrics one to figure out what the desired design is. I could probably write one for `queen_attack`.

Insti:
Manually adding the missing cases is fine.

PR #444, opened 2016-10-04
Queen-Attack missing test cases #439

remcopeereboom:
Removed test for default positions. It seems arbitrary for the given
problem, the common tests don't include it as a requirement and it seems
best to leave this as a design decision.

Added tests to ensure that queens are placed on the board.

Refactored test to use an inline block, same as the other tests.

Added a version constant test.

### DIFF
--- a/exercises/queen-attack/queen_attack_test.rb
+++ b/exercises/queen-attack/queen_attack_test.rb
@@ -4,12 +4,6 @@ require 'minitest/autorun'
 require_relative 'queen_attack'
 
 class QueensTest < Minitest::Test
-  def test_default_positions
-    queens = Queens.new
-    assert_equal [0, 3], queens.white
-    assert_equal [7, 3], queens.black
-  end
-
   def test_specific_placement
     skip
     queens = Queens.new(white: [3, 7], black: [6, 1])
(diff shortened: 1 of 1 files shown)


kotp:
Should this also be coordinated with a change for [queen-attack/canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/queen-attack/canonical-data.json) as a pull request?  I you opened up exercism/x-common#396 and if you are working on the generator, I would say proceed with that, and work on this proposal via a PR in the `.json` file.

That way, it isn't a fixed locally, then generated with the current, and then fixed with the data.

remcopeereboom:
> Of course, if you aren't knee deep in the generator, then fixing it and having it sooner is better than later.

I was working on the generator code, but I had some deeper issues with it that I can't quite recall right now. I think I was unhappy with the common json test data as it was currently and I didn't want to change it because I'm not sure how other languages use it. And I don't want to implement a generator for the current json data if it's going to be changed. Either way, I felt it best to patch up the tests in the meantime.

Note that it says travis failing, but I'm not getting any details as to why. Any ideas?

remcopeereboom:
Is the build failing for a test implementation of queen attack? I couldn't find one in the repository. Did I miss it?

remcopeereboom:
Okay, see it now: `example.rb`. I was looking for `queen_attack.rb`. Will update and resubmit.

Closing this, we can move the discussion to #439 or to the new submission.

Issue #443, opened 2016-10-04
Update Rubocop to match project philisophy.


Insti:
Following on from the discussion on an exercism.io PR: https://github.com/exercism/exercism.io/pull/3114

The current xruby Rubocop configuration is too strict.
Which Rubocop options are important here?

kotp:
Investigating..

kotp:
It boils down to currently just these offenses, I think.  

17  Style/FormatString
16  Metrics/MethodLength
6   Metrics/LineLength
3   Style/TrailingCommaInArguments
2   Style/SpaceAroundEqualsInParameterDefault
1   Style/RaiseArgs

This does not state which way the style should be, such as the trailing commas, nor the limit for Line Length, or Method Length, just that they are the ones coming up.

Understand that in my interpretation I am ignoring the `example.rb` files, which are, as of yet, still "Proof of concept" and not meant to be exemplar programming examples...

kotp:
Space around equals in parameter default, line length, and method length, those are the only ones listed I really have more than passing opinions on, for varying reasons.

kotp:
After doing some thinking of how to get the rules out there, so that they can be discussed, I realized that I can list and disable all of them.  See the PR #449 referenced above...

kotp:
It doesn't necessarily complete this issue, as this is a long and ongoing issue, I suspect.  But... merged in the "no-op" version of the rubocops.

PR #447, opened 2016-10-09
Fix Rubocop code offenses

pvcarrera:


kotp:
Do you agree with these changes, @pvcarrera ?  I am not sure I do.

The only one that I agree with is the one in anagram.rb.

exercises/anagram/anagram_test.rb
```diff
@@ -59,7 +59,7 @@ def test_multiple_anagrams
     skip
     detector = Anagram.new('allergy')
     anagrams =
-      detector.match %w( gallery ballerina regally clergy largely leading)
+      detector.match %w(gallery ballerina regally clergy largely leading)
```
kotp: This change is good.

Insti:
@pvcarrera What Rubocop settings are you using? None of these things generate rubocop warnings on my system.

exercises/bowling/bowling_test.rb
```diff
@@ -139,18 +139,20 @@ def test_should_not_allow_rolls_with_negative_pins
     skip
     assert_raises(
       RuntimeError,
-      'Pins must have a value from 0 to 10') do
-        @game.roll(-1)
-      end
+      'Pins must have a value from 0 to 10',
+    ) do
```
Insti: I'd prefer to see the assert_raises made into one line rather than 4.

pvcarrera: Yes, I agree

pvcarrera:
@Insti You are right, I was running a different version run version: `0.42.0`. When I run the project version `bundle exec rubocop` I only get:
`lib/alphametics_cases.rb:1:1: C: Missing frozen string literal comment.`
`lib/generator.rb:1:1: C: Missing frozen string literal comment.`

Should we close this PR then? I can open a new one with only the changes related to this version.

kotp:
I am in the process of disabling all cops until they can be discussed.  The one change is good, but very minor.  Please look at #449 and #443 for further discussion as to what is happening.  Looking forward to your input as well.

kotp:
As to opening a new PR, no need to do that, just force push the changes that you would really like to see happen, @pvcarrera then the reasoning and discussion stays intact.

pvcarrera:
Sorry at the end this PR only removes a withe space 😄

kotp:
Definitely worthwhile!

Issue #396, opened 2016-07-15
Exercise tests should be generated


tommyschaefer:
I recently came across #381, and saw that @kytrinyx suggesting converting all existing exercises to use generated tests as a first step.

On PR #394, @Insti recommended creating an issue to track progress towards this goal.

Here's a list of exercises and whether they generate their tests:

## Converted

- [x] acronym
- [x] all-your-base
  - Generator at #510
- [x] alphametics
- [x] anagram
- [x] beer song
- [x] binary
- [x] bowling
- [x] bracket-push
- [x] clock
- [x] connect
- [x] custom-set
- [x] difference-of-squares
- [x] dominoes
- [x] etl
- [x] gigasecond
- [x] grains
  - Generator at #514 
- [x] hamming
- [x] hello-world
- [x] isogram
- [x] largest-series-product
- [x] leap
- [x] luhn
- [x] nth-prime
- [x] ocr-numbers
- [x] pangram
- [x] pig-latin
- [x] queen-attack
- [x] raindrops
- [x] rna-transcription
- [x] roman-numerals
- [x] run-length-encoding
- [x] say
- [x] sieve
- [x] space-age (#583)
- [x] tournament
- [x] transpose
- [x] triangle
- [x] two-bucket
- [x] word-count
- [x] wordy
  - Generator at #509

## Has Canonical Data in x-common
- [ ] allergies
- [ ] atbash-cipher
- [ ] binary-search
- [ ] bob
- [ ] change
- [ ] crypto-square
- [ ] flatten-array
- [ ] food-chain
- [ ] forth
- [ ] grep
- [ ] markdown
- [ ] meetup
- [ ] minesweeper
- [ ] pascals-triangle
- [ ] pov
- [ ] rail-fence-cipher
- [ ] react
- [ ] rectangles
- [ ] robot-simulator
- [ ] scrabble-score
- [ ] sublist
- [ ] sum-of-multiples

## Needs Canonical Data

- [ ] accumulate
- [ ] binary-search-tree
- [ ] circular-buffer
- [ ] diamond
- [ ] grade-school
- [ ] hexadecimal
- [ ] house
- [ ] kindergarten-garden
- [ ] linked-list
- [ ] list-ops
- [ ] matrix
- [ ] nucleotide-count
- [ ] octal
- [ ] palindrome-products
- [ ] perfect-numbers
- [ ] phone-number
- [ ] point-mutations
- [ ] poker
- [ ] prime-factors
- [ ] protein-translation
[...]

PR #448, opened 2016-10-09
Add word count generator

pvcarrera:
In the list of exercises that need a generator #396 , I saw word-count. Once I started working on it, I found the template `exercises/word-count/example.tt ` but I couldn't find any generator associated to it, so I decided to modify the current template and create the generator that you can see in this PR.
 
The `canonical-data.json` file from x-common has less examples that the current version of `word_count_test.rb` Should I add these examples to x-common?

### DIFF
--- a/lib/word_count_cases.rb
+++ b/lib/word_count_cases.rb
@@ -4,7 +4,7 @@ class WordCountCase < OpenStruct
   end
 
   def object_under_test
-    "Phrase.new('#{input}')"
+    %Q(Phrase.new("#{input}"))
   end
 
   def skipped?
--- a/lib/word_count_cases.rb
+++ b/lib/word_count_cases.rb
@@ -4,7 +4,7 @@ class WordCountCase < OpenStruct
   end
 
   def object_under_test
-    %Q(Phrase.new(#{input.to_json}))
+    %Q(Phrase.new(#{input.inspect}))
   end
 
   def skipped?
--- a/lib/word_count_cases.rb
+++ b/lib/word_count_cases.rb
@@ -1,0 +1,19 @@
+class WordCountCase < OpenStruct
+  def name
+    'test_%s' % description.tr(' ', '_')
+  end
+
+  def object_under_test
+    "Phrase.new('#{input}')"
+  end
+
+  def skipped?
+    index.nonzero?
+  end
+end
+
+WordCountCases = proc do |data|
+  JSON.parse(data)['cases'].map.with_index do |row, i|
+    WordCountCase.new(row.merge('index' => i))
+  end
+end
--- a/lib/word_count_cases.rb
+++ b/lib/word_count_cases.rb
@@ -4,7 +4,7 @@ class WordCountCase < OpenStruct
   end
 
   def object_under_test
-    %Q(Phrase.new("#{input}"))
+    %Q(Phrase.new(#{input.to_json}))
   end
 
   def skipped?
--- a/exercises/word-count/example.tt
+++ b/exercises/word-count/example.tt
@@ -7,9 +7,10 @@ require_relative 'word_count'
 # <%= sha1 %>
 class PhraseTest < Minitest::Test<% test_cases.each do |test_case| %>
   def <%= test_case.name %><% if test_case.skipped? %>
-    skip<% end %><% if test_case.raises_error? %>
-    assert_raises(ArgumentError) { <%= test_case.do %> }<% else %>
-    assert_equal <%= test_case.expected %>, <%= test_case.do %><% end %>
+    skip<% end %>
+    phrase = <%= test_case.object_under_test %>
+    counts = <%= test_case.expected %>
+    assert_equal counts, phrase.word_count
   end
 <% end %>
 <%= IO.read(XRUBY_LIB + '/bookkeeping.md') %>
--- a/exercises/word-count/word_count_test.rb
+++ b/exercises/word-count/word_count_test.rb
@@ -35,9 +35,7 @@ class PhraseTest < Minitest::Test
 
   def test_handles_expanded_lists
     skip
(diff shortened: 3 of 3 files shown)


kotp:
If the prior non-generated version of the test had more tests than what is in x-common, then we need to review the discussions that lead to that difference.

Do you feel like the additional tests are something that is Ruby specific?  Do you feel that the additional tests may be redundant or a good addition to other languages?

kotp:
BTW: Thanks for jumping in here.

pvcarrera:
The additional cases are: 
```
test_count_everything_just_once
test_handles_cramped_lists
test_with_apostrophes
test_with_quotations
```
I think this are valid test that cover cases that are not covered in the x-common cases

Insti:
@pvcarrera create a PR in [exercism/x-common](https://github.com/exercism/x-common) adding the additional cases to the canonical-data.json file.

lib/word_count_cases.rb
```diff
@@ -0,0 +1,19 @@
...
+  def name
+    'test_%s' % description.tr(' ', '_')
+  end
+
+  def object_under_test
+    "Phrase.new('#{input}')"
+  end
+
+  def skipped?
+    index > 0
```
kotp: Or even `index.notzero?`   No action needed, just an annotation.

pvcarrera:
I think I need to update this branch/PR: Merge #451, then remove the generate-word-count scripts. Also, the generated tests don't follow the project style rules. I ran `rubocop -a` after the test cases where generated but, after the changes I did in #454, I realised that the test must be generated with the correct format.

kotp:
Don't worry about updating the branch against 451, you should be able to rebase against master.

There is another change (unsure if it is in yet, though) that sets style enforcement/checks back to 0.  But this would also be on master.

The process for doing this would be on your fork of the project...

(Not tested, just written out here as a guide)

```bash
git checkout master
git pull origin master
git checkout add-word-count-generator
git rebase master
```

That should move your branch to the latest master branch.

pvcarrera:
Done, I rebased, squashed and forced push. 

I'm not sure if this is open to discussion, but this is my opinion about PRs workflow:

Rebasing an open PR rewrites the PR history, this means that comments are not synchronized with changes anymore. I don't usually rebase/force push once a PR is open, on the other hand, I add commits and merge master in the branch as many times as necessary. Once the PR is ready, I use the GitHub Squash and merge option. This reduces the PR to one commit but still keeps a link to the closed PR request where you can find the original code with its comments.

kotp:
I was only stating about rebasing from master, since I changed how the Rubocop is behaving.

Thanks for the notes on the PR workflow.  It is very similar to how things are normally done here, though we reduce to "a few" if appropriate rather than always a single commit.  

I did not realize we were ready for that, but you doing so, definitely signals that we are ready for "final approach" on this.

kotp:
#451 Common generation script is now in master.

pvcarrera:
Cool, this PR is already up to date with #451

kotp:
OK, I attempted to do a generate, might me doing something wrong since it changed recently.
Is that working for you?

kotp:
I am using `ASSIGNMENT=word-count make test-assignment` to test it locally, it is coughing due to the line 68 in the tests.

```text
running tests for: word-count
/tmp/word-count.CuENNcRPR7/word_count_test.rb:68: syntax error, unexpected tIDENTIFIER, expecting ')'
    phrase = Phrase.new('First: don't laugh. Then: don't cry.')
```

bmulvihill:
Looks like its due a to a use of single quotes as you mentioned in https://github.com/exercism/xruby/pull/463.  I pulled locally and changed the object_under test method as so 
```ruby
  def object_under_test
    "Phrase.new(\"#{input}\")"
  end
```

Output:
screen shot 2016-10-18 at 3 35 02 pm

Insti:
> `"Phrase.new(\"#{input}\")"`

This will still break if the string contains double quotes. 
Much better to find a solution that correctly escapes quote characters.

pvcarrera:
@Insti The input string comes from a json file, example: `"input": "Joe can't tell between 'large' and large."` I guess if one of the strings contains a double quote it'll come scaped, something like: `"input": "Joe can't tell between \"large\" and large."`

Insti:
It's still a problem because the json is parsed before we try to insert it.

Example:
```ruby
require 'json'
json = { :example => 'Contains " double quote' }.to_json
puts json
input = JSON.parse(json)['example']
puts %Q(Quote not escaped "#{input}")
```
Output:
```
{"example":"Contains \" double quote"}
Quote not escaped "Contains " double quote"
```

This might work though:
```
%Q(Quote now escaped #{input.to_json})
```
(Note that the surrounding quotes are now provided by the .to_json method.)

lib/word_count_cases.rb
```diff
@@ -4,7 +4,7 @@ def name
   end
 
   def object_under_test
-    %Q(Phrase.new("#{input}"))
+    %Q(Phrase.new(#{input.to_json}))
```
Insti: Sorry to be nitpicky, but..
I think I prefer using `inspect` here, since it's not REALLY json that we want.

Insti:
I think this is pretty much ready now. :+1:
@kotp?

lib/word_count_cases.rb
pvcarrera: Yep, I agree `inspect` reveals the intention better than `to_json`. Thanks

Insti:
Thanks for all your work on this @pvcarrera :heart:

pvcarrera:
I'm really happy to help, thanks a lot for all of your feedback 😍

kotp:
Thanks @pvcarrera :tophat: to @Insti

PR #450, opened 2016-10-10
Add test for 10 card; add test comparing 5-high straight to other str…

### LINKED ISSUE #446: Poker test suite
The poker exercise test suite should include at least one test that includes a '10' card to test the integrity of the rank/suit parser.

One of the tests is called `test_an_ace_high_straight`. It is actually testing a five-high straight. The name should be changed to avoid confusion.

There should be a test of a five-high straight against another straight. The five-high straight is the lowest possible straight, and it constitutes a special case in which the ace ranks lower than a 2 both for purposes of determining if the straight exists (which is already tested) and for scoring purposes (which is not currently tested).

There should also be a test comparing high-card hands that are identical except for the last card, to ensure the scoring algorithm is complete.

I have written tests for these cases; they can be found [here](http://exercism.io/submissions/56200d8c069b4786be1e2ec3b8bdbeb1).

moveson:
…aight; add test comparing two high-card hands where all cards but the last are equal in rank; update to version = 2. Also changed the name of `test_an_ace_high_straight` to `test_a_5_high_straight` and changed some other test names and variables to match poker convention of reporting two pair with highest pair first. Closes #446.

### DIFF
--- a/exercises/poker/example.rb
+++ b/exercises/poker/example.rb
@@ -1,80 +1,138 @@
 module BookKeeping
-  VERSION = 1
+  VERSION = 2
 end
 
 class Poker
-  def initialize(hands)
-    @hands = hands
+
+  def initialize(hand_arrays)
+    @hands = hand_arrays.map { |hand_array| Hand.new(hand_array) }
   end
 
   def best_hand
-    best_score = @hands.map { |hand| hand_rank(hand) }.max
-    @hands.select { |hand| hand_rank(hand) == best_score }
+    hands.select { |hand| hand.score == highest_score }.map(&:to_a)
   end
 
   private
 
-  def hand_rank(hand)
-    hand_parser(hand)
-    top_hand || low_hand
+  attr_reader :hands
+
+  def highest_score
+    hands.map(&:score).max
   end
 
-  def hand_parser(hand)
-    @card_ranks = hand.map { |ele| '..23456789TJQKA'.index(ele[0]) }
-    @card_suits = hand.map { |ele| '.HSDC'.index(ele[1]) }
-    @card_split = @card_ranks.each_with_object(Hash.new(0)) { |r, h| h[r] += 1 }
-    @groups = @card_split.values.sort.reverse
+end
+
+class Hand
+
+  def initialize(hand_array)
+    @hand_array = hand_array
+    @cards = hand_array.map { |rank_and_suit| Card.new(rank_and_suit) }
   end
 
-  def top_hand
-    straight_flush || square || full || flush || straight
+  def to_a
+    hand_array
   end
 
-  def low_hand
-    three_of_a_kind || double_pair || pair || 0 + @card_ranks.max
+  def score
+    [hand_score, card_score].flatten
   end
 
-  def straight_flush
-    800 + @card_ranks.max if straight? && flush?
+  private
+
+  attr_reader :hand_array, :cards
+
+  def hand_score
+    scoring_hands.map.with_index { |scoring_hand, i| i if scoring_hand }.compact.max
   end
 
-  def square
-    700 + @card_split.key(4) if @groups == [4, 1]
+  def card_score
+    five_high_straight? ? [5, 4, 3, 2, 1] : card_score_array
   end
 
-  def full
-    600 + @card_split.key(3) if @groups == [3, 2]
+  def card_score_array
+    rank_count_hash
+        .sort_by { |rank, count| [-count, -rank] }
+        .map { |count_rank_array| count_rank_array[0] }
   end
 
-  def flush
-    500 + @card_ranks.max if flush?
+  def scoring_hands
+    [high_card?, one_pair?, two_pair?, three_of_a_kind?, five_high_straight?,
+     straight?, flush?, full_house?, four_of_a_kind?, straight_flush?]
   end
 
-  def straight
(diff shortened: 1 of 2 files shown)


kotp:
Don't forget to add the corresponding `example.rb` to show proof of concept.

Insti:
@moveson ,  You'll need to remove the `poker.rb` file and only have `example.rb` as the example solution.

tejasbubane:
Also add `closes #446` to the description so that the issue is automatically closed when this PR is merged.

moveson:
Closes #446

tejasbubane:
@moveson Not sure if a comment would work, you will need to edit your [first description](https://github.com/exercism/xruby/pull/450#issue-181959150)  and add it there. 

Refer: https://github.com/blog/1506-closing-issues-via-pull-requests

PS: This is not super-important, but just that we don't forget to close the issue after this is merged.

moveson:
OK, thanks for your patience. This is my first collaborative project for which I am not the owner.

Insti:
This is failing the CI Check due to a Rubocop issue:
```
9.26s$ rubocop -fs -D
== exercises/poker/poker_test.rb ==
C: 62:  5: Style/VariableName: Use snake_case for variable names.
C: 63:  5: Style/VariableName: Use snake_case for variable names.
```

The variables in question are these, imo sensibly named, variables:
```
62: +    double_pair_K_and_2 = %w(2S KH 2S KD JH)
63: +    double_pair_Q_and_J = %w(JS QH JS 8D QH)
```

The best solution for now is probably to wrap the `test_another_two_double_pair` method with the Rubocop disabling comments: 
`#rubocop:disable Style/VariableName`
and
`#rubocop:enable Style/VariableName`

moveson:
I started looking more closely at local variable names in the tests, which made me look more closely at the cards used in the tests, which led me to add validations to uncover any duplicate cards in the tests. There were a surprising number of invalid hands and hand comparisons because of duplicate cards, and these have been fixed in the latest push.

kotp:
If those are invalid hands only because of duplicate cards, realize that there is nothing indicating a single deck hand.  Indeed, the tests even support a multi-deck hand.  Before removing this, consider the value of this discussion on the exercism site for premature validation or defensive programming.

moveson:
I have been assuming the exercise is for single-deck poker. If the intention is to support multi-deck poker, then shouldn't the highest scoring hand be 5-of-a-kind? I don't feel strongly one way or the other, but this discussion indicates that the exercise should specify one way or the other, and of course the tests should reflect the stated expectation one way or the other.

Insti:
I vote for single deck, poker gets more complicated with more decks!

kotp:
Don't forget how the tests can drive implementation, and may remove some discussion from the iterations.  This has no tests that test for "valid hands" in a single deck use, that I have spotted.  There is code in the example for this though.  The example.rb should be driven by the tests... I have not gone through extensively to ensure this is happening at the moment.

moveson:
Yes, understood. I wrote the hand-validation code to ferret out duplicates and clean up the tests (this is really hard to do by just eyeballing them), but I understand that hand validation is not strictly necessary and may make the problem more complicated than it needs to be. I think the best course of action moving forward should be:

1. State expressly in the README file that this exercise relates to single-deck poker
2. To avoid confusion, update the tests to remove duplicate cards (as the PR already reflects)

and then either:

3(a) Add tests to require a raise if there is card duplication, or
3(b) Remove the hand-validation code from the example.

I feel 1 and 2 are important but don't feel strongly about the choice between 3(a) and 3(b).

kotp:
I say leave the possibility for assumption, write the tests to the most "default" position (being a single test), don't test for duplication (this forces a single deck).  This can bring up the discussion of "How easily would it be to expand this to multiple decks?" etc.  Removing the duplicate card(s) is fine, I think.

Even having the duplication in the test itself in no way invalidates any requirement at the moment.  But it does open up the discussion as to why it may be there, what the tests are trying to communicate (That they are not complete, that it may be remnants of a real life project that does involve multiple decks, etc.).

I like having the work done that you did, it doesn't enforce a single deck, only lightly suggests it, it leaves a discussion open.

moveson:
OK, so not sure what you want me to do next. Should I remove the hand-validation code from example.rb? Are any changes needed in the tests?

kotp:
I don't think there are changes in the tests necessary.  Yes for the cleanup for the code in example.rb.  Anything that is not driven by the tests, and not mentioned in the Readme.

If you are interested, I think the next steps would be to start the process for exercism/x-common for putting together the common data.  I think Hamming exercise may be a good example to work from.  That will bring in other tracks and conversations about some things mentioned in this discussion should be covered.

After that, the following step would be writing the generator for this track, based on that common data.

Are you game?

kotp:
Looks good to me, let's squash the commits as appropriate.

moveson:
Would you like me to squash commits on my end and force a push? Or will you squash commits during the merge?

Insti:
Best if you squash them, as you are the best person to make sure that the log message contains all the relevant information from the squashed commits.

Insti:
Looks good to me, I'll let @kotp merge this.

kotp:
Thanks @moveson I just brought the changes in.  I think there is some time before fetching will be bringing in the new tests, but it should start happening soon.

PR #451, opened 2016-10-11
Common generate script

bmulvihill:
While creating a generator for acronym and noticed how each exercise had it own generate script. This leads to a lot of code duplication since each script is almost identical. 

This PR would allow for a single generate script that takes an exercise name as the argument, and regenerates the exercise.  If the exercise is not found and error message will display.  I tested it out on all the exercises and I had to modify two generators to ensure consistent naming.

Example usage:
`bin/generate raindrops`

Let me know if this is something that would be helpful!


Insti:
Nice @bmulvihill, This is something I'd been planning on doing when I found some time, but it looks like you beat me to it.

bmulvihill:
@kotp @insi the contributing doc would need to change slightly.  I can update it in this PR.  I can also remove all the old generate scripts.

kotp:
That would be awesome!

bmulvihill:
@kotp @Insti Ok this should be good to go

kotp:
Was blocking #454

Issue #453, opened 2016-10-12
Alphametics: Update tests


Insti:
The common [alphametics test data has been updated](https://github.com/exercism/x-common/pull/402) so the tests need to be regenerated.

kotp:
This is a straggler, it should have been "autoclosed" with a commit comment from 455.

PR #455, opened 2016-10-13
Update tests for Alphametics

mattkimmel:
Updated the tests for Alphametics as per #453. I kept `test_solve_long_puzzle` commented out, since it certainly does take a long time to run. I also observed that `test_solve_puzzle_with_many_words` has the same problem, and commented that out as well.


Insti:
Great! Thanks :heart:

mattkimmel:
Happy Hacktoberfest! 😄

PR #457, opened 2016-10-16
remove superfluous `require time` from gigasecond test

hilary:


kotp:
Thank you @hilary.

This test file is generated.  I think the change belongs in the `exercises/gigasecond/example.tt` file, which will generate the test file.

hilary:
Doh! Learning my way around ... and here I thought I had read through the docs 😊

kotp:
Thank you @hilary it is all good...

kotp:
I missed it, doh!  Time was not meant to be removed from here, as Time is being used as a convenience in the tests, not a requirement for the solution.  This makes the solution a bit less flexible.

exercises/gigasecond/example.tt
```diff
@@ -1,5 +1,4 @@
 require 'minitest/autorun'
-require 'time'
```
kotp: This should not have been removed.

PR #458, opened 2016-10-17
Generate test files without requirement on time

kotp:
re: #457


Insti:
Ah ok, because #457 already removed it.

Insti:
Thanks @kotp :heart:

PR #461, opened 2016-10-17
bin/generate usage instructions.

Insti:
Give usage instructions when bin/generate is called with no arguments.
The instructions also provide a list of the available generators.

Edit: Here's what it looks like:

``` shell
$ bin/generate 
Exercise name required!
Usage: bin/generate exercise_generator

Available exercise generators:
acronym alphametics binary bracket_push clock connect custom_set difference_of_squares gigase
cond hamming hello_world largest_series_product leap nth_prime pangram raindrops rna_transcri
ption roman_numerals run_length_encoding sieve two_bucket
```

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -1,10 +1,29 @@
 #!/usr/bin/env ruby
 
-exercise = ARGV[0]
-cases = "#{exercise.tr('-','_')}_cases"
-
 require_relative '../lib/helper'
 require 'generator'
+
+def available_generators
+  cases = File.join( __dir__, '../lib/*_cases.rb')
+  Dir[cases].map {|filename| /([^\/]*)_cases\.rb$/.match(filename).captures}
+end
+
+def usage
+  "Usage: #{$PROGRAM_NAME} exercise_generator\n\n" +
+  "Available exercise generators:\n" +
+  available_generators.sort.join(' ')
+end
+
+exercise = ARGV[0]
+
+unless exercise
+  STDERR.puts "Exercise name required!\n"
+  puts usage
+  exit
+end
+
+cases = "#{exercise.tr('-','_')}_cases"
+
 begin
   require "#{cases}"
 rescue LoadError
@@ -14,3 +33,4 @@ end
 
 klass = Object.const_get(cases.split('_').map(&:capitalize).join)
 Generator.new(exercise, klass).generate
+


bmulvihill:
Works good for me - one style note, do you think it would be better to separate the list of generators with a newline for readability?

```
Exercise name required!
Usage: bin/generate exercise_generator

Available exercise generators:
acronym
alphametics
binary
bracket_push
clock
connect
custom_set
difference_of_squares
gigasecond
hamming
hello_world
largest_series_product
leap
nth_prime
pangram
raindrops
rna_transcription
roman_numerals
run_length_encoding
sieve
two_bucket
```

Insti:
Thanks for checking it out :heart:

>  do you think it would be better to separate the list of generators with a newline for readability?

That's a good suggestion, but I'd prefer not to do that because it pushes the useful part of the error message off the screen. (My terminal windows are usually about 25 lines high.) 

But I would also like to add a command line option that lists all the available generators, and that could be a good way to display them. (But that would be a different PR.)

kotp:
It is OK to have the list columnized, single column or as it is.  The important thing when considering this is how does the message present?  Is it to standard out or standard error?  Can I page it if I need to and can I search for what I am looking for?

kotp:
What is the significance of signal 2 for the exit when no exercise is given?

Insti:
> What is the significance of signal 2

"A generator does not currently exist for #{exercise}!" returned `1` and I wanted the error code to be distinct, so I just used the next number :smile:

However it seems that [2 is a reserved code](http://www.tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF)

So I'll bump it up to 3.

Insti:
> Is it to standard out or standard error? Can I page it if I need to and can I search for what I am looking for?

Currently it uses STDOUT. I think this is fine since we're not using STDOUT for the action of the command (which alters the files directly.) 

I guess the answer to the other 2 questions is 'yes'.

kotp:
I might give it an exit status of 0, since it is not an error but an informational move.  It may be done in a series of commands at the command line, and given a non-zero response, it would interrupt that sequence.

I would make sure that any purposeful non-zero status is the right move, it does effect things pretty fundamentally.

Insti:
Good point, it now returns 0.

bin/generate
```diff
@@ -1,6 +1,24 @@
 #!/usr/bin/env ruby
 
 exercise = ARGV[0]
```
kotp: This line should be after the helper methods definitions.

bin/generate
```diff
@@ -1,6 +1,24 @@
...
+end
+
+unless exercise
+  puts usage
+  exit
+end
+
 cases = "#{exercise.tr('-','_')}_cases"
 
 require_relative '../lib/helper'
```
kotp: This should be at the top of the file.

kotp:
It may be useful to provide stderr output for the error part and the stdout for the informational part.

kotp:
@insti can you squash the commits, I think the message will need to be change to give the intended information.  And then feel free to bring it in, I think.

Insti:
> I think the message will need to be change to give the intended information.

What do you mean by this?

kotp:
If you squash it, the message may look like this:

```
bin/generate usage instructions.

Give usage instructions when bin/generate is called with no arguments.
The instructions also provide a list of the available generators.

Return 0 on exit when providing usage information.

It wasn't REALLY an error situation, the script completed correctly, you
just didn't provide enough arguments for it to do anything else.

Move 'require's up to the top of the file.

```

But you had done the "Edit" in the first message, so that information will be missing... so "the message will need to be changed to give the intended information." based on your edit.

PR #462, opened 2016-10-17
Bin files independent of pwd  (Fixes #291)

### LINKED ISSUE #291: bin files not independent of pwd
For example, in `enable-executable` and `executable-tests-check`, we are using a simple `Dir.glob`, which is relative to pwd, not the file.
Care should be taken to preserve legible output (instead of full, absolute paths).

Insti:
Use path relative to the executable Ruby file in the xruby 'bin' directory rather than the current working directory which was the default behaviour.

This will enable the commands to be run from any working directory and still be able to access the files they need to check/update.

fixes #291

### DIFF
--- a/bin/enable-executable
+++ b/bin/enable-executable
@@ -2,14 +2,18 @@
 
 require 'fileutils'
 
-puts 'Ensuring test files are executable...'
+puts 'Ensuring test files are executable.'
 
-(Dir.glob('**/*test.rb') + Dir.glob('bin/*')).each do |f|
-  if File.executable?(f)
+# Assume that this file lives in #{base}/bin
+base = File.join(__dir__,'..')
+files = Dir.glob("#{base}/**/*test.rb") + Dir.glob("#{base}/bin/*")
+
+files.each do |file|
+  if File.executable?(file)
     print '.'
   else
-    print "\nAdding exec bit to #{f}"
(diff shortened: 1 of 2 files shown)


bin/enable-executable
```diff
@@ -2,14 +2,18 @@
...
 require 'fileutils'
 
-puts 'Ensuring test files are executable...'
+puts 'Ensuring test files are executable.'
 
-(Dir.glob('**/*test.rb') + Dir.glob('bin/*')).each do |f|
-  if File.executable?(f)
+# Assume that this file lives in #{base}/bin
+base = File.join(__dir__,'..')
+files = Dir.glob("#{base}/**/*test.rb") + Dir.glob("#{base}/bin/*")
```
Insti: Added an intention revealing variable here.

bin/enable-executable
```diff
@@ -2,14 +2,18 @@
...
+
+files.each do |file|
+  if File.executable?(file)
     print '.'
   else
-    print "\nAdding exec bit to #{f}"
-    FileUtils.chmod('u+x', f)
+    print "\nAdding exec bit to #{file}"
+    FileUtils.chmod('u+x', file)
   end
```
Insti: Replaced single letter variable `f` with the more descriptive `file`
It probably could be `filename`

Issue #452, opened 2016-10-11
Common generator interface


bmulvihill:
While working on a generator I noticed most exercise generators ({exercise-name}_cases.rb) have a different method names for the same purpose. For example the test values, some generators call this method assertion, others call it do, or work_load.  The same goes for the test naming and skip methods.  This can lead to some confusion and also code duplication.

I think a common interface would go a long way towards reducing some code duplication and making it easier to create new generators.

Insti:
I'd also like to see the generator files move into the exercise sub directories.
(This will need support for `.meta` added in x-api or trackler or both. see:  https://github.com/exercism/x-api/issues/113)

But would mean that all the code for generating the tests for a particular exercise lives with that exercise, eg:

`exercises/slug/.meta/generator/slug_cases.rb`
`exercises/slug/.meta/generator/generic_template.tt`
`exercises/slug/.meta/generator/slug_template.tt`

(This may be a different issue, but I thought I'd add it here for now as a wish.)

kotp:
I think `workload` has been discussed for this.  I would like to see the same terms used as well.

The `skipped` method used in `lib/hamming_cases.rb` and the related `exercises/hamming/example.tt` file is my preferred way, reducing complexity in the template file (the view).

kotp:
Thank you @bmulvihill this is being addressed via the generator work currently done with @hilary  and previously done with @tommyschaefer and @Insti and others.

PR #463, opened 2016-10-18
All generators use common interface

bmulvihill:
Finishes https://github.com/exercism/xruby/issues/452
Starts the work of creating a common interface.  This PR attempts to standardize 3 public methods
- **test_name** - The method will output the name of the test
- **workload** - This method will output the body of the test
- **skipped** - This method will output skip syntax
  
  I ran all the generators locally and they appear to be working as intended. Once this gets standardized, we may be able to utilize a single example.tt file.

Future ideas:
- Utilize a single example.tt for all generated tests
- Implement some sort of Generator abstract class/module (Generatable?) where subclasses are required to implement two methods: **test_name** && **workload**. We can then move common methods i.e. skipped into this class/module


kotp:
Quick visual scan says "YAY!"

.ruby-version
```diff
@@ -0,0 +1 @@
+2.3.0
```
Insti: This file probably shouldn't be updated as part of this.

lib/hello_world_cases.rb
```diff
@@ -3,12 +3,18 @@ def test_name
     'test_%s' % description.gsub(/[ -]/, '_')
   end
 
-  def do
-    defined?(name) ? "HelloWorld.hello('#{name}')" : 'HelloWorld.hello'
+  def work_load
+    assertion =
```
Insti: Can you extract this to a method?

lib/two_bucket_cases.rb
```diff
@@ -8,15 +8,15 @@ def do
     "TwoBucket.new(#{bucket_one}, #{bucket_two}, #{goal}, '#{start_bucket}')"
   end
 
-  def test_body
+  def work_load
     "two_bucket = #{self.do}
     assert_equal #{expected['moves']}, two_bucket.moves
     assert_equal '#{expected['goal_bucket']}', two_bucket.goal_bucket
     assert_equal #{expected['other_bucket']}, two_bucket.other_bucket"
```
Insti: I'd like to see work_load **not** have to be responsible for whitespace/indenting.

Insti:
The first one I tried `bin/generate acronym` didn't work correctly. :disappointed: 
The test methods had no names.

Insti:
The way to test this would be to regenerate all the test files and ensure that the output was identical to the old versions.

.ruby-version
bmulvihill: Whoops! yeah I'll remove this

lib/two_bucket_cases.rb
bmulvihill: I agree, I tried not to do too much refactoring though and keep the scope of this PR narrow - just focusing and getting the API in place

bmulvihill:
@Insti I did regenerate all of them locally, but obviously I missed something 🙂, Ill fix it in my next push!

bmulvihill:
@Insti @kotp Made some additional changes, its a lot of files so hopefully I haven't missed anything else.  As mentioned above I tried not to do any refactoring and just implement the API.

kotp:
I just tried to do the generate on word-count as it is on #448 which is rebased on master since #451 has been brought in, getting errors there as well.  In that case, it is due to single quotes inside of strings that have single quotes as the delimiters.  (That is of course the first cause, once fixed the other errors may fix, I did not look that closely).

lib/two_bucket_cases.rb
Insti: OK, that's sensible.

Insti:
@bmulvihill the conversions are still very broken. I think you need to go through them all individually and check that the new generated file is the same as the old one.

Both of the ones I "randomly" looked at: leap, acronym were very broken.
Running: `rubocop exercises` will detect syntax errors in the files.

Maybe it would be better to split this PR into smaller PRs that look at each generator individually?

Insti:
I'd like to see a PR with some documentation of what the common interface is, you've given a good summary in the OP here, but it probably belongs in the README.md

Insti:
I'd rather **work_load** be renamed **workload** if that's the name we're going with.

'work' and 'load' are 2 words with many different meanings, 'workload' is one word with a more specific meaning.

My but preference for name would be **assertion** which would make the generic template:
```ruby
    <%= test_case.skipped %>
    <%= test_case.assertion %>
  end
```

Insti:
Those last 2 probably should be separate issues so that they can be discussed individually.

I really appreciate the effort you're putting into this @bmulvihill :heart: and agree that it is an important thing to work on, so if we're going to do a lot of work getting it right, we should try to make sure we're doing the right thing and I want to try to help that be as easy as possible.

bmulvihill:
@Insti I agree with the name **assertion** over **work_load,** I believe in the original issue work_load was suggested.  Another name which might make more sense is **test_body,** since it could technically include more than just an assertion.  

I can make that change if you want and add a README update to this PR, since I think that makes sense

Insti:
@kotp, @bmulvihill is there anywhere other than #452 where naming was discussed?

> **test_body** [...] it could technically include more than just an assertion. 

Good point. I'd prefer just `body` over `test_body` since the `test_` prefix is redundant.
```
    <%= test_case.skipped %>
    <%= test_case.body %>
  end
```
Either that or go fully prefixed to try and avoid the situation where one of the json files in the future has a property called 'skipped' or 'body', or 'test_body'
```
  def <%= test_case.test_case_name %>
    <%= test_case.test_case_skipped %>
    <%= test_case.test_case_body %>
  end
```
But that seems a bit redundant.

I guess we could discuss it forever.

As the person making the PR you have the power to decide what you call it :wink: so choose what you think is best.

kotp:
I do prefer the one word `workload` as it is a single word, and means what it says.

Here is where any mention of [workload or work_load](https://github.com/issues?utf8=%E2%9C%93&q=is%3Apublic+user%3Aexercism+sort%3Aupdated-desc+workload+OR+work_load) was discussed or mentioned.

bmulvihill:
@Insti @kotp I went back through each exercise and generated/ran the tests and everything appears to be working.  What I hadn't done before, which I should have, was actually run the tests for each newly generated file.  I also renamed **work_load** to **body**.

kotp:
The name "body" I think in the discussions that we did have was determined to be too generic.  Body of what?  I think this was an important reason why `workload` or `work_load` was finally chosen.  I can't look at the discussions right now, but I hope you did at least scan through them via the link I gave, if so, you could verify, better than I can at the moment.

bmulvihill:
@kotp I scanned through the discussions, but nothing jumped out at me besides ours and one where @kytrinyx mentioned work_load. Its an easy switch to **workload** if you find that more descriptive, I think consistency is the most important thing here

Insti:
I looked through that search result and also couldn't find any discussion on naming. 
If we need to compromise, I'm OK with using **workload**

kotp:
I would go with what @kytrinyx mentioned, because it is likely to be the same in other tracks.

bmulvihill:
@Insti @kotp This has been rebased, and the method has been renamed to **workload** - I also added in the two new generators that have been added since this opened, word-count and anagram

kotp:
I think I sent you a message on Gitter @bmulvihill Let me know if you can't get to it.

Can you look at this error, when I generated all files, ran `make` to run the tests (and this takes a while due to alphametics) but got this error.

```text
running tests for: alphametics
Run options: --seed 62490

# Running:

.F.....

Finished in 306.960573s, 0.0228 runs/s, 0.0228 assertions/s.

  1) Failure:
AlphameticsTest#test_solve_puzzle_with_many_words [/tmp/alphametics.7i3sjoNLoE/alphametics_test.rb:60]:
--- expected
+++ actual
@@ -1 +1 @@
-{"A"=>5, "D"=>3, "E"=>4, "F"=>7, "G"=>8, "N"=>0, "O"=>2, "R"=>1, "S"=>6, "T"=>9}
+nil


7 runs, 7 assertions, 1 failures, 0 errors, 0 skips
```

bmulvihill:
@kotp That test is currently commented out - and when I run it prior to these changes it fails the same way

kotp:
Alright.  I will look into it more tonight.  Wonder why it tests if it is commented out.

bmulvihill:
@kotp I think maybe the tests were manually commented out, so when it was regenerated it doesn't have the comments

kotp:
Yep, I think we hit this before, where we would have to manually comment them out, which is not something I think we should have to remember to do.  Unfortunately, there is nothing in the canonical to indicate these are "long running tests" to help to programmatically trigger this commenting behavior.

Insti:
It is possible to detect and special case them in the `*_cases.rb` file though.

bmulvihill:
@Insti @kotp I updated the README as well.  As far as programmatically triggering the comments, I believe that should be a different PR, do you agree?

Insti:
@bmulvihill it probably is a different PR, but it needs to happen before this one can be merged as the tests need to run at a reasonable speed.

bmulvihill:
@Insti hmmm ok - I'll take a look at that

Insti:
@bmulvihill I'm currently working on the alphametics generator.

Insti:
Alphametics generator: https://github.com/exercism/xruby/pull/469

bmulvihill:
@Insti @kotp Im going to close this PR - I think since the README is updated and the generator implementation is still kind of up in the air we can tackle this on a smaller scale

Insti:
@bmulvihill fine, thanks for getting us going on this. It started a lot of useful discussion.

I need to find some time to work on my new generator branch again.

PR #464, opened 2016-10-18
Add anagram generator

fredrb:
#396 - Anagram exercise generator


exercises/anagram/anagram_test.rb
```diff
@@ -3,70 +3,120 @@
...
 require_relative 'anagram'
 
+# Test data version:
+# 3dfde31
 class AnagramTest < Minitest::Test
   def test_no_matches
-    detector = Anagram.new('diaper')
-    assert_equal [], detector.match(%w(hello world zombies pants))
+     # skip
+     assert_equal [], Anagram.new("diaper").match(["hello", "world", "zombies", "pants"])
```
Insti: indentation should be 2 spaces, you've got 3 here.

lib/anagram_cases.rb
```diff
@@ -0,0 +1,24 @@
+class AnagramCase < OpenStruct
+  def test_name
+    'test_%s' % description.gsub(/[ -]/, '_')
+  end
+
+  def work_load
+    actual = "Anagram.new(\"#{subject}\").match(#{candidates})"
```
Insti: If you use single quotes around #{subject} you don't need to escape the quotes.

lib/anagram_cases.rb
```diff
@@ -0,0 +1,24 @@
...
+    "assert_equal #{expected}, #{actual}"
+  end
+
+  def skipped
+    index.zero? ? '# skip' : 'skip'
+  end
+
+  def show_comment
+    comment.nil? ? '' : "# #{comment}\n\t\t"
```
Insti: `\t` is the wrong character to be using.
Ideally there would be no formatting characters in this file.
Perhaps you can find a solution that leaves the formatting in the template file and the logic in this file?

exercises/anagram/anagram_test.rb
```diff
@@ -3,70 +3,120 @@
...
+
+  def test_anagrams_must_use_all_letters_exactly_once
+     skip
+     assert_equal [], Anagram.new("tapper").match(["patter"])
+  end
+
+  def test_detects_unicode_anagrams
+     skip
+     # These words don't make sense, they're just greek letters cobbled together.
+		assert_equal ["ΒΓΑ", "γβα"], Anagram.new("ΑΒΓ").match(["ΒΓΑ", "ΒΓΔ", "γβα"])
```
Insti: Use spaces rather than tabs for indentation.

exercises/anagram/anagram_test.rb
```diff
@@ -3,70 +3,120 @@
...
   end
 
-  def test_detect_simple_anagram
-    skip
-    detector = Anagram.new('ant')
-    anagrams = detector.match(%w(tan stand at))
-    assert_equal ['tan'], anagrams
+  def test_detects_simple_anagram
+     skip
+     assert_equal ["tan"], Anagram.new("ant").match(["tan", "stand", "at"])
```
Insti: Why have you changed from the old 3 line "detector, anagrams, assertion" format?

exercises/anagram/anagram_test.rb
```diff
@@ -3,70 +3,120 @@
...
+     assert_equal [], Anagram.new("good").match(["dog", "goody"])
   end
 
-  def test_eliminate_anagrams_with_same_checksum
-    skip
-    detector = Anagram.new('mass')
-    assert_equal [], detector.match(['last'])
+  def test_detects_anagram
+     skip
+     assert_equal ["inlets"], Anagram.new("listen").match(["enlists", "google", "inlets", "banana"])
```
Insti: This is a personal preference of mine and non necessarily something that needs changing if you have a different opinion, but I'd rather see single quotes around strings that don't require interpolation.

Insti:
Travis CI is failing because the example solution does not pass all the tests:
```
  1) Failure:

AnagramTest#test_detects_unicode_anagrams [/tmp/anagram.oOJIjvdSpC/anagram_test.rb:87]:

Expected: ["ΒΓΑ", "γβα"]

  Actual: ["ΒΓΑ"]
```

Insti:
Thanks @fredrb :heart: It's great to see people contributing new generators.
You've covered all the important functionality and the things to change are are mostly just minor stylistic and consistency issues. 

Great work.

Insti:
> Expected: ["ΒΓΑ", "γβα"]
>   Actual: ["ΒΓΑ"]

I see you've raised [an issue about this in x-common](https://github.com/exercism/x-common/issues/413).
That was the right thing to do.

lib/anagram_cases.rb
fredrb: I was enforcing this because I didn't want a blank line in the generated file if there are no comments for the test case. But it makes sense that no formatting should be done here.

exercises/anagram/anagram_test.rb
fredrb: Agreed. But both the `expected` and `matched` lists come from the canonical json file.

fredrb:
Thanks for the review @Insti. I fixed most of the things you pointed out and left a comment on others. I'll check the issue with the greek letters in x-common then, if it gets fixed, I'll re-generate the files here.

lib/anagram_cases.rb
Insti: There are other ways to achieve the same effect.

exercises/anagram/anagram_test.rb
```diff
@@ -7,95 +7,111 @@
...
+    
+    assert_equal ["inlets"].sort, Anagram.new('listen').match(["enlists", "google", "inlets", "banana"]).sort
   end
 
   def test_detects_multiple_anagrams
-     skip
-     assert_equal ["gallery", "regally", "largely"], Anagram.new("allergy").match(["gallery", "ballerina", "regally", "clergy", "largely", "leading"])
+    skip
+    
+    assert_equal ["gallery", "regally", "largely"].sort, Anagram.new('allergy').match(["gallery", "ballerina", "regally", "clergy", "largely", "leading"]).sort
```
Insti: You didn't address my other question:
> Why have you changed from the old 3 line "detector, anagrams, assertion" format?

That would help avoid situations like this where you end up with one really long line.

lib/anagram_cases.rb
```diff
@@ -4,16 +4,16 @@ def test_name
...
+    "assert_equal #{expected}.sort, #{actual}.sort"
   end
 
   def skipped
     index.zero? ? '# skip' : 'skip'
   end
 
   def show_comment
-    comment.nil? ? '' : "# #{comment}\n\t\t"
+    comment.nil? ? '' : "# #{comment}"
```
Insti: Try moving the comment outputing into the `work_load`, this should allow you to avoid empty lines.

fredrb:
@Insti, I implemented the changes you suggested, could you please review again?
Our tests are failing because exercism/x-common#414 is still pending.

lib/anagram_cases.rb
```diff
@@ -0,0 +1,41 @@
...
+    ident_lines([show_comment, detector, anagram, assert].compact)
+  end
+
+  def skipped
+    index.zero? ? '# skip' : 'skip'
+  end
+
+  private
+
+  def ident_lines(code, ident = 2)
```
Insti: `indent` is missing an `n`

Insti:
x-common/pull/414 has now been merged.

exercises/anagram/anagram_test.rb
```diff
@@ -3,70 +3,156 @@
...
   end
 
-  def test_detect_simple_anagram
+  def test_detects_simple_anagram
     skip
     detector = Anagram.new('ant')
-    anagrams = detector.match(%w(tan stand at))
-    assert_equal ['tan'], anagrams
+    anagrams = detector.match(["tan", "stand", "at"])
+    assert_equal ["tan"].sort, anagrams.sort
```
Insti: (This is an 'in an ideal world' type issue please don't think that you **have** to do this.)

Based on the principle that we want to make the test cases as nice as possible for people and would prefer generator complexity over test case complexity. 
It could be nice to omit the `.sort` when there is only one element, and pre-sort the 'expected' arrays that have multiple elements.

lib/anagram_cases.rb
```diff
@@ -0,0 +1,41 @@
...
+  def detector
+    "detector = Anagram.new('#{subject}')"
+  end
+
+  def anagram
+    "anagrams = detector.match(#{candidates})"
+  end
+
+  def assert
+    "assert_equal #{expected}.sort, anagrams.sort"
```
Insti: "ideal world" suggestion: (see my other comment)

`assert_equal #{expected.sort}, anagrams#{expected.size > 1 ? '.sort' : ''}`

or something more elegant.

exercises/anagram/example.tt
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env ruby
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'anagram'
+
+# Test data version:
+# <%= sha1 %>
+class AnagramTest < Minitest::Test<% test_cases.each do |test_case| %>
```
Insti: Can we move the each down to the next line?

fredrb:
Good idea omiting the `.sort` when expected list has either a single value or it's empty. If you check the test class, only two test cases ended up with `anagrams.sort` :+1:

Insti:
I think this is ready to go now, I'll just wait a bit and give @kotp a chance to look over it before merging.

Thanks for your work on this and your responsiveness to feedback, I think we've ended up with a really nice generator AND nice test cases. Great work. :heart:

fredrb:
@Insti It was a pleasure! Thanks for the code reviews and suggestions

kotp:
@fredrb can you squash these commits into one or (if you think it makes sense) 2 commits.  It makes it easier to review, and gives structure to the commit message when it comes in, as to what you would like communicate in the log.

I end up doing it locally, to review, but it is nice to give you a chance to cultivate the commit, as I may miss something.

fredrb:
Squashed them into one, @kotp. Also made the small fix for the trailing whitespace.

kotp:
Thank you @fredrb

PR #465, opened 2016-10-19
Replace hard-coded Bowling tests with test generator (Fixes #456)

### LINKED ISSUE #456: Changes to Bowling tests
We recently rewrote the test suite for Bowling. Since this track implements Bowling, please [take a look at the new canonical_data.json file](https://github.com/exercism/x-common/blob/9049dc7f748fd038b140c878c4eaa89d09b73199/exercises/bowling/canonical-data.json) and see if your track should update its implementation or tests.

The new test suite reorders tests so that students can get to green quickly. It also reduces the number of tests so that students can focus on solving the interesting edge cases.

More details on this change are available in [this issue](https://github.com/exercism/x-common/issues/409)

gchan:
As per #456

Also updated the example solution to address new test case

Feedback is appreciated. I ran `rubocop` but I'm uncertain if it actually was able to detect offences (I deliberately indented incorrect to test `rubocop`).


Insti:
Awesome, thanks.
I'll have a look through this shortly.

The current `rubocop` config is very permissive and doesn't detect much.

exercises/bowling/bowling_test.rb
```diff
@@ -3,224 +3,199 @@
...
+class BowlingTest < Minitest::Test
   def setup
     @game = Game.new
   end
 
-  def test_must_be_able_to_roll_with_number_of_pins
-    assert_respond_to @game, :roll
-    assert_equal 1, @game.method(:roll).arity
+  def roll(rolls)
+    rolls.each { |pins| @game.roll(pins) }
```
Insti: I like that you've extracted this helper method. Good variable naming too :+1:

exercises/bowling/example.tt
```diff
@@ -0,0 +1,31 @@
...
+class BowlingTest < Minitest::Test
+  def setup
+    @game = Game.new
+  end
+
+  def roll(rolls)
+    rolls.each { |pins| @game.roll(pins) }
+  end
+<% test_cases.each do |test_case| %>
+  def <%= test_case.name %><% if test_case.skipped? %>
```
Insti: Try to keep logic out of the template file.
Have a look at the PR for anagram generator: https://github.com/exercism/xruby/pull/464 for some examples of how to do this.

Basically the test template should look like:

```ruby
    <%= test_case.skipped %>
    <%= test_case.work_load %>
  end
```

Insti: Also we're trying to standardise the method names in generators, so using `test_name`, `skipped`, `work_load` is preferred.

gchan:
Thanks, I have moved the logic into the cases file. Do you have any ideas how I could improve `work_load`?

lib/bowling_cases.rb
```diff
@@ -0,0 +1,47 @@
...
+  def test_name
+    'test_%s' % description.downcase.tr(' ', '_')
+  end
+
+  def skipped
+    index.zero? ? '# skip' : 'skip'
+  end
+
+  def work_load
+    if assert_error?
```
Insti: You're checking `assert_error?` twice, (again on line 30)
What if you moved all this logic into the `assert` method?

lib/bowling_cases.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  private
+
+  def roll
+    "roll(#{rolls})"
+  end
+
+  def assert
+    "assert_equal #{expected}, @game.score" unless assert_error?
```
Insti: hint: `assert` could return an array of lines that get joined later..

lib/bowling_cases.rb
gchan: Oops, the `assert_error?` conditional check isn't required in `assert`.

If we moved the logic to the `assert` method, we will still have one `assert_error?` check?

lib/bowling_cases.rb
gchan: Something like this?

```
  def assert
    if assert_error?
      [
        "assert_raises StandardError do",
        "  #{roll}",
        "  @game.score",
        "end"
      ]
    else
      [roll, "assert_equal #{expected}, @game.score"]
    end
  end
```

Insti:
Thanks @gchan, This looks much better now. 

I'd personally refactor `assert` but it works fine as it is and I'm trying to cut down on micro-managing.

Nice work updating the example.rb to pass the tests. :+1:

kotp:
For Running rubocop with your own configuration (or a default) you can use this:

```text
Style/Documentation:       
  Exclude:                 
    - './spec/**/*' 
    - './test/**/*' 
```

Copy that and create a file named, `default_rubocop.yml` or whatever you want to name it, wherever you want to store it.

Then invoke `rubocop -c where_I_stored_it/default_rubocop.yml` and you will get the default cops to run...

gchan:
Thanks for the feedback @Insti . The change to move the logic out of the view was a good idea.

Also thanks @kotp for the tip for running rubocop with default or custom configuration.

kotp:
Up to you @gchan if you do want to do the quick updates based on feedback from Rubocop.  Or we can bring it in, according to me.  @Insti has been looking at it closer than I have, I think.  I will leave it up to his and your decision to bring it in, either as it stands or with those slight modifications.

gchan:
There are some Rubocop violations that can be addressed in `lib/bowling_cases.rb` and `exercises/bowling/example.rb`. I can look at making those slight modifications tomorrow before we bring it in.

kotp:
Thanks @gchan.

gchan:
Corrected some Rubocop violations in a separate commit. Let me know if you need me to squash the commits.

kotp:
The way the two commits are now, one is style related, the other is the actual "Cargo".  So these two commits are good unsquashed, if you like.  Together, you would not have to show the changes.  It is up to you. But the style changes may be "noise" rather than "signal".

exercises/bowling/example.rb
```diff
@@ -20,7 +20,7 @@ def roll(pins)
   end
 
   def validate(pins)
-    raise 'Invalid number of pins' unless (RULES[:MIN]..RULES[:MAX]).cover?(pins)
+    raise 'Invalid number of pins' unless (PINS[:MIN]..PINS[:MAX]).cover?(pins)
```
Insti: According to this `1.5` is a valid number of pins.

Insti: The correctness of `example.rb` is not super important, just something I noticed while I was looking through the code.

kotp: Nice catch.  Do we test for this?  I don't think it is an important aspect in relation to what things are learned from the exercise.  But definitely fodder for discussion in solutions!

I think when we were putting this together we didn't even consider it.

gchan: That's a good point. I guess the xcommon data doesn't test for non-integer numbers given that it needs to support languages that are typed?

gchan:
You're right about the style changes being 'noise' and is better off as a separate commit. 

Is there anything else I need to do on this PR to bring it in?

exercises/bowling/example.rb
kotp: So I think for this aspect, the discussion should go to x-common... find out if this is an aspect for others, and if we can change it in common.

Insti: The kata is about scoring bowling, not worrying if your test inputs are valid. So I would argue against adding test cases for it. 
It *is* something that could be discussed during solution reviews.

exercises/bowling/bowling_test.rb
```diff
@@ -3,224 +3,200 @@
...
     end
   end
 
-  def test_should_not_calculate_score_before_fill_balls_have_been_played
+  def test_bonus_rolls_for_a_strike_in_the_last_frame_must_be_rolled_before_score_can_be_calculated
     skip
-    roll_n_times(10, 10)
+    assert_raises StandardError do
+      roll([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10])
+      @game.score
```
Insti: The error you are detecting is raised by the roll method, the way it is called hides this.
I'd like to see it made explicit that game.roll is what is raising the exception.

The way it is now @game.score is never executed, but it's still sitting there in the test looking like it is what will cause the problem. This is very misleading.

gchan: Some of the test cases expect an exception to be raised by `game.roll` and an invalid roll while some are raised by `game.score` when the game is incomplete. 

As there is no data in x-common to indicate whether the exception could be raised by `score` or by `roll`, it is difficult to procedurally generate test cases to include or exclude `game.score` from  test cases. I'm open to any ideas.

The x-common data for exceptions sets the expected value as -1 and the decision of when to error is left up to the implementation (error when an invalid roll is made, error when scoring a game with an invalid roll, or error in both cases). With the current test cases, an implemented solution may decide to only error on `game.score`, `game.roll`, or both. This way a person could implement a solution that never errors on `roll` but only error on `score`.

Insti: Thanks for that good explanation. 

I think you've made the right decision and we need to leave the tests the way they currently are.

We can revisit this if necessary if/when the canonical data is updated.

Insti:
Thanks for all your work on this @gchan :heart:

gchan:
No problem. Thanks for the feedback. I learnt a few things along the way

exercises/bowling/bowling_test.rb
kotp: I just hit the first review for bowling.  The topic of discussion will be on "custom errors" as the solution is good... Thanks @gchan and @insti for getting these changes in.

PR #466, opened 2016-10-19
Add transpose

gchan:
Test generator for the transpose exercise and example solution.


Insti:
Super, Thanks :heart: 
I'll have a look at this later today.

config.json
```diff
@@ -85,7 +85,8 @@
     "list-ops",
     "diamond",
     "all-your-base",
-    "isogram"
+    "isogram",
+    "transpose"
   ],
```
Insti: Added to the end, great, that is the right place to put new problems. :+1:

exercises/transpose/example.tt
```diff
@@ -0,0 +1,22 @@
...
+
+# Test data version:
+# <%= sha1 %>
+class TransposeTest < Minitest::Test
+<% test_cases.each do |test_case| %>
+  def <%= test_case.test_name %>
+    <%= test_case.skipped %>
+    expect = <%= test_case.expect %>
+    actual = <%= test_case.work_load %>
+    assert_equal expect, actual
```
Insti: I really like the 3 line format you're using here. 
I'd call it `expected` but `expect` is ok too.

Insti:
This looks good :heart:

:thinking: I wonder if there's some way of making the tests better show the transposed results.

gchan:
Thanks @Insti, I've renamed the variable `expect` to `expected`

Could potentially use HEREDOC or string concatenation with `\` to make the transposed results look better in the test file. WDYT?

Insti:
> Could potentially use HEREDOC or string concatenation with \ to make the transposed results look better in the test file. WDYT?

If you want to try it and see how it looks that would be good.

You could add other temporary files to the PR called `test_cases_with_heredoc.rb` or something, so we can compare the different versions. Then when we've decided which is best we can delete the redundant ones before merging.

Another thing we could do is add a comment to the test_cases.rb file asking people to submit an issue if they have an idea for a better way to display it.

gchan:
Added files for a heredoc version for test generation. 😄

Insti:
Ok, Thanks for doing that :heart:

The initial version is my preferred one.
I **really** don't like how the heredoc test cases end up looking.

exercises/transpose/transpose_test_with_heredoc.rb
```diff
@@ -0,0 +1,304 @@
...
+EXPECTED
+    assert_equal expected.strip, actual
+  end
+
+  def test_simple
+    skip
+    input = <<-INPUT
+ABC
+123
+INPUT
```
kotp: With a tilde `~` in front of the Heredoc delimiter, you can present these heredocs with the proper indentation, and they will end up looking good here and clean in the generated files.  Including indentation in side the text.  This does require Ruby 2.3 though, but it is not a hard thing to write.  Facets gem has this functionality for a fairly long time (I wrote one in 2008, unsure if Facets got my version).

kotp:
@Insti You still prefer the first version, or did the heredocs change do it for you?

Insti:
The indented heredocs are **much** nicer.
I think we should just use the heredoc version.

lib/transpose_cases.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def workload
+    'Transpose.transpose(input)'
+  end
+
+  def expect
+    [
+      "<<-EXPECTED.gsub(/^ {6}/, '')",
+      indent_lines(expected),
+      indent_line('EXPECTED')
```
Insti: I'd extract an `indented_heredoc` method.

gchan: Replace `indent_line('EXPECTED')` with `indented_heredoc('EXPECTED')?

Insti: No. replace
```
"<<-EXPECTED.gsub(/^ {6}/, '')",
indent_lines(expected),
indent_line('EXPECTED')
```
with 
`indented_heredoc(expected)`

kotp: Create a helper method to indent those things, all of them are by 6, that is the argument to the method.

`<<-EXPECTED.trim_to(6)`

gchan: Please see latest changes. Not sure if there is a `trim_to` method for Ruby `String`?

kotp: There could be. :)  You can make one.  That is what I was saying. The name would probably be `left_trim` though to be clear.

gchan: For this to work I will have to monkey patch the String class with a `left_trim` method in the generated tests?

kotp: Refinements may be an approach as well.

kotp: Consider that this is not critical.  So this should be something to think about, and play with.  The duplication here should not stop this PR.

Insti:
Thanks for all your work on this @gchan :heart:

gchan:
No problem. Thanks for the feedback. I learnt a few things along the way

PR #469, opened 2016-10-24
[alphametics] Warn about long running tests in the common test data.

Insti:
Ignore test cases based on the complexity of the expected result.

Hardcoded commented out versions of the hard(er) tests along with
instructions on why you might want to enable them.

Refactored lib/alphametics_cases.rb.
Use the common generator method names.
Tidied up expected value string generation code.

Re-generate tests based on version based on proposed changes.
https://github.com/exercism/x-common/pull/425

Update `example.rb` to pass the current version of tests.


exercises/alphametics/alphametics_test.rb
```diff
@@ -3,63 +3,72 @@
...
+  def test_puzzle_with_three_letters
     # skip
-    expect = {
-      'I' => 1, 'B' => 9, 'L' => 0
-    }
-    actual = Alphametics.new.solve('I + BB == ILL')
-    assert_equal(expect, actual)
+    input = 'I + BB == ILL'
+    expected = { 'B' => 9, 'I' => 1, 'L' => 0 }
+    assert_equal expected, Alphametics.solve(input)
```
Insti: Was: `Alphametics.new.solve(input)`
Now: `Alphametics.solve(input)`
Instantiating an object is unnecessary, so it now uses an instance method.

exercises/alphametics/example.tt
```diff
@@ -3,16 +3,37 @@ gem 'minitest', '>= 5.0.0'
...
-    assert_equal(expect, actual)
+    input = <%= test_case.input %>
+    expected = <%= test_case.expect %>
+    <%= test_case.workload %>
   end
 <% end %>
+  # These tests have been commented out due their long runtime.  If you are
+  # interested in optimsing your solution for speed these are a good tests to
+  # try.
+  #
```
Insti: The 'long' tests are ignored by the test case generator but are hardcoded here for reference. 
It would also be possible (but more complicated) to generate them (I think I forgot that's what I started out trying to do.)

bmulvihill:
This looks/works good to me!

kotp:
@moveson will be helping with this, if I am reading the comments on his solution correctly.

Insti:
The reason they were commented out is because it was taking the example solver too long to solve them during CI checks.

Once @moveson contributes his example.rb we will no longer need to comment out the "slow" tests since his solver is fast enough to solve them quickly.

Should they still remain commented out in order to help the people who are trying to solve the problem?

By including them, we make it a much harder problem by implying that you also need to solve the 8 and 10 letter cases, but by having them commented out we have a simpler problem but with some easily accessible extensions to suggest in the solution reviews on the site.

My vote would be to keep them commented out.

kotp:
Place them at the end, so that by the time they are commenting out the `skip` it is apparent if they have a slow solution or not.  Place a comment there near the skip that states perhaps they can submit their current solution, without the following tests having ran.

bmulvihill:
I like @kotp's solution.  It keeps the template the same as the others, and it follows the simple pattern of uncommenting the skip to open a test up.

exercises/alphametics/alphametics_test.rb
moveson: In the final test, `input = 'AND + A + STRONG + OFFENSE + AS + A + GOOD = DEFENSE'`, the `=` should be `==`. Also, in the commentary, `optimsing` should be `optimising`.

Insti:
Waiting on @moveson to submit the improved `example.rb`

moveson:
The new `example.rb` is ready. What's the next step?

Insti:
@moveson create a new pull request that only updates the `example.rb` file.
Once that's there I can do what needs to be done to merge it into this one.

moveson:
This is ready when you are. See PR #477.

Insti:
I still need to merge in @moveson's `example.rb` and remove the test case filtering from the *_cases.rb file. I will do this when I get some time over the next few days.

Insti:
I've pulled in @moveson's solver and re-added the slow testcases with comments warning that they might be slow.

@bmulvihill can you have a look over it and see if it all looks ok now?

exercises/alphametics/alphametics_test.rb
```diff
@@ -3,64 +3,75 @@
...
+    skip
+    input = 'HE + SEES + THE == LIGHT'
+    expected = { 'E' => 4, 'G' => 2, 'H' => 5, 'I' => 0,
+                 'L' => 1, 'S' => 9, 'T' => 7 }
+    assert_equal expected, Alphametics.solve(input)
+  end
+
+  # The obvious algorithm can take a long time to solve this puzzle,
+  # but an optimised solution can solve it fairly quickly.
+  # (It's OK to submit your solution without getting this test to pass.)
```
Insti: Is the wording of this comment OK?

kotp: Yes.

PR #471, opened 2016-10-25
[isogram] Eliminate trailing whitespace in test file.

Insti:
When the first test case is generated it leaves out the 'skip' But the
line still contains the whitespace present in the template file.

Add commented out skip to avoid having trailing whitespace.

I'm not sure this is an improvement, but it does provide the solver with
a hint about what they're supposed to do with the skips to make the
tests run.

### DIFF
--- a/lib/isogram_cases.rb
+++ b/lib/isogram_cases.rb
@@ -13,7 +13,7 @@ class IsogramCase < OpenStruct
   end
 
   def skip
-    'skip' unless index.zero?
+    index.zero? ? '# skip' : 'skip'
   end
 end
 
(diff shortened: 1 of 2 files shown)


bmulvihill:
👍

kotp:
Yes, I prefer the commented skip, I think it was what I did in the Hamming generator.  That way if someone has been away from Exercism for a while, it serves as a reminder as well.

lib/isogram_cases.rb
```diff
@@ -13,7 +13,7 @@ def assertion
   end
 
   def skip
-    'skip' unless index.zero?
+    index.zero? ? '# skip' : 'skip'
```
kotp: While we are here, we may as well make this `skipped` to match other implementations.  Keeping things as "standardized" as possible.  (Ref: Hamming)

Insti:
I *was* planning on standardising this, but that can easily done in another PR or by someone else. :)

kotp:
Sorry, was looking good to come in.  :blush:

Insti:
No problem, my expectation that everyone else would psychically know my intentions was probably unrealistic. :laughing:

PR #473, opened 2016-10-26
Adds Queen Attack test cases from x-common

samjonester:


Insti:
Great, thanks.
Did you manually add the missing test cases?

samjonester:
I did. Would you like me to create a generator instead? This exercise has tests that aren't a part of x-common, related to presentation of the board, which would complicate the generator.

Insti:
Updating the test file is a good start but if you would like to create a generator you are very welcome to, we need to do it at some point, it may as well be now.

The generator should only use the tests that are part of the x-common data. Any tests that are not part of x-common can be ignored.
If you think they are important and need to be included you can make a PR to add them to the canonical-data.json in x-common.

samjonester:
Cool! I'll make the generator with just the test cases from x-common. I think the extra tests confuse the purpose of the exercise anyways.

Insti:
The documentation on how to create generators is a little out of date, so have a look at the recent Transpose test case generator  https://github.com/exercism/xruby/pull/466 for an example of what to do.

The generator should have methods called: `skipped`, `test_name`, `workload`

Insti:
Thanks @samjonester :heart:,  I've merged in these changes so the test file is up to date.
You can create a new PR when you have made the generator.

PR #477, opened 2016-10-31
Fast alphametics solver that uses partials to limit the set of viable…

moveson:
… permutations.

@Insti Please try again merging with PR #469. Should not have conflicts this time.


moveson:
Checks should pass under V4 tests. With this solver you can add a 10-digit test, if you like. Try: `FIFTY + STATES == AMERICA`.

Insti:
Pulled into #469 
Thanks @moveson :heart:

PR #478, opened 2016-11-01
Update README with Exercise Generators

bmulvihill:
Related to: https://github.com/exercism/xruby/issues/452 and https://github.com/exercism/xruby/pull/463

Updates the README to include information about the generator interface.  This way the README has up to date information before https://github.com/exercism/xruby/pull/463 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,9 +114,9 @@ If you wish to create a new generator, or edit an existing one, the generators c
 
 All generators currently adhere to a common public interface, and must define the following three methods:
 
-- `test_name` - Output the name of the test
-- `workload` - Output the body of the test
-- `skipped` - Output skip syntax
+- `test_name` - Returns the name of the test (i.e `test_one_equals_one`)
+- `workload` - Returns the main syntax for the test.  This will vary depending on the test generator and its underlying implementation
+- `skipped` - Returns skip syntax (i.e. `skip` or `# skip`)
 
 ## Pull Requests
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,16 @@ then additional inputs/outputs should be submitted to the x-common repository.
 Changes to the test suite (style, boilerplate, etc) will probably have to be made to
 `example.tt`.
 
+### Exercise Generators
+
+If you wish to create a new generator, or edit an existing one, the generators currently live in the lib directory and are named `$PROBLEM_cases.rb`.  For example, the hamming generator is `lib/hamming_cases.rb`.  
+
+All generators currently adhere to a common public interface, and must define the following three methods:
+
+- `test_name` - Output the name of the test
+- `workload` - Output the body of the test
+- `skipped` - Output skip syntax
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,6 +108,16 @@ then additional inputs/outputs should be submitted to the x-common repository.
 Changes to the test suite (style, boilerplate, etc) will probably have to be made to
```
Insti: 'track specific' might be redundant since this is only talking about the xruby track anyway.

README.md
```diff
@@ -108,6 +108,16 @@ then additional inputs/outputs should be submitted to the x-common repository.
...
 `example.tt`.
 
+### Exercise Generators
+
+If you wish to create a new generator, or edit an existing one, the generators currently live in the lib directory and are named `$PROBLEM_cases.rb`.  For example, the hamming generator is `lib/hamming_cases.rb`.  
+
+All generators currently adhere to a common public interface, and must define the following three methods:
+
+- `test_name` - Output the name of the test
+- `workload` - Output the body of the test
```
Insti: Replace 'Output' with 'Return' 
'Output' is ambiguous, since we don't want to be `puts`ing things.

Insti: I think there's still some discussion to be had about what workload should be returning.

I don't think 'the body of the test' is what it should be doing.

bmulvihill: Any thoughts about what it should do?  Just return an assertion?  If we are just returning an assertion than we might need to consider defining additional methods (i.e. `input` and `expected`.  I think it boils down to what belongs in the template vs the cases class.

Some recent generators (tournament for example) are defining an input and expect variable in the view, but the workload method is relying on the input variable being defined in the view, which does not seem ideal to me.  

Keeping all the presentation confined to the workload method allows you to keep everything together at the same level of abstraction.  When I originally thought about the interface I was thinking the Cases classes would be almost like a presenter and workload would just spit out the test body.  Since every exercise needs to define a test body, it is really just a matter of implementing its presentation.

Anyway these are just some of my thoughts, let me know what you think.

kotp: The template should be as simple as possible.  It should contain as little "logic" as is feasible.  This was the driving factor for skipped.

bmulvihill: @Insti I would like to see it in action, it sounds interesting.

Insti: @bmulvihill here is what I have so far: https://github.com/Insti/xruby/pull/1

I'm working on this in my own fork until it is ready for merging into the main branch here. You should be able to pull the changes from my fork into your own if you want to test them out or submit PRs to me.

Everyone is welcome to comment/contribute.

bmulvihill: @Insti Awesome I'll pull it down and take a look

bmulvihill:
@Insti made some small updates based on your review.  I think this should work until your changes are made to the generator.

Insti:
Great, thanks. It's good to document the current process even if there is a new one coming (slowly).

PR #479, opened 2016-11-02
lib/generator.rb: Ignore blank template lines.

Insti:
If you have a template that contains a value that might be nil or an empty string such as:
```
<%= test_case.comment %>
```
You often do not want a blank line appear in the output if there is no comment.

This patch changes the ERB setttings so that these blank lines are not included in the output. 

This may cause minor issues where this behaviour has been relied on such as templates that expect a Ruby syntax related 'end' to insert a blank line.
```
<% end %>
```
But these are easily fixed by inserting a new blank line into the template.

Technical details:

It does this by setting the ERB *trim_mode* to '<>'
'<>' omits newline for lines starting with <% and ending in %>

See also: http://ruby-doc.org/stdlib-2.1.1/libdoc/erb/rdoc/ERB.html#method-c-new

### DIFF
--- a/lib/generator.rb
+++ b/lib/generator.rb
@@ -64,7 +64,8 @@ class Generator
 
   def generate_test_file
     File.open(path_to("#{name.gsub(/[ -]/, '_')}_test.rb"), 'w') do |f|
-      f.write ERB.new(File.read(path_to('example.tt'))).result binding
+      template = File.read(path_to('example.tt'))
+      f.write ERB.new(template, nil, '<>').result binding
     end
   end
 


Insti:
This will require most `example.tt` files to be tweaked but I'm hoping these can be tackled on a case-by-case basis the next time that specific generator is run. Unless someone else wants to go through and do them all at once.

The tweak will be to add a new blank line after:
`<% end %>`
and 
`<%= IO.read(XRUBY_LIB + '/bookkeeping.md') %>`

kotp:
I considered the same issues when I did Hamming, addressing the `skip` vs `skipped`.  How prevalent is this need?  It changes the default behavior of the tooling I think.

Insti:
> How prevalent is this need?

"Very", I think. Every problem will need skipped or not skipped and this is a more elegant way to handle that. As well as cases with optional comments which will be possible to do now.

> It changes the default behavior of the tooling I think.

Yes it does, but for the better, and the implications for any that do not need this functionality are extremely minor. 
```
<% end %>
```
becomes
```
<% end %>

```
(1 extra newline character)

PR #482, opened 2016-11-05
Replace throw with raise

kotp:
Standardize the skip while we are making this change


exercises/leap/.version
```diff
@@ -1 +1 @@
-2
\ No newline at end of file
+3
```
Insti: I don't think this warrants a version bump.
None of the user facing test cases changed.

Insti:
Thanks again @kotp :heart:

PR #424, opened 2016-08-28
[WIP] add dominoes

seanreed1111:
working on dominoes

PR #484, opened 2016-11-08
dominoes: Add generator and example

petertseng:
With my apologies to #424, but @seanreed1111 claimed it was OK to let someone else take it.

This PR complies with the new names in #463.

---

Note that this exercise appears to differ from other xruby exercises in
the following significant manner:

Verification that the student function under test has produced a correct
answer is NOT done by comparing the observed output to an expected
output (or even a list of expected outputs), because the set of
acceptable outputs is quite large even for a small input set.

Thus, verification is done by running the observed output against a
function that verifies various properties about the output.

https://github.com/exercism/x-common/blob/master/exercises/dominoes/canonical-data.json
explains the approach taken.


lib/dominoes_cases.rb
```diff
@@ -0,0 +1,23 @@
+class DominoesCase < OpenStruct
+  def test_name
+    'test_%s' % description.gsub(/['= -]+/, '_')
+  end
+
+  def workload
+    <<-WL.chomp
+input_dominoes = #{input}
+    output_chain = Dominoes.chain(input_dominoes)
+    #{can_chain ? 'assert_correct_chain(input_dominoes, output_chain)' : 'assert_equal nil, output_chain, "There should be no chain for #{input_dominoes}"'}
```
petertseng: Line is uncomfortably long. Any way to shorten?

Note that it is not so long in the generated output, since only one of two alternatives is taken.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,138 @@
...
+    input_dominoes = [[1, 1]]
+    output_chain = Dominoes.chain(input_dominoes)
+    assert_correct_chain(input_dominoes, output_chain)
+  end
+
+  def test_singleton_that_can_t_be_chained
+    skip
+    input_dominoes = [[1, 2]]
+    output_chain = Dominoes.chain(input_dominoes)
+    assert_equal nil, output_chain, "There should be no chain for #{input_dominoes}"
```
petertseng: try assert_nil

exercises/dominoes/example.rb
```diff
@@ -0,0 +1,37 @@
...
+      elsif domino_right == chain_right
+        # Try to add domino (flipped) to chain
+        if (subchain = try_subchain(other_dominoes, chain_left, domino_left))
+          return [[domino_right, domino_left]] + subchain
+        end
+      end
+    }
+
+    # Found no suitable chain.
+    nil
```
petertseng: unfortunately, it's going to have to be `nil` and not `[]`.

Why? Because for the input `Dominoes.chain([])`, `[]` is a valid answer. Therefore, returning `[]` when you have no chain will mean that the answer `[]` is ambiguous.

petertseng:
Oh yeah, I was going to mention I did this to test out the instructions for writing a generator. Seems to work OK. I'd mention that you need to touch `.version` or the generator won't run (`increment_version` requires that such a file exists). And you need to have an example file too. I tried to generate my tests before working on my example, but `increment_version_in_example` requires that the example actually exist.

exercises/dominoes/example.tt
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env ruby
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'dominoes'
+
+# Test data version:
+# <%= sha1 %>
+class DominoesTest < Minitest::Test
+  def assert_correct_chain(input_dominoes, output_chain)
```
Insti: Could you move this to the bottom of the class please.

In real life I'd put it at the top, but here it would be good if the first thing the student saw was the first test they need to implement.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,138 @@
...
+# Test data version:
+# 018c48d
+class DominoesTest < Minitest::Test
+  def assert_correct_chain(input_dominoes, output_chain)
+    refute_nil output_chain, "There should be a chain for #{input_dominoes}"
+
+    input_normal = input_dominoes.map(&:sort).sort
+    output_normal = output_chain.map(&:sort).sort
+    assert_equal input_normal, output_normal,
+      'Dominoes used in the output must be the same as the ones given in the input'
```
Insti: I'd extract this test.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    output_chain.each_cons(2).with_index { |(d1, d2), i|
+      _, d1_right = d1
+      d2_left, _ = d2
+      assert_equal d1_right, d2_left,
```
Insti: Could you use `.first` and `.last` here instead?

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,138 @@
...
+        "In chain #{output_chain}, right end of domino #{i} (#{d1}) and left end of domino #{i + 1} (#{d2}) must match"
+    }
+
+    return if output_chain.empty?
```
Insti: Move this up above line 17. If we're short-circuiting short circuit as early as possible.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,138 @@
...
+    output_normal = output_chain.map(&:sort).sort
+    assert_equal input_normal, output_normal,
+      'Dominoes used in the output must be the same as the ones given in the input'
+
+    output_chain.each_cons(2).with_index { |(d1, d2), i|
+      _, d1_right = d1
+      d2_left, _ = d2
+      assert_equal d1_right, d2_left,
+        "In chain #{output_chain}, right end of domino #{i} (#{d1}) and left end of domino #{i + 1} (#{d2}) must match"
+    }
```
Insti: I'd extract this too.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    return if output_chain.empty?
+
+    first_domino = output_chain.first
+    last_domino = output_chain.last
+    d1_left, _ = first_domino
+    _, d2_right = last_domino
+    assert_equal d1_left, d2_right,
+      "In chain #{output_chain}, left end of first domino (#{first_domino}) and right end of last domino (#{last_domino}) must match"
```
Insti: Again extract.

Insti:
> I did this to test out the instructions for writing a generator. 
Great, thanks. (We're working on improving/simplifying the generator creation process.)

Insti:
Thanks for doing this @petertseng 
& Great commit message :heart:

Insti:
Missclicked close :crying_cat_face:

exercises/dominoes/example.rb
petertseng: Sure. I'll note that that comment was partially targeting https://github.com/exercism/xruby/pull/469#discussion_r84790337 to explain why in this case it cannot be that this method always returns the same object - it has to sometimes return nil instead of list. It's useful to have the comment somewhere, so `example` it will be. Putting it in the test file didn't seem right since it wasn't relevant to all students.

petertseng:
TODO list for me so I understand what feedback I've addressed.

- [ ] explain in code comment why `nil` and not `[]` is the appropriate answer for "no chain possible"
- [ ] move `assert_correct_chain` to bottom
- [ ] extract "same dominoes" check.
- [ ] extract "consecutive dominoes" check.
- [ ] extract "dominoes at the ends" check.
- [ ] in `assert_correct_chain`, short-circuit before the `each_cons` check.
- [ ] first and last instead of `_, d1_right`

exercises/dominoes/dominoes_test.rb
petertseng: I am assuming since made it a separate comment that the "consecutive" test and the "dominoes at the ends" should be in two separate methods instead of the same. I'll go with that for now, it's easy to bring them together if I was mistaken in my interpretation.

petertseng:
Review note: These commits are meant to be squashed. The only reason they are currently separate is for ease of understanding what chaged.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  def test_can_reverse_dominoes
+    skip
+    input_dominoes = [[1, 2], [1, 3], [2, 3]]
+    output_chain = Dominoes.chain(input_dominoes)
+    assert_correct_chain(input_dominoes, output_chain)
+  end
+
+  def test_can_t_be_chained
```
petertseng: slightly ugly name. Should it be `cant` instead? Can be done without too much pain (`.tr("'", '')` the test name instead of replace with underscore)`

lib/dominoes_cases.rb
```diff
@@ -1,6 +1,6 @@
 class DominoesCase < OpenStruct
   def test_name
-    'test_%s' % description.gsub(/['= -]+/, '_')
+    'test_%s' % description.tr("'", '').gsub(/[= -]+/, '_')
```
Insti: or even better: `gsub(/can't/,'can not')`

petertseng: I like it, suggestion taken.

Insti:
:tada: Thanks @petertseng 
I wish I could leave reaction icons on pull requests.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,142 @@
...
+    input_dominoes = [[1, 1]]
+    output_chain = Dominoes.chain(input_dominoes)
+    assert_correct_chain(input_dominoes, output_chain)
+  end
+
+  def test_singleton_that_can_not_be_chained
+    skip
+    input_dominoes = [[1, 2]]
+    output_chain = Dominoes.chain(input_dominoes)
+    assert_nil output_chain, "There should be no chain for #{input_dominoes}"
```
Insti: What do you think about making a helper method `refute_correct_chain`?
I'm not entirely sure it's a good idea, but I like the symmetry of all the tests using similar terminology.

petertseng: From the point of view of the student looking at this test generator: There is a lot of repetitive `"There should be no chain for #{input_dominoes}"`. Moving it into `refute_correct_chain` removes this repetition. So that's one point in favour of it. One point against is that it'd be very simple, but that's very minor.

Here is another proposal that might also go along with `refute_correct_chain`:

Should `output_chain = Dominoes.chain(input_dominoes)` go inside `assert_correct_chain` and `refute_correct_chain`? Again it would reduce duplication, and then those two methods would only have to take in the `input_dominoes`. Though this might make it harder for students to understand what they're supposed to implement  (they have to look for the definitions of `{assert,refute}_correct_chain`).

So currently: I am thinking yes to `refute_correct_chain` and no to "move `output_chain = Dominoes.chain(input_dominoes) into `{assert,refute}_correct_chain`"

Insti: > I am thinking yes to refute_correct_chain and no to "move output_chain = Dominoes.chain(input_dominoes) into{assert,refute}_correct_chain`"

I agree with this.
It is important not to hide the subject of the test.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env ruby
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'dominoes'
+
+# Test data version:
+# 82eb00d
```
Insti: Another minor minor thing, I've started putting these on 1 line.

lib/dominoes_cases.rb
```diff
@@ -0,0 +1,23 @@
+class DominoesCase < OpenStruct
+  def test_name
+    'test_%s' % description.gsub("can't", 'can not').gsub(/[= -]+/, '_')
+  end
+
+  def workload
+    <<-WL.chomp
+input_dominoes = #{input}
+    output_chain = Dominoes.chain(input_dominoes)
+    #{can_chain ? 'assert_correct_chain(input_dominoes, output_chain)' : 'assert_nil output_chain, "There should be no chain for #{input_dominoes}"'}
```
petertseng: continuing a conversation in https://github.com/exercism/xruby/pull/484#discussion_r87055734 about whether we should have `refute_correct_chain`

Yet another argument for:

There's a bit of asymmetry in this line.

There's a slight bit more logic in `assert_nil output_chain, "There should be no chain for #{input_dominoes}"` versus `assert_correct_chain(input_dominoes, output_chain)` where all the logic is offloaded to the `example.tt` file. I think it makes more sense to be consistent, so I have added `refute_correct_chain`. If we change our minds about wanting it we just remove the commit.

Insti:
Once this is ready to go I'll try solving it\* and see how the tests work in practice.

\* Maybe not today.

exercises/dominoes/dominoes_test.rb
petertseng: Ah, okay. One line it is. I think putting it on a separate line was an artifact of the time when both the sha and the git commit summary was included... nowadays since it's just the sha it probably makes sense on one line.

exercises/dominoes/dominoes_test.rb
```diff
@@ -0,0 +1,145 @@
...
+  #   VERSION = 1 # Where the version number matches the one in the test.
+  # end
+  #
+  # If you are curious, read more about constants on RubyDoc:
+  # http://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/constants.html
+  def test_bookkeeping
+    skip
+    assert_equal 1, BookKeeping::VERSION
+  end
+
```
Insti: Would it be worth inserting the comment about why we can't just use input/output tests in between the bookkeeping and the custom asserts?

petertseng: Seems reasonable. Added.

petertseng:
:+1: 

Note! Starting in three days from now my availability will be limited. Of course, if everything is fine, I am sure the squash button will do just fine, I think it'll only be necessary to keep the first commit message.
If something needs my attention after that time, be prepared for responses that are slower than they were today.

PR #489, opened 2016-11-24
accumulate: add bookkeeping

raadler:
I'm not sure if this would be the first version of the tests, so if anyone knows differently please let me know and I can edit it.


Insti:
Thanks @raadler , If it doesn't have a BookKeeping module then we tend to consider it the first version anyway.

PR #497, opened 2016-12-01
Templates for Issues and Pull Requests

kotp:
This will provide a template for when folks open an issue or pull request.

This is based on the generator as introduced at [GitHub Issue and Pull Request Templates: Choose Your Own Adventure](https://wptavern.com/github-issue-and-pull-request-templates-choose-your-own-adventure) blog.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,43 @@
+<!--- Provide a general summary of the issue in the Title above -->
+
+## Expected Behavior
+<!--- If you're describing a bug, tell us what should happen -->
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
+
+## Current Behavior
+<!--- If describing a bug, tell us what happens instead of the expected behavior -->
+<!--- If suggesting a change/improvement, explain the difference from current behavior -->
+
+## Possible Solution
+<!--- Not obligatory, but suggest a fix/reason for the bug, -->
+<!--- or ideas how to implement the addition or change -->
+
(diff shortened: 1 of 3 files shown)


kotp:
These will work once they are on master branch.  So seeing it live is possible.  But it must be on master of your fork, if you decide to see how this works.

Insti:
I'm worried that these are going to lead to a lot of useless paperwork.
I can see how something like this could be good for a project with 100s of contributors, but that's not us right now.

What are these trying to solve?
We do not currently have a problem with too many, or bad, issues/PRs.

kotp:
I was hoping that it could be modified as needed for different aspects.  General use, 24 pull requests, hacktoberfest, or however it may want to be used.  Remember that the comments are just comments, not shown in the issue or pr, but perhaps just a reminder.  Depends on what we end up with.  I know I often forget to reference (though that is more me ignoring my local template) issues that are related.

kotp:
Oh, I should mention, the reason for doing this here is that it impacts less of the Exercism maintaining community.  If we find it helpful, then it is simple to put into place in other areas of github Exercism.

tommyschaefer:
One small side-effect of these changes for pull requests:

image

Because the template has checkboxes, GitHub shows the progress bar on PRs. I don't have a huge opinion on this, but is a little bit confusing at first glance.

kotp:
Are there instructions to delete the lines that aren't needed?  Are there things you like about it, think should change?  Please do.  They are supposed to be here to cover "most cases" if we can generalize, and have it help with title expectations, etc.

PR #499, opened 2016-12-02
binary: Reconsiders responsibility of constructor

samjonester:
The constructor in the binary exercise was performing validation,
computations, and managing the state of the Binary object. A string
doesn't necessarily seem like the best representation of state for this
object, and providing it as an argument to initialize makes this class
inflexible to future changes.

This change will eliminate calculations happening in the constructor,
encouraging better maintainability to those learning idiomatic Ruby.


kotp:
I approve of the change.  The only downside to this is that Binary is going away for [All  Your Base](https://github.com/exercism/xruby/tree/master/exercises/all-your-base) exercise... What do you think @samjonester can you review that, and see if the same kind of change can/should be made there?

Insti:
Looks good, that redundant constructor annoyed me too.

samjonester:
I'd be glad to! I'll report back after taking a look.

PR #503, opened 2016-12-15
Start unit testing lib/generator

Insti:
Adds unit tests for `lib/generator` In order to make development on the test generator easier and less scary.

The tests can be run via: `bundle exec rake test`

In order to get this set up a few other things need to happen:

Add `rake` gem for easy test running.
Add `simplecov` gem to provide test coverage reports.
Create `test/` directory to hold the tests.
Exclude `test/` directory from executable tests check.
Create `test/generator_test.rb` to hold unit tests of `lib/generator.rb`

Once that was done I was able to start writing some unit tests. 

I have added a configurable `metadata_repostitory_path` in order to aid testing by being able to supply a test fixture rather than having to rely on the global x-common data.

Not much else is covered so far this shall be improved in future PRs.

This pull request is ready to merge `bin/generator` will still work (as long as it is run from the xruby root directory.)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source 'https://rubygems.org'
 
 gem 'minitest'
 gem 'rubocop', '0.36.0'
+gem 'simplecov'
+gem 'rake'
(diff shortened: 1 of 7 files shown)


kotp:
Thanks @Insti looks good.

PR #509, opened 2016-12-19
wordy: Generate tests

tommyschaefer:
Hi there!

I've made a first pass at converting the `wordy` exercise to use generated tests (#396). 

I based the updates on #474, but I still have a few questions:

1. The original tests use a mixture of the following forms:

```
question = '...'
assert_equal ...
```
and
```
assert_equal 1, WordProblem.new('...').answer
```

I'm assuming this has to do with not going past 80 characters per line. In my first pass I decided to just use the first form for all of the tests. Is this okay, or should I put some work into switching between the two forms based on the length of the assertion?

2. Does anything special have to happen since this wasn't originally book-kept?

3. In the `test_add_then_multiply` / `test_addition_and_multiplication` test, the xruby implementation includes a custom error message. Should this error message be removed (as it is currently), or should we add the custom error message to the shared metadata and generate based on that?

**Thank you so much for your time :smile: !**

PR #510, opened 2016-12-20
WIP: all-your-base: add test generator

tommyschaefer:
Hi there!

Checking off exercises from #396.

I've made some good progress converting `all-your-base` to use generated tests, but I've run into a bit of a snag implementing the 13 undefined tests from the shared data.

I've opened exercism/x-common#473 to see if this has been solved before, or if some new information will have to be added to the shared test data.

In the future, would it be preferable for me to wait to make a PR until the changes are fully ready for review? I was thinking it may be better to just go ahead and make the PR so that people could see it had some work being done on it, but if thats not the case I won't do it in the future 😄!


Insti:
Works in progress are definitely encouraged.

Insti:
You'll need to pre-process the canonical data and add the appropriate track level details.

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -131,9 +135,9 @@ def test_empty_list
   def test_single_zero
     skip
     digits = [0]
-    input_base  = 10
+    input_base = 10
     output_base = 2
-    expected = [0]
+    expected = WHAT TO DO?
```
Insti: expected = [0]

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -142,12 +146,12 @@ def test_single_zero
...
 
-  def test_multiple_zeroes
+  def test_multiple_zeros
     skip
     digits = [0, 0, 0]
-    input_base  = 10
+    input_base = 10
     output_base = 2
-    expected = [0]
+    expected = WHAT TO DO?
```
Insti: expected = [0]

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -175,7 +179,7 @@ def test_negative_digit
     digits = [1, -1, 1, 0, 1, 0]
     input_base = 2
     output_base = 10
-    expected = nil
+    expected = WHAT TO DO?
```
Insti: raise error about invalid input

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -161,7 +165,7 @@ def test_leading_zeros
     digits = [0, 6, 0]
     input_base = 7
     output_base = 10
-    expected = [4, 2]
+    expected = WHAT TO DO?
```
Insti: expected = whatever the answer to [6,0] is. 
I can't do base 7 in my head :(

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -231,7 +235,7 @@ def test_first_base_is_zero
     digits = []
     input_base = 0
     output_base = 10
-    expected = nil
+    expected = WHAT TO DO?
```
Insti: raise error about invalid base.

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -245,7 +249,7 @@ def test_second_base_is_zero
     digits = [7]
     input_base = 10
     output_base = 0
-    expected = nil
+    expected = WHAT TO DO?
```
Insti: raise error about invalid base.

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -259,7 +263,7 @@ def test_first_base_is_negative
     digits = [1]
     input_base = -2
     output_base = 10
-    expected = nil
+    expected = WHAT TO DO?
```
Insti: raise error about invalid base.

exercises/all-your-base/all_your_base_test.rb
tommyschaefer: Currently we just expect the output to be nil. Should the cases with invalid data be changed to expect an error to be raised?

exercises/all-your-base/all_your_base_test.rb
tommyschaefer: After looking at it with some fresh eyes, I think this is the only example that I'll have trouble implementing.

For the other examples I can check whether their input and bases are valid and generate a test on that. Also the empty / 0 input cases shouldn't be hard to catch either.

Because we ignore leading zeros in the input, the expected output for the given inputs is not included in the shared test data. I'm not sure how to handle this case in a way that would allow the input or bases to change without new generations breaking. Do you have any ideas on this one?

Thank you so much for your help and time!

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -1,9 +1,10 @@
-gem 'minitest', '>= 5.0.0'
```
tommyschaefer: Just wanted to double check that it's okay to remove this based on your comment on #509? Thanks!

tommyschaefer:
Hmm. The reason the checks are failing is because, without specifying the `TargetRubyVersion` in the `.rubocop.yml` file, it doesn't like keyword arguments.

Adding:
```yml
AllCops:
    TargetRubyVersion: 2.1
```

solves this problem.

Should I open an issue / PR for discussing that change, or should I include it in this PR?

Insti:
>  specifying the TargetRubyVersion in the .rubocop.yml

Doing this as a separate PR would be great, thanks.

exercises/all-your-base/all_your_base_test.rb
Insti: yes. remove this line.

lib/all_your_base_cases.rb
```diff
@@ -1,4 +1,8 @@
 class AllYourBaseCase < OpenStruct
+  def self.build(row:, index:)
+    new(PreProcessor.new(row: row, index: index).run)
```
Insti: You can get rid of this method completely by just calling the preprocessor directly on line 89

lib/all_your_base_cases.rb
```diff
@@ -37,8 +41,51 @@ def error_message
...
+  def run
+    row.merge('expected' => expected_value, 'index' => index)
+  end
+
+  private
+
+  attr_reader :row, :index
+
+  def expected_value
+    return row['expected'] unless row['expected'].nil?
```
Insti: This is headed in the right direction.

An alternate method would be to special case each one based on the description (or the input values.)

tommyschaefer:
Just realized a couple comments were in outdated diffs, so I'll repost them here for visibility. One was:

> Currently we just expect the output to be nil. Should the cases with invalid data be changed to expect an error to be raised?

The second was regarding the test for leading zeros:

> Because we ignore leading zeros in the input, the expected output for the given inputs is not included in the shared test data. I'm not sure how to handle this case in a way that would allow the input or bases to change without new generations breaking. Do you have any ideas on this one? Is it okay to add a "hard-coded" special case for this?

lib/all_your_base_cases.rb
tommyschaefer: Which do you prefer?

Personally, I like the fact that it's resilient to the test data changing how it is, but I can go both ways. It does feel a bit strange that some of the exercise logic is reproduced in the generator, though.

tommyschaefer:
Hmm... sorry. I unfortunately messed up the rebase :arrow_up: Should be fixed now though.

lib/all_your_base_cases.rb
kotp: `return row['expected'] if row['expected']`

Insti:
> Currently we just expect the output to be nil. Should the cases with invalid data be changed to expect an error to be raised?

Yes these should raise an error.
(See `bowling` for some clues about this)

> Because we ignore leading zeros in the input, the expected output for the given inputs is not included in the shared test data. I'm not sure how to handle this case in a way that would allow the input or bases to change without new generations breaking. Do you have any ideas on this one? Is it okay to add a "hard-coded" special case for this?

Hardcoding values in the pre-processor is acceptable. Changes to the canonical-data are allowed to break the test generation.

exercises/all-your-base/example.rb
```diff
@@ -1,17 +1,22 @@
 module BookKeeping
-  VERSION = 1
+  VERSION = 2
```
tommyschaefer: Bumped test version since we now expect invalid base / input to raise an error rather than return nil

exercises/all-your-base/example.rb
```diff
@@ -1,17 +1,22 @@
...
 end
 
 class BaseConverter
   def self.convert(base_from, number_array, base_to)
-    return if number_array.any?{|number| number < 0 || number >= base_from}
-    return if base_from <= 1 || base_to <= 1
+    fail ArgumentError if invalid_inputs?(base_from, number_array, base_to)
```
tommyschaefer: I tried to change as little as possible here to get tests to pass. Does this look okay?

Insti: that's fine.

Insti:
This looks good @tommyschaefer.
Once @kotp or @bmulvihill have reviewed it one of them can merge it in.

tommyschaefer:
Great! Thanks for all of your help! Should I go ahead and squash the commits now, or wait for some more feedback from them before doing that?

kotp:
@tommyschaefer you can squash them, and you can remove the WIP tag...

lib/all_your_base_cases.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  def handle_special_cases
+    [4,2] if row['input_digits'] == [0, 6, 0]
+  end
+end
+
+AllYourBaseCases = proc do |data|
+  JSON.parse(data)['cases'].map.with_index do |row, i|
+    AllYourBaseCase.new(
+      AllYourBaseCase::PreProcessor.new(row: row, index: i).run
```
Insti: Whenever you see a class with a single method called run it's a sign that it probably should be a class method rather than an instance method.

```
AllYourBaseCase::PreProcessor.call(row: row, index:i)
```

Insti: And the preprocessor doesn't even use the index parameter, so why pass it in?

tommyschaefer: > And the preprocessor doesn't even use the index parameter, so why pass it in?

Hmm. I guess I was thinking that adding the index to the data passed into `AllYourBaseCase` was part of preprocessing. The alternative would look like
```ruby
AllYourBaseCase.new(
  AllYourBaseCase::PreProcessor.call(row: row).merge(index: i),
)
```

Would this be preferable? I can see it either way.

Thank you for your feedback!

Insti: It doesn't matter too much *when* you merge it, but it is part of the `row` data.
I have a slight preference for doing it as late as possible, so your alternative above looks good.

kotp:
Why the change to class call rather than instance?  Just wondering.

tommyschaefer:
I made the change based on some feedback from @Insti on a now outdated diff (copied here for easier reading):

> Whenever you see a class with a single method called run it's a sign that it probably should be a class method rather than an instance method.

```ruby
AllYourBaseCase::PreProcessor.call(row: row, index:i)
```

Though after going through a few different options of implementing it, I'm not entirely sure that the preprocessor is better for the change. It certainly could be that I just haven't stumbled upon the correct way to do it yet though.

Do you all have thoughts on how it can be improved? I'll keep working at it as well. 😄

Thank you all so much for your time and feedback! I really appreciate all of the help I've gotten!

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -175,129 +176,118 @@ def test_negative_digit
...
+  # module BookKeeping
+  #   VERSION = 1 # Where the version number matches the one in the test.
+  # end
+  #
+  # If you are curious, read more about constants on RubyDoc:
+  # http://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/constants.html
   def test_bookkeeping
     skip
-    assert_equal 1, BookKeeping::VERSION
+    assert_equal 2, BookKeeping::VERSION
```
kotp: This does not need to be incremented, as it is not live yet.

exercises/all-your-base/example.rb
kotp: I did not see this in master, do we need to bump the version?

kotp:
Thanks @tommyschaefer I did end up seeing it in my e-mail.

I think we could bring in the changes.  The class or module choice is not that important at the moment, I don't think.

lib/all_your_base_cases.rb
```diff
@@ -20,9 +20,7 @@ def skipped
   private
 
   def indent(size, lines)
-    lines.split("\n").map do |line|
-      ' ' * size + line
-    end.join("\n")
+    lines.lines.each_with_object('') { |line, obj| obj << ' ' * size + line }
```
Insti: `lines.lines` looks weird.
If you need to call `lines` on it that suggests to me that it's not lines at all, so either needs to be lines that get passed in or needs a different name.

tommyschaefer: Completely agreed.

Would `text` or `code` be better, maybe?

Insti: I have a slight preference for `text` but it doesn't make too much difference.

lib/all_your_base_cases.rb
```diff
@@ -0,0 +1,106 @@
...
+      '"Expected #{expected} but got #{converted}."',
+    ].join("\n")
+  end
+
+  def error_assertion
+    [
+      "",
+      indent(4, "assert_raises ArgumentError do"),
+      indent(6, "BaseConverter.convert(input_base, digits, output_base)"),
+      indent(4, "end"),
```
Insti: One thing that bugs me about this otherwise great solution\* is all the micro-management of indentation that is going on.
There must be a more elegant way of handling it.

\* not just in this method, but throughout the whole file.

Insti:
Is there anything more than needs to be done here?

(@tommyschaefer you should come to the Gitter chat room: https://gitter.im/exercism/xruby)

tommyschaefer:
As far as I know, this should be good to go unless there's any other changes you all would like to see?

Insti:
I noticed this in the travis log:
> /home/travis/build/exercism/xruby/lib/all_your_base_cases.rb:68: warning: private attribute?

tommyschaefer:
I guess one thing to mention is that this generator isn't tested. Is that something you would like to see happen before thinking about bringing these changes in?

tommyschaefer:
@Insti nice catch! I'll look into it 😄

lib/all_your_base_cases.rb
```diff
@@ -0,0 +1,108 @@
...
+  class << self
+    def call(row)
+      @row = row
+
+      row.merge('expected' => expected_value)
+    end
+
+    private
+
+    attr_reader :row
```
tommyschaefer: Hmm looks like the warning in Travis is coming from this line.

Insti:
> I guess one thing to mention is that this generator isn't tested. Is that something you would like to see happen before thinking about bringing these changes in?

You're welcome to add them if you want, but I'd be fine bringing it in without tests.

- Generators can be easily tested by comparing the output with the previous version.
- The whole generator process is still a bit "in flux" as we move towards a new, more integrated way of generating the tests files. Because of that everything is likely to change in the near future and the changes are likely to be directly comparable so the tests will need to get re-written anyway.

tommyschaefer:
> You're welcome to add them if you want, but I'd be fine bringing it in without tests.

Sounds good. We can move forward without them as far as I'm concerned. I'll circle back around to these once the new generator stuff gets added.

Is it okay if I add an issue to remember this?

Insti:
> Is it okay if I add an issue to remember this?

Yep. That's one of the things they're for.

lib/all_your_base_cases.rb
```diff
@@ -57,16 +57,17 @@ def error_message
...
+
     def call(row)
       @row = row
 
       row.merge('expected' => expected_value)
     end
 
+    private :row
     private
```
kotp: Why not:

```ruby
    private

    attr_reader :row
```

Move the private things past the `private` declaration...  Leaving `call` above that line.

kotp:
Not ready to merge yet...

lib/all_your_base_cases.rb
tommyschaefer: This is what we were chatting about on Gitter the other day.

In ruby  < 2.3.0 private reader methods made like that cause a warning. In 2.3 that warning is removed. Links to some info about this are in the last commit message.

Thanks for the feedback!

kotp: This is the last commit, and instead of moving the lines, one more line was created.
If you move the attr_reader to below the private line, there is no warning for 2.4 (in regards to that) that I am finding.

tommyschaefer: The warning was removed in ruby 2.3, so in all later versions it isn't there and you won't see it in 2.4. Since we use 2.1.2, it shows up in CI and whatnot.

I'm not sure if moving the reader under private but keeping the explicit private with method name will still make a warning. I'll give this a try when I'm home.

Sorry for the lack of code samples... aparently the iOS keyboard doesn't have a back tick key? 😞

tommyschaefer: Copying commit message here so it can be read inline:

This change is to remove a warning present in ruby versions
less than 2.3.0. A thread discussing this warning can be found
at https://bugs.ruby-lang.org/issues/10967. It is removed in
ruby/ruby@32a5a09.

tommyschaefer: Just confirmed that

```ruby
private :row
private

attr_reader :row
```

does not work:

```
undefined method `row' for class `Class' (NameError)
```

kotp: Right... remove the `private :row` above... it would not work because it doesn't exist yet, and would not be the thing to do.

tommyschaefer: Yeah, that's what I had originally but it's what causes the warning mentioned above. Ruby versions less than 2.3.0 have a warning when a private attribute is defined unless its made private explicitly with private :method_name

kotp: As it is a warning, and not functionally wrong, I would probably leave it at that.  This also means as time goes on and the version of Ruby moves forward, it will eventually take care of itself.

Insti: I'd prefer that the code not emit warnings. If that means temporarily doing some sub-optimal things to keep the interpreter happy then so be it.

kotp: Definitely prefer warning free.  How will we trigger when it should be "corrected" when the warning would naturally go away?  Not that it is an important change...

Insti: I don't think it needs an explicit trigger. Next time that code gets edited, whoever modifies it will think it looks wrong,  (unjustly) curse @tommyschaefer for being an idiot :stuck_out_tongue_winking_eye: and fix it.

Insti:
Thanks for all your work on this @tommyschaefer :heart:

PR #511, opened 2016-12-20
Add TargetRubyVersion for rubocop

tommyschaefer:
In the process of working on some test generators, I ran into a little issue with the current rubocop and using keyword arguments.

Here I just added a target ruby version the rubocop config. I set the target version to `2.1`, but that was only to make keyword arguments. I'm not sure if that's the actual version we want to target, or if it really matters.

Thanks so much for your time!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,6 @@
+AllCops:
+  TargetRubyVersion: 2.1
+
 # Type 'Lint' (48):
 Lint/AmbiguousOperator:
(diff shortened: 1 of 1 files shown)


Insti:
Thanks! :heart:

2.1 is fine for now, we can always change it if we find the need.

bmulvihill:
I think this makes sense since we are running CI against Ruby 2.1.  I think we should also update the README to say you need Ruby 2.1 +

Insti:
And maybe the Gemfile?

PR #512, opened 2016-12-20
Update README to use Ruby version 2.1

tommyschaefer:
Making another PR to update the README based on #511

The [Gemfile](https://github.com/exercism/xruby/blob/master/Gemfile) doesn't specify a Ruby version, but would you like me to add it there? If so, should I use `2.1.2`, which is what CI runs on?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Exercism Exercises in Ruby
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


bmulvihill:
Looks good to me!

kotp:
@exercism/ruby Realize that 2.1 is in Security Maintenance mode and will EOL (end of life) soon.

PR #514, opened 2016-12-24
grains: add test generator

tommyschaefer:
Hi there!

Checking off exercises from #396.

I tried to include tests for the generator this time around. Is this okay? If so, should I add a `cases` (or another name) test directory to group the tests? Or is it okay as it is?

I also had a couple other questions, but I'll add them as comments on specific lines.


exercises/grains/example.rb
```diff
@@ -1,9 +1,14 @@
+module BookKeeping
+  VERSION = 1
+end
+
 module Grains
   def self.square(number)
+    fail ArgumentError if number <= 0 || number > 64
```
tommyschaefer: I had to make an adjustment to the implementation because the shared test data includes some error cases that weren't in the original ruby implementation.

How does this work with book keeping when it wasn't there originally? Normally, I'm assuming we would bump the `BookKeeping::VERSION`, but this exercise wasn't using book keeping yet.

exercises/grains/example.rb
```diff
@@ -1,9 +1,14 @@
...
     2**(number - 1)
   end
 
   def self.total
-    square(65) - 1
+    2**64 - 1
```
tommyschaefer: Because of the change to the square method, calling `square(65)` would raise an `ArgumentError`, so I just made the smallest change possible to make the tests pass. Is this okay?

exercises/grains/grains_test.rb
```diff
@@ -1,45 +1,81 @@
...
+  def test_4
     skip
     assert_equal 8, Grains.square(4)
   end
 
-  def test_square_16
+  def test_16
     skip
-    assert_equal 32_768, Grains.square(16)
+    assert_equal 32768, Grains.square(16)
```
tommyschaefer: I elected to use the`32769` notation rather than `32_768` for simplicities sake. Is this okay, or should I update it to use the other notation?

exercises/grains/example.rb
Insti: The adjustment looks reasonable.

BookKeeping is related to the tests, so you might need to add it in so that the example.rb still passes the tests.

exercises/grains/example.rb
Insti: This is fine.

exercises/grains/grains_test.rb
Insti: This is fine.

Although supporting the underscore version should just require a simple helper method.

exercises/grains/grains_test.rb
```diff
@@ -1,45 +1,81 @@
...
     skip
-    assert_equal 2_147_483_648, Grains.square(32)
+    assert_equal 2147483648, Grains.square(32)
   end
 
-  def test_square_64
+  def test_64
     skip
-    assert_equal 9_223_372_036_854_775_808, Grains.square(64)
+    assert_equal 9223372036854775808, Grains.square(64)
```
Insti: When we get to numbers like this the underscore representation is REALLY helpful.

lib/grains_cases.rb
```diff
@@ -0,0 +1,71 @@
...
+  cases = data['square']['cases'].map.with_index do |row, i|
+    assertion = GrainsCase::SquareAssertion.new(
+      expected: row['expected'],
+      input: row['input'],
+    )
+
+    GrainsCase.new(row.merge('assertion' => assertion, 'index' => i))
+  end
+
+  assertion = GrainsCase::TotalAssertion.new(
```
Insti: Why does this look different to the `square` method code?

tommyschaefer: This is because of the format of the canonical data:
```json
{
  "square": {
    "cases": [
      { "description": "...", "input": 1, "expected": 1 },
      ...
    ]
  },
  "total": {
    "description": "...",
    "expected": "..."
  }
}
```

Insti:
Sorry my review pace has slowed down, the holidays seem to leave me with less time rather than more!. I've still got my eye on these PRs, I'm just trying to find some time to give them a final test/review before merging them.

tommyschaefer:
@Insti No problem at all! I totally understand 😄

lib/grains_cases.rb
```diff
@@ -0,0 +1,49 @@
...
+end
+
+GrainsCases = proc do |data|
+  data = JSON.parse(data)
+
+  cases = data['square']['cases'].map.with_index do |row, i|
+    GrainsCase::SquareMethod.new(row.merge('index' => i))
+  end
+
+  cases << GrainsCase::TotalMethod.new(data['total'])
```
Insti: This case doesn't know its `index`.

lib/grains_cases.rb
```diff
@@ -0,0 +1,49 @@
...
+  def workload
+    return error_assertion if expected < 0
+
+    "assert_equal #{formatted_expected}, Grains.square(#{input})"
+  end
+
+  private
+
+  def error_assertion
+      "assert_raises(ArgumentError) { Grains.square(#{input}) }"
```
Insti: dodgy indentation.

lib/grains_cases.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def skipped
+    index == 0 ? '# skip' : 'skip'
+  end
+
+  private
+
+  def formatted_expected
+    expected.to_s.reverse.gsub(/...(?=.)/,'\&_').reverse
```
Insti: This would be nicer as a function that took an argument.
```
def underscore_format(number)
```
Then you use it as: `underscore_format(expected)`

lib/grains_cases.rb
Insti: btw: This looks much nicer than it did last time I looked at it.

test/grains_cases_test.rb
```diff
@@ -0,0 +1,87 @@
+require_relative 'test_helper'
+
+class GrainsTest < Minitest::Test
```
Insti: I really like that you're unit testing the `Case` classes. :+1: 

Hopefully doing this will help you realise which methods should be part of some kind of `BaseCase` class that could be inherited from rather than always starting from `OpenStruct`

test/grains_cases_test.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def test_workload_is_not_implemented
+    assert_raises(NotImplementedError) { GrainsCase.new.workload }
+  end
+
+  def test_skipped_with_zero_index
+    assert_equal '# skip', GrainsCase.new(index: 0).skipped
+  end
+
+  def test_skipped_with_nil_index
```
Insti: Should a case ever have a `nil` index?

tommyschaefer: This is a little bit of a tricky one that I'm not sure on the best way to handle.

In my (limited) experience, when the canonical data has more than one "type" of test case, they both each follow the same JSON structure (e.g. [allergies](https://github.com/exercism/x-common/blob/master/exercises/allergies/canonical-data.json)). This lets you do something like this when assembling the cases (not an optimal solution necessarily, just demonstrating):

```ruby
i = -1

JSON.parse(data).flat_map do |type, test_data|
  test_data['cases'].map do |test_case|
    i += 1
    # Determine which kind of Case to make based on type
    Case.new(...) 
  end
end
```

But in this case the [JSON](https://github.com/exercism/x-common/blob/master/exercises/grains/canonical-data.json) is structured like:

```json
{
  "square": {
    "cases": [
      { "description": "...", "input": 1, "expected": 1 },
      ...
    ]
  },
  "total": {
    "description": "...",
    "expected": "..."
  }
}
```

Because of this, I'm not too sure how to get a real index for this case.

Do you have any thoughts on this? Would it be better to explicitly pass `index: false` for that case? It's still pretty strange though 😞

Insti: What does the `index` represent? What is it the index of?

tommyschaefer: It's the index of the set of test cases. Would something like this work:

```ruby
data = JSON.parse(data)
i = -1

cases = data['square']['cases'].map do |row|
  i += 1
  GrainsCase::SquareMethod.new(row.merge('index' => i))
end

cases << GrainsCase::TotalMethod.new(data['total'], index: i+1)
```

It adds a bit of complexity, but at least the index is correct.

Thank you so much for your help and feedback!

Insti: Getting closer :smile:

In the above example, how does `i` relate to `cases.size`?

Insti:
Is this ready to merge?

tommyschaefer:
As far as I know, this should be good to go 😄

Unless there's any other changes you all would like to see, in which case I'm happy to keep working at it!

Insti:
Thanks for all your work on this @tommyschaefer :heart:

PR #522, opened 2016-12-29
bin/generator: Big rewrite

Insti:
Can now generate tests without updating the test version number.

New command line options:
```
Usage: bin/generate [options] exercise-generator
	-f, --freeze                     Don't update test version
	-h, --help                       Displays this help message
	-v, --verbose                    Display progress messages
```

`sha1` in the "Common test data version: <%= sha1 %>" comment is now the
commit ID of the relevant canonical-data.json file

Files will only be written if they have changed.

Split code up into classes.
100% test coverage.

Added dependency: `require_all` gem
    
Just requiring files in (potentially random) directory order was not sufficient to satisfy all class dependencies. The `require_all` gem ensures dependencies are met as the files are `require`d


bmulvihill:
@Insti this looks cool, I'll see if I can take a look today!

Gemfile
```diff
@@ -4,3 +4,5 @@ gem 'minitest'
 gem 'rubocop', '0.36.0'
 gem 'simplecov'
 gem 'rake'
+gem 'values'
```
bmulvihill: I am always hesitant to add new dependencies to a project, do these gems provide anything that is essential and we can't handle ourselves?

lib/generator/files.rb
```diff
@@ -0,0 +1,44 @@
...
+      def relative_filename
+        Pathname.new(filename).relative_path_from(Pathname.new(repository_root)).to_s
+      end
+
+      def git_path
+        File.join(repository_root, '.git')
+      end
+    end
+
+    class MutableFile < ImmutableFile
```
bmulvihill: Does inheritance here make sense?  Does this pass the is a? test for inheritance, a MutableFile is a ImmutableFile?

lib/generator/files/exercise_files.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      private
+
+      def self.exercise_name(filename)
+        %r{([^/]*)_cases\.rb$}.match(filename).captures[0].tr('_','-')
+      end
+
+    end
+
+    # Assumes methods exist: paths, exercise_name
```
bmulvihill: Maybe we can add those methods into this module which raise NotImplementedError to ensure classes which include this module are in compliance

```ruby
  def paths
    raise NotImplementError
  end
```

Gemfile
Insti: Good point.

It depends on where you want to draw the line, the 2 I added were convenient for adding the functionality I wanted. I'll look into what it would take to eliminate the dependencies. I suspect not much.

bmulvihill: Yeah I don't suspect these 2 would cause any major issues - but I just know from experience dependency creep can be a big problem!

lib/generator/files.rb
Insti: This may be a case of misleading naming.

It's basically some files need to have a save method, and others don't. 
This way the intention of which files are meant to be `save`able is more clear.

It may not actually be important, and might just be a case of "just don't call save on the ones you don't want to save". 

Do you have any suggestions on how to handle this?

Doing the inheritance the other way around seems wrong if we need to define a save method which we then disable.

There's surely an elegant way of handling this hiding in there somewhere.

lib/generator/files/exercise_files.rb
Insti: Excellent point. Thanks.

This sort of thing is what I love about code reviews. :heart:

Insti:
Thanks @bmulvihill, Was that all you found or did you need to stop part way through?

bmulvihill:
@Insti I got pulled away midway through, I will try and finish looking it over this evening!

lib/generator/files.rb
bmulvihill: Would `class WritableFile < ReadableFile` make more sense?

lib/generator/files/exercise_files.rb
```diff
@@ -0,0 +1,83 @@
+require 'erb'
+
+module Generator
+  module Files
+    module GeneratorCases
+      def self.available(track_path)
```
bmulvihill: I would consider using `extend self` here since this seems to be more of a utility module, that would eliminate the need for self references
```ruby
module GeneratorCases
  extend self
```

bmulvihill:
Made some more small comments, overall I think this looks good and I like that there is test coverage now + more options when generating a test! 👍

lib/generator/files/exercise_files.rb
Insti: The style guide seems to suggest `module_function` rather than `extend_self` so I'll use `module_function` instead: https://github.com/bbatsov/ruby-style-guide#module-function

Insti:
- Addressed feedback provided by @bmulvihill.
- Applied some minor renamingss which will hopefully make things clearer and more consistent.

Anything (other than squashing) that needs to be done before we merge?

@kotp, @tommyschaefer : Have you had a chance to look over this?

lib/generator/template_values.rb
```diff
@@ -0,0 +1,48 @@
...
+      @test_cases = test_cases
+    end
+
+    def get_binding
+      binding
+    end
+  end
+
+  # This is not a standalone module, is that a problem?
+  module TemplateValuesFactory
```
tommyschaefer: Are the `TODO`s in this module meant to be completed after merging?

I think it's completely reasonable if so, especially since these changes make the process of generating test sooooooo much more pleasant (`--freeze` 😍).

Just wanted to double check that these didn't get missed 😄

Insti: Yes, the TODOs are all things that are intended to be separate follow-up PRs as they involve modifications to the exercise files.

kotp:
@insti let's squash what you want together and make this easier to do a "last review" on it?

Insti:
Ready for final review. Let's get this merged. Any objections?
@kotp, @tommyschaefer, @bmulvihill

bin/generate
```diff
@@ -3,34 +3,9 @@
...
-  require "#{cases}"
-rescue LoadError
-  puts "A generator does not currently exist for #{exercise}!"
-  exit(1)
-end
-
-klass = Object.const_get(cases.split('_').map(&:capitalize).join)
-Generator.new(exercise, klass).generate
+generator = Generator::CommandLine.new(paths).parse(ARGV)
+exit 1 unless generator
```
tommyschaefer: Would it be reasonable to add a little message here if generation is successful? Something simple maybe like the message you get when you initialize a new git repository.

kotp: It may be that silence signals success.

lib/generator/command_line.rb
```diff
@@ -0,0 +1,119 @@
...
+    def default_options
+      {
+        freeze: false,
+        verbose: false,
+        exercise_name: nil
+      }
+    end
+
+    def validate_options
+      return true unless @options[:help]
```
kotp: Since this is a method that is required to return a boolean, the name could be `valid_options?`  This communicates that usage.

lib/generator/command_line.rb
```diff
@@ -0,0 +1,119 @@
...
+      $stdout.puts usage
+      false
```
kotp: puts returns nil, and so would be sufficient to let the line above be the return.

lib/generator/command_line.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    def validate_paths
+      validate_metadata_repository
+    end
+
+    def validate_exercise
+      return true if @options[:exercise_name]
+      $stderr.puts "Exercise name required!\n"
+      $stdout.puts usage
+      false
```
kotp: Same here for query name and puts returning nil.

lib/generator/command_line.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    def validate_cases
+      return true if available_generators.include?(@options[:exercise_name])
+      $stderr.puts "A generator does not currently exist for #{@options[:exercise_name]}!"
+      false
```
kotp: Seeing a pattern here.

lib/generator/command_line.rb
```diff
@@ -0,0 +1,119 @@
...
+      @option_parser ||= OptionParser.new do |parser|
+        parser.banner = "Usage: #{$PROGRAM_NAME} [options] exercise-generator"
+        parser.on('-f', '--freeze', 'Don\'t update test version') { |value| @options[:freeze] = value }
+        parser.on('-h', '--help', 'Displays this help message') { |value| @options[:help] = value }
+        parser.on('-v', '--verbose', 'Display progress messages') { |value| @options[:verbose] = value }
+      end
+    end
+
+    def usage
+      option_parser.help +
```
kotp: Prefer using `<<` to concatenate the strings, as it uses a single string, appropriately.  One string for the garbage collector.

lib/generator/files.rb
```diff
@@ -0,0 +1,56 @@
...
+        Pathname.new(filename).relative_path_from(Pathname.new(repository_root)).to_s
+      end
+
+      def git_path
+        File.join(repository_root, '.git')
+      end
+    end
+
+    class Writable < Readable
+      def save_if_changed(new_content)
```
kotp: The name should be `save`.  We don't care if it is changed or not, but for efficiency, internally, it can make that distinction.  Also, if it is new content, it is not changed.  `content` should be sufficient.  Unless it is truly requiring that it is indeed new content.

lib/generator/files.rb
```diff
@@ -0,0 +1,56 @@
...
+        save(new_content) unless new_content == to_s
+        new_content
+      end
+
+      def save(new_content)
+        File.open(filename, 'w') do |file|
+          file.write new_content
+        end
+      end
```
kotp: same comments as save_if_changed.

lib/generator/template_values.rb
```diff
@@ -0,0 +1,48 @@
...
+      @version = version
+      @test_cases = test_cases
+    end
+
+    def get_binding
+      binding
+    end
+  end
+
+  # This is not a standalone module, is that a problem?
```
kotp: Comment is likely ready to be removed.  Or at least the question part of it, if it is for documenting/guidance.

lib/generator/template_values.rb
```diff
@@ -0,0 +1,48 @@
...
+      require 'ostruct'
+      require 'json'
+      require cases_require_name
+
+      # Compensate for the version.next that appears in template files.
+      # TODO: remove the .next from the template files and remove compensation
+      compensated_version = version - 1
+
+      TemplateValues.new(
+        # TODO: rename sha1 to abbreviated_commit_hash
```
kotp: This TODO appearse to be done.

bin/generate
Insti: Default is totally silent.
But if you use -v you get a success message.

test/fixtures/xruby/exercises/alpha/example.tt
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env ruby
```
kotp: This is a fixture, the folder as an example could be named `foo` or `bar` as traditionally used as examples, even working examples.

`alpha` may suggest there will be a `beta` and `delta`, etc.

bin/generate
tommyschaefer: Ah! Thats awesome! I totally missed that. Thank you 😄

test/fixtures/xruby/lib/alpha_cases.rb
```diff
@@ -0,0 +1,20 @@
+class AlphaCase < OpenStruct
```
kotp: `FooCase` or `BarCase` due to the heavy stroop effect on the word "alpha".

test/generator/files/track_files_test.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      def test_tests_template
+        subject = TestTrackFiles.new
+        assert_instance_of TestsTemplateFile, subject.tests_template
+      end
+    end
+
+    class TestsVersionFileTest < Minitest::Test
+      class TestTestsVersionFile < TestsVersionFile
+        def save_if_changed(new_content)
```
kotp: This will hopefully just be save, as mentioned in prior comment.

test/generator/template_values_test.rb
```diff
@@ -0,0 +1,59 @@
...
+    def test_get_binding
+      subject = TemplateValues.new(sha1: nil,version: nil, test_cases: nil)
+      assert_instance_of Binding, subject.get_binding
+    end
+  end
+
+  class TemplateValuesFactoryTest < Minitest::Test
+    class TestTemplateValuesFactory
+      def exercise_name
+        'alpha'
```
kotp: Consider `foo` or `bar` name instead.

test/fixtures/xruby/exercises/alpha/example.tt
Insti: I thought there might have been when I created it, but it turned out further letters weren't needed.

test/fixtures/xruby/lib/alpha_cases.rb
Insti: Although that is also the name of a computer hardware manufacturer..
Naming is hard.

kotp: This is a sanity test, really, right?  Could be `SanityCase`.

bin/generate
Insti: I'd like to change that, so the default was to print progress messages and a --quiet flag would make it silent, but that's for a future PR. I was trying to keep the functionality of this version as close to the one it was replacing as possible. (But some other things crept in.)

kotp: I really like the "silence is golden" approach.  Noisy if you ask for it.

lib/generator/files.rb
Insti: good points, fixed.

lib/generator/command_line.rb
```diff
@@ -95,8 +95,8 @@ def option_parser
     end
 
     def usage
-      option_parser.help +
-        "\nAvailable exercise generators:\n" +
+      option_parser.help <<
```
kotp: Quick safety check on the value that is returned from the `help` method.  The `<<` that I was referencing was the second concatenation, rather than both.  The reason is I did not peek at the help method, to make sure that modifying that in place would not cause a change that affects other calls to help.

Insti: `option_parser.help` returns a brand new string on every call, so no problems there.

Insti:
:tada: Hooray!

Thanks everybody for your reviews/feedback and help! :heart:

PR #523, opened 2016-12-30
Run non-exercise tests in CI

### LINKED ISSUE #519: Non-Exercise tests in CI
Hi there!

Just opening this to get some feedback. Do you all think that non-exercise (e.g. generator) tests should be run on CI? Currently all exercise tests get run as a part of CI, but tests in the `/test` directory do not.

If they should be run on CI, do you think it would be best to add `rake test` to the Travis scripts, or should the `test` target of the Makefile be modified to run these tests?

tommyschaefer:
This PR adds `rake test` to the travis config to run non-exercise tests in CI as discussed in #519.

Resolves #519.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ install:
 script:
   - rubocop -fs -D
   - bin/executable-tests-check
+  - rake test
   - make test
   - bin/fetch-configlet
   - bin/configlet .
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,15 +1,18 @@
 $LOAD_PATH.unshift File.expand_path('../lib', __FILE__)
 
-require 'simplecov'
 require 'minitest/autorun'
 require 'minitest/pride'
 
-SimpleCov.start do
-  add_filter '/test/'
-  add_group 'Utilities' do |file|
-    !(file.filename =~ /_cases\.rb$/)
+unless ENV['CI']
+  require 'simplecov'
+
+  SimpleCov.start do
+    add_filter '/test/'
+    add_group 'Utilities' do |file|
+      !(file.filename =~ /_cases\.rb$/)
+    end
+    add_group 'Cases', '_cases.rb'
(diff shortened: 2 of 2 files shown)


Insti:
I think we need to first add some magic to not create test coverage files in Travis.

Insti:
Is this something you can do @tommyschaefer ?

tommyschaefer:
From the Travis log:

```
$ rake test
Run options: --seed 21175

# Running:

...................

Fabulous run in 0.009044s, 2100.8717 runs/s, 2432.5883 assertions/s.

19 runs, 22 assertions, 0 failures, 0 errors, 0 skips


The command "rake test" exited with 0.
```

Insti:
Clearly the answer was yes. :heart:

Insti:
This is a separate issue that I noticed while looking through the Travis log:

> Use assert_nil if expecting nil from /tmp/all-your-base.FzoC9cwNdy/all_your_base_test.rb:280:in `test_second_base_is_negative'. This will fail in MT6

Are these still the old tests? Is this an issue in the new ones?

tommyschaefer:
> Are these still the old tests? Is this an issue in the new ones?

Yes, this is a warning from the old tests. The new tests generated in #510 don't have this issue.

The old tests use `assert_equal nil, ...` rather than `assert_nil ...` which causes the warnings, but the new tests never expect `nil` since it's actually an error case. Now we expect to raise an argument error.

Insti:
FWIW: This is a situation where I would have recommended NOT squashing the commits, as adding the rake command to the travis config and altering the test_helper.rb are two separate actions.

It doesn't matter too much here, but it's perhaps a little lesson that squashing is not ALWAYS the thing to do.

tommyschaefer:
Ah! Good point! Thank you for the feedback!

kotp:
You can still separate them out, if you wanted to in this PR. But yes, no worries.

PR #524, opened 2016-12-31
series: series_test.rb test for strings not arrays

stevensonmt:
Previous test looked for arrays of Ints but exercise prompt in readme called for strings.  This update makes tests look for strings.

### DIFF
--- a/exercises/series/example.rb
+++ b/exercises/series/example.rb
@@ -1,27 +1,16 @@
 class Series
-  def initialize(numeric_string)
-    @digits = convert_to_digits(numeric_string)
+  def initialize(series)
+    @series = (series)
   end
-
-  def slices(length)
-    if length > digits.length
-      fail ArgumentError.new('Not enough digits')
+  def slices(n)
+    if n > @series.length
+      raise ArgumentError
     end
-    result = []
-    i = -1
-    begin
-      i += 1
-      i2 = i + length - 1
-      result << digits[i..i2]
-    end while i2 < digits.size - 1
-    result
-  end
-
-  private
-
-  attr_reader :digits
-
-  def convert_to_digits(s)
-    s.chars.to_a.map(&:to_i)
+    substrings = []
+    until @series.length < n
+      substrings << @series[0..n-1]
+      @series = @series[1..-1]
+    end
+    substrings
   end
 end
--- a/exercises/series/example.rb
+++ b/exercises/series/example.rb
@@ -6,10 +6,11 @@ class Series
     if n > @series.length
       raise ArgumentError
     end
+    series = @series
     substrings = []
-    until @series.length < n
-      substrings << @series[0..n-1]
-      @series = @series[1..-1]
+    until series.length < n
+      substrings << series[0..n-1]
+      series = series[1..-1]
     end
     substrings
   end
--- a/exercises/series/series_test.rb
+++ b/exercises/series/series_test.rb
@@ -94,4 +94,11 @@ class SeriesTest < Minitest::Test
       series.slices(slice_string.length + 1)
     end
   end
+  
+  def test_sequential_slices
+    #skip
+    series = Series.new('1234')
+    series.slices(2)
+    series.slices(3)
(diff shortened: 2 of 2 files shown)


stevensonmt:
Not sure I understand why it's failing this check.  Looks as though it is checking it against a solution that is based on the previous tests looking for arrays of Ints.  I don't know where it's getting that master solution file from so can't change it.  Obviously that solution was written to pass the previous tests, which are not consistent with the exercise prompt in the readme.

Insti:
Thanks @stevensonmt I'm happy to see you worked out how to make a pull request.

The tests are run against an example solution which can be found at: https://github.com/exercism/xruby/blob/master/exercises/series/example.rb

Do you want to try and update that to match your new tests?

You can come and ask questions in the Gitter chatroom: https://gitter.im/exercism/xruby if you want some 'live' help.

stevensonmt:
Updated the example.rb so that it now passes the tests when the tests expect substrings instead of arrays of ints.

exercises/series/example.rb
```diff
@@ -1,27 +1,16 @@
...
-  end
-
-  private
-
-  attr_reader :digits
-
-  def convert_to_digits(s)
-    s.chars.to_a.map(&:to_i)
+    substrings = []
+    until @series.length < n
```
Insti: This has the same problem as your original solution, in that it won't work if called twice.

exercises/series/series_test.rb
```diff
@@ -6,87 +6,87 @@
 class SeriesTest < Minitest::Test
   def test_simple_slices_of_one
     series = Series.new('01234')
-    assert_equal [[0], [1], [2], [3], [4]], series.slices(1)
+    assert_equal ['0', '1', '2', '3', '4'], series.slices(1)
   end
 
   def test_simple_slices_of_one_again
-    skip
+    #skip
```
Insti: We need to start with all tests except the first skipped.

exercises/series/example.rb
Insti: Maybe it would help if you added an additional test case for this at the end.

Insti:
Nice work @stevensonmt this is heading in the right direction.

stevensonmt:
@Insti -- Should have caught that about skipping all but the first test.  Added test for sequential calls of slices method and updated the example so that it does not mutate the original string.

exercises/series/series_test.rb
```diff
@@ -6,92 +6,99 @@
...
     assert_raises ArgumentError do
       series.slices(slice_string.length + 1)
     end
   end
+  
+  def test_sequential_slices
+    #skip
+    series = Series.new('1234')
+    series.slices(2)
+    series.slices(3)
```
Insti: Remember to `assert` something. :smile:

kotp:
@stevensonmt when you are ready, you can squash the commits for the next review.  There is a way to leave the skips there and still be able to run the tests, which is why I think the tests may have been skipped (it would be a normal reaction while developing).

Here is that command:

`ASSIGNMENT=series make test-assignment`

This will run the tests and automatically skip them, while doing the other needful things.  Not magic, you can read the Makefile to see exactly what is happening.

Insti:
I'm still seeing the `skip`s are commented out.

Insti:
Great, thanks. Is there anything else you want to do before we merge this @stevensonmt ?

kotp:
Looks better.

stevensonmt:
I don't think there's anything else to do, but obviously I've a long way to go at recognizing the things that need doing.

Insti:
Congratulations @stevensonmt, you're an open source contributor! :medal_sports: 

Thanks a lot for noticing a problem and doing something about fixing it and for your all effort in working out what you needed to do.

There's usually always more to do :smile: , but no one person needs to do it all! :heart:

Insti:
PS. It might be a week or two before the change makes it to the live site, live updates are not happening so often over the holiday period. But you will see it there eventually.

PR #525, opened 2017-01-01
Replicate Makefile functionality in Rakefile

tommyschaefer:
Hi there!

In #519 @Insti mentioned:

> Longer term, I'd like to see `make` phased out in favor of `rake`.

This is a very rough (so far) first pass at making this change.

This change adds the following rake tasks:
- `$ rake test:all` or `$ rake` runs all tests (`/test/*`, and all exercise tests)
- `$ rake test:dev` runs all development tests (`/test/*`)
- `$ rake test:assignment[assignment_name]` runs the tests for a specific assignment
- `$ rake test:assignments` runs all exercise tests

Does this seem like a reasonable implementation? Theres definitely some work to be done in cleaning up the code, but I think could be a possible starting place.

**Still to be done before merging:**
- [x] Settle on command interface
- [x] Make changes suggested in feedback
- [x] Test task classes
- [x] Document once a interface has been settled on


tommyschaefer:
I ran both `rake test:all` and `make test` on CI and did a diff that can be found [here](https://www.diffchecker.com/dfXBqMWd).

Looks like the only difference is the ordering of exercises with `-` in the name.

.travis.yml
```diff
@@ -8,7 +8,7 @@ install:
 script:
   - rubocop -fs -D
   - bin/executable-tests-check
-  - rake test
+  - rake test:all
   - make test
```
tommyschaefer: I left this in with the `Makefile` for now to compare output.

Insti: Can we make `rake test` an alias for `test:all`
This would mean we wouldn't need to change line 11 either.

Insti:
Nice work @tommyschaefer 

> $ rake test:assignment[assignment_name] runs the tests for a specific assignment

Replace "assignment" with "exercise"

Exercise is the current Exercism terminology for a track specific implementation of a problem.

> I ran both rake test:all and make test on CI and did a diff

Excellent. Great idea.

> Looks like the only difference is the ordering of exercises with - in the name.

I don't think this is a problem.

A wishlist item: Can we make it so that `rake test all-your-base` runs the tests for the specified exercise without requiring extra `:[]` characters?

tommyschaefer:
> Replace "assignment" with "exercise"

👍  I was going to ask about this, but I forgot.

> A wishlist item: Can we make it so that rake test all-your-base runs the tests for the specified exercise without requiring extra :[] characters?

Yes, but I'm not sure it's the best thing to do.

When you call a rake task, each argument is expected to be a rake task. For example (from rails land):

```
rake db:create db:migrate
```

will run the `db:create` task and then the `db:migrate` task.

To make `rake test all-your-base` work, we could define empty rake tasks on the fly with something like:

```ruby
ARGV.each { |a| task a.to_sym do ; end }
```

to prevent rake from doing:
```


----------------------------------------------------------------
running tests for: all-your-base
/Users/tommyschaefer/.rvm/rubies/ruby-2.3.1/bin/ruby -I lib -r disable_skip.rb /var/folders/xg/p1hx36j92cv5v0dn3fsjtqrm0000gn/T/all-your-base20170102-12848-11i0cwc/all_your_base_test.rb
Run options: --seed 42978

# Running:

. * 22

Finished in 0.002615s, 8413.4690 runs/s, 8413.4690 assertions/s.

22 runs, 22 assertions, 0 failures, 0 errors, 0 skips
```
```
rake aborted!
Don't know how to build task 'all-your-base' (see --tasks)
/Users/tommyschaefer/.rvm/gems/ruby-2.3.1/gems/rake-12.0.0/exe/rake:27:in `<top (required)>'
/Users/tommyschaefer/.rvm/gems/ruby-2.3.1/bin/ruby_executable_hooks:15:in `eval'
/Users/tommyschaefer/.rvm/gems/ruby-2.3.1/bin/ruby_executable_hooks:15:in `<main>'
(See full trace by running task with --trace)
```

I'm not sure if this is the right thing to do. It feels a little bit icky to me, but if the convenience outweighs the *ickyness-cost* 😄 , I'm happy to make it work like that!

I haven't pushed it yet, but we also have some update documentation with an updated implementation of `xtest` that would let you do:

```
xtest all-your-base
```

But it sure would be nice rake just worked like that 😞

tommyschaefer:
For reference on the ⬆️  problem, [4 Ways to Pass Arguments to a Rake Task](http://cobwwweb.com/4-ways-to-pass-arguments-to-a-rake-task) is a nice resource 😄

Insti:
Thanks, that was all very informative.

Would `rake <exercise-name>` as a target to run an individual test make sense?

(I can't tell if I'm just making stuff up that isn't actually useful.)

kotp:
> Would rake exercise-name as a target to run an individual test make sense?
It surely would.  It is better than having another name in the brain to remember.

Rakefile
```diff
@@ -33,4 +34,4 @@ namespace :test do
       Rake::Task['test:exercise'].execute(exercise: exercise)
     end
   end
-end
+end
```
Insti: Tried using Github's merge conflict resolution web interface.
Won't be trying that again. :disappointed:

Rakefile
```diff
@@ -1,9 +1,37 @@
...
+namespace :test do
+  desc 'Run all development and exercise tests'
+  task :all do
+    Rake::Task['test:dev'].invoke
+    Rake::Task['test:exercises'].invoke
+  end
+
+  Rake::TestTask.new :dev do |task|
+    task.pattern = 'test/*_test.rb'  
+    task.pattern = 'test/**/*_test.rb'
```
Insti: Line 22 is correct, line 21 is redundant.
(This was my fault.)

Rakefile
kotp: I did the exact same thing.

tommyschaefer:
Still working on the implementation, but I like this interface a lot more than the original:

`$ rake test` runs all tests
`$ rake test:dev` runs all dev tests (tests in `/test`)
`$ rake test:exercises` runs all exercise tests
`$ rake test hamming` runs only the hamming exercise tests

exercises can also be chained, so `$ rake test hamming series` will run both hamming and series.

Arguments can be passed to test running command like so `$ rake test hamming -- -p`

Test running arguments are also passed to all running tests, so `$ rake test -- -p`, `$ rake test hamming series -- -p`, and `$ rake test:exercises -- -p` will run each of the requested tests with the `-p` option.

What do you all think of this? Is this intuitive and easy to use enough?

To me personally it feels more natural to use `rake test hamming` over `rake hamming`, but how do you all feel about that? For me, it isn't immediately obvious that `rake hamming` will run the hamming tests (whereas the alternative makes it very clear). I think if I was unfamiliar with the xruby project and I wanted to run exercise tests I would reach for `rake test <exercise>` over the alternative. In both cases, though, I think a definite improvement over the first pass.

Thank you all for your time and feedback!

lib/tasks/exercise_test_task.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def run_exercise_tests_in(dir)
+    ruby "-I lib -r disable_skip.rb #{dir}/#{srcfile}_test.rb #{test_options}"
+  end
+
+  def test_options
+    return '' unless ARGV.include?('--')
+
+    ARGV.drop_while { |e| e != '--' }.drop(1).join(' ')
```
kotp: This is definitely tolerable, except that it is very much duplicated in another area.  There is a PR on your repository to address this.

Rakefile
```diff
@@ -1,9 +1,40 @@
...
+  ExerciseTestTask.new :exercise do |task|
+    task.description = "Run the tests for a specific exercise"
+  end
+
+  desc 'Run the tests for all exercises'
+  task :exercises do
+    Dir.foreach('exercises').each do |exercise|
+      next if %w(. ..).include?(exercise)
+
+      Rake::Task['test:exercise'].execute(exercise: exercise)
```
Insti: I'd like to see a method extracted here that iterated over an array of exercises. It would probably be similar to what's happening around line 13.

Rakefile
```diff
@@ -1,9 +1,40 @@
...
+  end
+
+  ExerciseTestTask.new :exercise do |task|
+    task.description = "Run the tests for a specific exercise"
+  end
+
+  desc 'Run the tests for all exercises'
+  task :exercises do
+    Dir.foreach('exercises').each do |exercise|
+      next if %w(. ..).include?(exercise)
```
Insti: This is pretty horrible. Filter out the things you don't want before you get here.

Insti:
```ruby
FileList['exercises/*'].pathmap('%f')
```

Rakefile
```diff
@@ -1,22 +1,22 @@
...
   if exercises.any?
-    exercises.each do |e|
-      task(e) { Rake::Task['test:exercise'].execute(exercise: e) }
-    end
-  else
-    Rake::Task['test:dev'].invoke
-    Rake::Task['test:exercises'].invoke
+    exercises.each { |e| task(e) { } }
+    ExerciseTests.run(exercises)
+    next
```
Insti: I think I prefer the `else` version

Rakefile
```diff
@@ -1,22 +1,22 @@
...
   exercises = ARGV.take_while { |e| e != '--' }.drop(1)
 
   if exercises.any?
-    exercises.each do |e|
-      task(e) { Rake::Task['test:exercise'].execute(exercise: e) }
-    end
-  else
-    Rake::Task['test:dev'].invoke
-    Rake::Task['test:exercises'].invoke
+    exercises.each { |e| task(e) { } }
```
Insti: What is this line doing?

Rakefile
```diff
@@ -25,16 +25,8 @@ namespace :test do
...
-  end
-
   desc 'Run the tests for all exercises'
   task :exercises do
-    Dir.foreach('exercises').each do |exercise|
-      next if %w(. ..).include?(exercise)
-
-      Rake::Task['test:exercise'].execute(exercise: exercise)
-    end
+    ExerciseTests.run_all
```
Insti: This seems like the wrong level of abstraction.
You're hiding the similarity with line 14.

Rakefile
tommyschaefer: This defining no-op tasks for the given exercises so that rake doesn't do this:

```
rake aborted!
Don't know how to build task 'all-your-base'
```

Rakefile
tommyschaefer: Do you think `ExerciseTests.run(FileList['exercises/*'].pathmap('%f'))` then?

I would really like there to be a "name" for `FileList['exercises/*'].pathmap('%f')`, but I'm not sure where to put it. `ExerciseTests.run(ExerciseTests.all)` doesn't feel right to me.

Thank you so much for all of your feedback!

Rakefile
Insti: Is this something that anyone using rake should know?

If not putting it in an intention revealing method could be good:
```
create_null_tasks_for(exercises)
```

Rakefile
Insti: Introduce an intention revealing name.
```
all_exercises = FileList['exercises/*'].pathmap('%f')
ExerciseTests.run(all_exercises)
```

Rakefile
```diff
@@ -1,9 +1,35 @@
...
 require 'rake/testtask'
 
-desc 'rake with no argument will run "rake test"'
-task default: :test
+require_relative 'lib/tasks/exercise_tests_runner'
+require_relative 'lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure'
 
-Rake::TestTask.new do |task|
-  task.pattern = 'test/**/*_test.rb'
+KnowsWhatExercisesAreCalledAndWhereTheyLiveInTheDirectoryStructure.all_exercises.each do |exercise|
```
tommyschaefer: This can be extracted in the same way that `Rake::TestTask` does

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
```diff
@@ -2,4 +2,26 @@ class KnowsWhatExercisesAreCalledAndWhereTheyLiveInTheDirectoryStructure
   def self.all_exercises
     FileList['exercises/*'].pathmap('%f')
   end
+
+  def self.exercise_directory(exercise)
+    "exercises/#{exercise}/."
```
Insti: The trailing /. here might cause you issues in the future.

Rakefile
Insti: Yes definitely do that (eventually.)

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
```diff
@@ -0,0 +1,34 @@
+class Exercise
```
tommyschaefer: Still moving in the right direction, or back in the wrong direction 😄

tommyschaefer: Not sure if this was the right jump to make, but it feels "less icky" to me than before.

Exercises know where their files are and the test runner uses an exercise and test options to run the exercise tests.

kotp: The name could be "exercise_tracking.rb" for the file, perhaps.

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
```diff
@@ -0,0 +1,34 @@
...
+    "#{srcfile}.rb"
+  end
+
+  def test_file
+    "#{srcfile}_test.rb"
+  end
+
+  private
+
+  def srcfile
```
kotp: `srcfile` should be `source_file`.

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
```diff
@@ -0,0 +1,34 @@
+class Exercise
+  def self.all
+    FileList['exercises/*'].pathmap('%f').map { |e| new(e) }
```
tommyschaefer: If the exercise abstraction is a good one, should this return instances of exercises, or exercise names (and would be turned into exercise instances currently in the Rakefile)?

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
tommyschaefer: Ah, yes :+1:

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
tommyschaefer: Ah, yeah :man_facepalming: Forgot to rename the file

Insti: Generally the right direction. 
Renaming a class to something more generic like that can be dangerous.
But maybe it's the right time to do it.
I like how you've made it instantiable.

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
Insti: I like that it returns instances, and how  you've used `to_s` to make it do the sensible thing.

I would extract an `exercise_names` method.

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
Insti: ... But it's not a file, or a filename.
It's some kind of `file_base_name` perhaps?

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
kotp: Passing objects is good probably preferred.

lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure.rb
kotp: Indeed, the name I know is not quite right.  But what it is... to be discovered.

Rakefile
```diff
@@ -1,9 +1,31 @@
...
+  next if RakeArguments.more_tasks?
+
+  Rake::Task['test:dev'].invoke
+  Rake::Task['test:exercises'].invoke
+end
+
+namespace :test do
+  desc 'Run all development tests located in the test directory'
+  Rake::TestTask.new :dev do |task|
+    task.options = RakeArguments.task_options
```
tommyschaefer: This allows us to call `rake test:dev -- -p` and have it work. Previously it would only apply `-p` if you ran `TESTOPTS=-p rake test:dev`.

Also makes it use `-p` for `test:dev` tests when you run `rake test -- -p`

tommyschaefer:
I'm excited this feels like it's moving in the right direction to me! How do you all feel about these changes?

It feels like extracting and injecting these new classes has made everything *way* more testable.

Still a fair amount of work adding tests and updating documentations, but does this feel like it's getting closer to you all?

Rakefile
```diff
@@ -1,24 +1,19 @@
 require 'rake'
 require 'rake/testtask'
 
+require_relative 'lib/tasks/exercise'
+require_relative 'lib/tasks/exercise_test_tasks'
 require_relative 'lib/tasks/exercise_tests_runner'
```
Insti: `lib/tasks/exercise_test_tasks` is the only thing that should be included here.
The other includes belong inside that file.

Rakefile
```diff
@@ -1,24 +1,19 @@
...
-Exercise.all.each do |exercise|
-  task exercise do
-    ExerciseTestsRunner.new(
-      exercise: exercise,
-      test_options: ARGV.drop_while { |e| e != '--' }.drop(1).join(' ')
-    ).run
-  end
-end
+ExerciseTestTasks.new exercises: Exercise.all,
+                      options: RakeArguments.task_options
```
Insti: I really like this. 
I'd probably move it to the bottom of the file though.

Insti: I'm not sure `task_options` is the right name.

Rakefile
```diff
@@ -1,24 +1,19 @@
...
 
 task default: 'test'
 
 desc 'Run individual exercises or run all development and exercise tests'
 task :test do
-  exercises = ARGV.take_while { |e| e != '--' }.drop(1)
-  next if exercises.any?
+  next if RakeArguments.more_tasks?
```
Insti: This is really weird to read.

Rakefile
```diff
@@ -27,6 +22,7 @@ end
 namespace :test do
   desc 'Run all development tests located in the test directory'
   Rake::TestTask.new :dev do |task|
+    task.options = RakeArguments.task_options
```
Insti: so that's where the task_options name comes from.

There's probably a standard name for options after `--` that would be better. (see also name issue above.)

Rakefile
```diff
@@ -1,24 +1,19 @@
 require 'rake'
 require 'rake/testtask'
 
+require_relative 'lib/tasks/exercise'
+require_relative 'lib/tasks/exercise_test_tasks'
 require_relative 'lib/tasks/exercise_tests_runner'
-require_relative 'lib/tasks/knows_what_exercises_are_called_and_where_they_live_in_the_directory_structure'
+require_relative 'lib/tasks/rake_arguments'
```
kotp: since this is in tasks, perhaps `rake_arguments` is redundant.

Rakefile
kotp: `flags` or `options` are probably appropriate.

Rakefile
kotp: next unless done?

lib/tasks/rake_arguments.rb
```diff
@@ -0,0 +1,9 @@
+class RakeArguments
+  def self.task_options
+    ARGV.drop_while { |e| e != '--' }.drop(1).join(' ')
```
Insti: I'd like to see this method just return an array of arguments.
Let whoever cares about them being joined by spaces do that themselves.

lib/tasks/rake_arguments.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+
+  def self.more_tasks?
+    ARGV.take_while { |e| e != '--' }.drop(1).any?
```
Insti: I'd like to see this method just return an array of arguments.
Let whoever cares about them being `any?` do that themselves.

Rakefile
kotp: Hmm... the `--` generally signals that options are done being passed to the command being ran, and used for some command that the first command is utilizing.  The name task_options may be appropriate here.  Unsure what to do here.

lib/tasks/rake_arguments.rb
```diff
@@ -0,0 +1,9 @@
+class RakeArguments
+  def self.task_options
```
kotp: The filename... this has to do with the task options... so they aren't arguments to rake, they are task arguments... or task options... Perhaps my first choice of RakeArguments is to blame.  I think this module is more like TaskArguments, though generally that could be `Arguable`. Pun intended, but only half joking.

lib/tasks/rake_arguments.rb
Insti: The resulting method could probably be called `command_options` which makes `task_options` make more sense as a name.

lib/tasks/exercise_test_tasks.rb
```diff
@@ -0,0 +1,25 @@
+require 'rake/tasklib'
+
+class ExerciseTestTasks < Rake::TaskLib
+  def initialize(exercises:, options:, test_runner: ExerciseTestsRunner)
+    @exercises = exercises
+    @options = options
```
Insti: This is where you want to `join(' ')` them

lib/tasks/rake_arguments.rb
kotp: I agree with @Insti on lines 3 and 7.  Let the block be the end of it, let the return be the return.  This does suggest a couple of problems with the encapsulation on its own, though.  Let it be a class, instantiate it directly, and have it return itself, and respond to `any?` and `to_s`

Insti: actually
`tasks` and `options`

Rakefile
Insti: `options: RakeArguments.options`

Rakefile
Insti: `task.options = RakeArguments.options`

Rakefile
Insti:
```
next unless RakeArguments.tasks.size == 1

Comment (Insti) [inline Rakefile]: I still prefer wrapping the steps below in a conditional, to using `next`, but that may just be an unimportant stylistic preference.

I like blocks to have have a single exit point at the end, unless there is a **really** good reason not to do that.

Comment (Insti) [inline Rakefile]: The non-standard thing we are doing here is *ignoring* the test option if it comes first.

I'm beginning to think that we should not be doing that.

Comment (Insti) [inline Rakefile]: It also has implications if test is not the first task specified, consider:
`rake test isogram test wordy`

Comment (kotp) [inline Rakefile]: I think that example would be invalid.  Or should be.  `rake test deploy style hamming wordy` would not make sense.  Let the rake file present what things should be together, let the arguments end up being simple, intuitive, but not clever.

Comment (kotp) [inline Rakefile]: Having `rake test wordy test hamming` would also preclude us from having an exercise named 'test'.  `rake task target target target target` would be what I would aim for.

File: Rakefile (hunk 19 above)
Comment (tommyschaefer) [inline Rakefile]: Yeah, maybe `options` does make sense. The reason I went with `task_options` was to make it a little more clear to myself that it was the options for the task, not the options for rake.

File: Rakefile (hunk 18 above)
Comment (tommyschaefer) [inline Rakefile]: I'm beginning tho think this workaround has too many problems to be the right solution.

For example
```
rake test --backtrace
```

Will not run any tests because there are more "tasks". This could be fixed by checking the next task agains the list of tasks, but I think this is really just a bandaid on the fact that rake is really not made to work this way.

Personally, I think it would be preferable to have `rake test` **always** run all tests, and namespace the other tests, so they would be `rake test:hamming`. This makes calling multiple exercise tests at once a bit more wordy(`rake test:hamming test:series`), but I'm not sure that's a thing that happens too much.

File: lib/tasks/rake_arguments.rb (hunk 23 above)
Comment (Insti) [inline lib/tasks/rake_arguments.rb]: This a case of the name causing one to think in misleading directions.
There is nothing rake specific about the arguments, other than that they happen to be arguments to the rake command.

File: Rakefile (hunk 18 above)
Comment (kotp) [inline Rakefile]: Make the `--` be a separate word, ignore it otherwise.  This means `\b--\b`.  Should be an easy enough fix.

File: lib/tasks/exercise_test_tasks.rb (hunk 24 above)
Comment (Insti) [inline lib/tasks/exercise_test_tasks.rb]: if `join(' ')`ing them is even necessary...

File: Rakefile (hunk 18 above)
Comment (Insti) [inline Rakefile]: >  this is really just a bandaid on the fact that rake is really not made to work this way.

> have `rake test` always run all tests, and namespace the other tests, so they would be `rake test:hamming`

I think this is a sensible solution.

Comment (kotp) [inline Rakefile]: I agree, I can't really see a common use where we are going to run 2 or three different exercises in one command.  If so, we can do so easily with the shell.

File: Rakefile [hunk 25]
```diff
@@ -1,9 +1,28 @@
 require 'rake'
 require 'rake/testtask'
 
-desc 'rake with no argument will run "rake test"'
-task default: :test
+require_relative 'lib/tasks/exercise'
+require_relative 'lib/tasks/exercise_test_tasks'
+require_relative 'lib/tasks/rake_arguments'
```
Insti: I still only want to include `exercise_test_tasks` here.
The other things *are* references here, but *only* because `exercise_test_tasks` is.

lib/tasks/rake_arguments.rb
```diff
@@ -0,0 +1,5 @@
+module RakeArguments
+  def self.flags
+    ARGV.drop_while { |e| e != '--' }.drop(1)
```
Insti: Now we have a module, with one method of one line it might be time to consider inlining it again.

Rakefile line 17:
```
flags = ARGV.drop_while { |e| e != '--' }.drop(1)
```

I like it much better when the outer level is the only thing accessing the global/environment variables.

README.md
```diff
@@ -34,28 +34,35 @@ suite, you can run the file with a shim that temporarily disables them:
...
 root.  It will disable the skip calls for you automatically, it does the
 same thing as the above.
 
-If you would like to use the `make` tool to run a single test, while
+If you would like to use the `rake` tool to run a single test, while
 developing clock, for example, you can do something like this:
 
 ```sh
-ASSIGNMENT=clock ARGS='-p' make test-assignment
+rake test:exercise[clock]
```
Insti: Remember to update the instructions.

test/tasks/exercise_test.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      FileList.stub :[], FileList["#{dir}/*"] do
+        assert_equal true, Exercise.all.all? { |e| e.instance_of?(Exercise) }
+        assert_equal ['test', 'test2'], Exercise.all.map(&:name)
+      end
+    end
+  end
+
+  def test_name
+    exercise = Exercise.new(:name)
```
Insti: Is name being a symbol an expected usecase?

tommyschaefer: No, not anymore. I'll update this... thanks for the catch!

Rakefile
tommyschaefer: Hmm, how should that work then?

Should `exercise_test_tasks` require them, even though it doesn't use any of those classes?

lib/tasks/rake_arguments.rb
tommyschaefer: I definitely agree that the module is a bit overkill now, but I'm a little unsure of what the better alternative is. Right now the `flags` method is used in 2 places in the Rakefile (dev tests and exercise tests). Should I duplicate that in both places?

I'm not able to define a `flags` method in the Rakefile.

Insti: It's not a method it's a variable assignment.

Just delete the tests, they're not necessary any more.

Rakefile
Insti: > Should exercise_test_tasks require them, even though it doesn't use any of those classes?

Well, There's a philosophical starting point for you, the question becomes why not?

1. When RakeArguments goes away and options becomes a variable that problem goes away.

2. You need `Exercise` so you can call `Exercise.all` as an argument to `ExericiseTestTasks.new` and when defining the `exericses` task, but I would argue that defining the task to run all the exercises belongs in `ExerciseTestTasks` also.

And maybe `exercises:` should not be an argument to `ExerciseTestTasks.new` at all.

When you take care of these 2 things, the `Exercise` dependency disappears and everything becomes more elegant, which suggests to me it's the right thing to do.

Rakefile
```diff
@@ -1,9 +1,24 @@
...
 
-desc 'rake with no argument will run "rake test"'
-task default: :test
+require_relative 'lib/tasks/exercise_test_tasks'
 
-Rake::TestTask.new do |task|
-  task.pattern = 'test/**/*_test.rb'
+task default: 'test'
+
+flags = ARGV.drop_while { |e| e != '--' }.drop(1)
```
Insti: Define this closer to where it's used, within the `namespace :test` block

Insti: Also, in both places you use it you `join(' ')` it, so maybe you could add that back in here.

lib/tasks/exercise_test_tasks.rb
```diff
@@ -0,0 +1,36 @@
...
+    @test_runner = test_runner
+
+    define
+  end
+
+  def define
+    desc 'Run the tests for all exercises'
+    task exercises: exercises
+
+    exercises.each { |exercise| define_task_for(exercise) }
```
Insti: You have an abstraction level mismatch in the `define` method.

lib/tasks/exercise_tests_runner.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  attr_reader :exercise, :test_options
+
+  def setup_exercise_files_in(dir)
+    FileUtils.cp_r exercise.directory, dir
+    FileUtils.mv "#{dir}/#{exercise.example_file}", "#{dir}/#{exercise.testable_example_file}"
+  end
+
+  def run_exercise_tests_in(dir)
+    ruby "-I lib -r disable_skip.rb #{dir}/#{exercise.test_file} #{test_options}"
```
tommyschaefer: Any ideas on how to test this? I'm pretty sure I can test the rest with stubs and mocks, but I'm not sure about this part.

I'd like to test something like: given an `exercise` and `options`, I expect this command to be run. Or maybe I expect this output, but that's tricky because of random seeds (I think anyway?).

Insti: You want to mock `ruby` and assert that it gets called with the right options.

When you go down that road you end up extracting several methods/objects that purely exist to aid testing. (I don't think this is a bad thing, but that is not a generally popular opinion, but most people are **really** bad at testing.)

1. test that the arguments to the ruby command are as expected. (extract a method that composes them, this is easily tested by string comparison.)
2. test that the `ruby` command is called with some arguments (because of 1 we don't care what they are, so if they are `-e 'p true'` it doesn't really matter, and when you get to THAT stage you begin to question whether mocking `ruby` is worth the effort or if you should just call it anyway.)

Note: that `ruby` here is a rake FileUtils method: http://rake.rubyforge.org/FileUtils.html#method-i-ruby

tommyschaefer:
Sounds great! Thank you so much @insti and @kotp for the feedback!! I really like how this turned out with guidance from you to!

I'm going to work on this a bit more tonight (adding some tests and updating documentation). Once I'm finished, should I wait for you all to merge, or should I merge myself?

kotp:
You should squash the commits... it will make for a simpler final pass.

README.md
```diff
@@ -34,42 +34,32 @@ suite, you can run the file with a shim that temporarily disables them:
...
-ARGS is where you can put additional arguments, here I am demonstrating
-the `-p` argument which may give you 'pride' output.
-
-If you use zsh, you can use the following function to make this process
-simple.
-
+To pass arguments to the test command, like `-p` for example, you can run
+the following:
 ```sh
-xtest () { ASSIGNMENT=$1 ARGS=$2 make test-assignment }
```
tommyschaefer: Should `xtest` still be in the docs for running tests?

I went ahead and removed it, but I have no objection to putting it back in 😄

bin/local-status-check
```diff
@@ -3,7 +3,7 @@
 puts "Starting #{__FILE__} checks..."
 
 commands_to_run = [
-  'make test',
+  'rake test',
```
tommyschaefer: Heads up:

If you're using `local-status-check` and git hooks, you will have to run `bin/setup-git-hooks` again to update the hook.

tommyschaefer:
Note that we have the same warnings discussed in one of the other PRs:
```
/home/travis/build/exercism/xruby/lib/tasks/exercise_tests_runner.rb:23: warning: private attribute?
/home/travis/build/exercism/xruby/lib/tasks/exercise_tests_runner.rb:23: warning: private attribute?
/home/travis/build/exercism/xruby/lib/tasks/exercise_test_tasks.rb:24: warning: private attribute?
/home/travis/build/exercism/xruby/lib/tasks/exercise_test_tasks.rb:24: warning: private attribute?
```

What was the consensus... Include workaround the gets rid of these warnings for ruby versions < 2.3.0, or ignore them?

README.md
Insti: removing it is good. `rake test:exercise` will do the same thing with less trouble.

Insti:
Include workarounds, we should be warning free.

README.md
kotp: Remove it... `rake` is more idiomatic when it comes to Ruby.  `xtest` was a solution to an itch.

Insti:
:tada: Thanks for all your work on this @tommyschaefer :heart:

PR #528, opened 2017-01-04
bin/generate: Default display progress messages.

Insti:
Default `bin/generate exercise` to displaying informative progress messages:

```none
$ bin/generate isogram
Incremented tests version to 3
Updated version in example solution to 3
Generated tests for isogram
```

If you want it to be silent use the `--quiet` command line switch.

This patch removes the `--verbose` option and adds `--quiet`

### DIFF
--- a/lib/generator/command_line.rb
+++ b/lib/generator/command_line.rb
@@ -30,10 +30,10 @@ module Generator
     end
 
     def repository
-      if @options[:verbose]
-        LoggingRepository.new(paths: @paths, exercise_name: @options[:exercise_name], logger: logger)
-      else
+      if @options[:quiet]
         Repository.new(paths: @paths, exercise_name: @options[:exercise_name])
+      else
+        LoggingRepository.new(paths: @paths, exercise_name: @options[:exercise_name], logger: logger)
       end
     end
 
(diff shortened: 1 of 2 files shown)


kotp:
That is kind of sad, since most tools use the verbose flag... including Ruby itself.  `-v, --verbose   print version number, then turn on verbose mode`

Insti:
The current behavior is:
```none
$ bin/generate isogram
$
```
Which is not particularly friendly or helpful. Did it even do anything?

Insti:
Also does anyone want a `--quiet` mode that couldn't just do `bin/generate exericise > /dev/null`?

kotp:
`Generated tests for isogram` would be ok... and if there were a problem `Error generating tests for isogram`.

Or if it is particularly interesting, `Generate and incremented isogram` or `Generated without incrementing isogram`.

If you want it quiet you could do -v0, where -v1 would be normal (default) and -v2 would have the version numbers reporting as you have now.

The options would be:

`-v --verbose[n]  Verbosity of 0 is quiet, 1 is the default behavior, 2 gives test/exercise version information`

tommyschaefer:
I really like the idea of having a simple message that indicates something happened on a successful run. When generating tests, I often find myself with the "Did it even do anything?" feeling.

I also like @kotp's idea for `-v <level>`. This is how ruby's `-W` works as well.

Insti:
Thanks for the feedback, I'll try a version that keeps the verbose but adds a success message.

PR #535, opened 2017-01-12
bin/generate: Default display success message.

Insti:
The old behavior was:

```
$ bin/generate isogram
$
```

Which is not particularly friendly or helpful. Did it even do anything?

The new default behavior is to report success to stdout:
```
$ bin/generate isogram
Generated isogram tests version 3
$
```

The original behavior can be replicated by redirecting stdout to /dev/null
```
$ bin/generate isogram > /dev/null
$
```

The `-v` option still provides additional information.
```
$ bin/generate -v isogram
Incremented tests version to 4
Updated version in example solution to 4
Generated isogram tests version 4
```

### DIFF
--- a/lib/generator/repository.rb
+++ b/lib/generator/repository.rb
@@ -30,33 +30,33 @@ module Generator
         template: tests_template.to_s,
         values: template_values
       )
-      display_result
-    end
-
-    def display_result
-      $stdout.puts "Generated #{exercise_name} tests version #{version}"
-    end
-  end
-
-  # A repository that also logs its progress.
-  class LoggingRepository < Repository
-    def initialize(paths:, exercise_name:, logger:)
-      super(paths: paths, exercise_name: exercise_name)
-      @logger = logger
-    end
-
-    def update_tests_version
-      super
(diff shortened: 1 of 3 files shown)


lib/generator/repository.rb
```diff
@@ -30,6 +30,11 @@ def create_tests_file
         template: tests_template.to_s,
         values: template_values
       )
+      display_result
+    end
+
+    def display_result
+      $stdout.puts "Generated #{exercise_name} tests version #{version}"
```
tommyschaefer: Does this need to be `$stdout.puts` since `puts` writes to `STDOUT` by default?

Insti: It will work without it, but I consider using a global to be a smell and like to make it explicit that's what's happening.

This has inspired me to refactor to eliminate it.

Insti:
Eliminated the usage of `$stdout` within the `Repository` class by *always* using a `LoggingRepository` and adjusting the log levels appropriately.

Refactoring `LoggingRepository` to use composition rather than inheritance made the testing much nicer/easier.

PR #538, opened 2017-01-14
bin/generator: Remove version compensation.

Insti:
All the test templates were updating the version:
```
assert_equal <%= version.next %>, BookKeeping::VERSION
```

But the responsibility for mainlining the template version belongs with the generator, not the template.

So we update the template to use the version passed to it:
```
assert_equal <%= version %>, BookKeeping::VERSION
```

And once the templates have been updated, we can delete the [code required to compensate](https://github.com/exercism/xruby/compare/master...Insti:Remove_version_compensation?expand=1#diff-35b0b1d4ffa097f11ffe659d2ee3398d)  for the version incrementing.


Insti:
Superseded by:  #540

PR #540, opened 2017-01-14
bin/generator: Fix outstanding TODOs

Insti:
This patch addresses the outstanding TODOs in the recent generator refactor.

Changes all `exercise.tt` files:
* Remove compensation for version increment in templates.
* Rename `sha1` to `abbreviated_commit_hash`
* Standardise common test data reference comment.

Changes all `lib/*_cases.rb` files:
* move the exercise cases `require`s out of the generator.

I was initially planning on doing this as 3 Pull Requests, but the default git settings weren't quite clever enough to work out how to merge the independent changes without conflicts so I've applied them all in one PR. The individual commits show the independent steps.

### DIFF
--- a/test/generator/template_values_test.rb
+++ b/test/generator/template_values_test.rb
@@ -2,26 +2,26 @@ require_relative '../test_helper'
 
 module Generator
   class TestCasesValuesTest < Minitest::Test
-    def test_sha1
-      expected_sha1 = '1234567'
-      subject = TemplateValues.new(sha1: expected_sha1, version: nil, test_cases: nil)
-      assert_equal expected_sha1, subject.sha1
+    def test_abbreviated_commit_hash
+      expected_abbreviated_commit_hash = '1234567'
+      subject = TemplateValues.new(abbreviated_commit_hash: expected_abbreviated_commit_hash, version: nil, test_cases: nil)
+      assert_equal expected_abbreviated_commit_hash, subject.abbreviated_commit_hash
     end
 
     def test_version
       expected_version = '1234567'
-      subject = TemplateValues.new(version: expected_version, sha1: nil, test_cases: nil)
+      subject = TemplateValues.new(version: expected_version, abbreviated_commit_hash: nil, test_cases: nil)
       assert_equal expected_version, subject.version
     end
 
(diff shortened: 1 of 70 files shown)


test/generator/template_values_test.rb
```diff
@@ -2,26 +2,26 @@
...
     def test_test_cases
       expected_test_cases = 'should be TemplateValues class'
-      subject = TemplateValues.new(test_cases: expected_test_cases, sha1: nil, version: nil)
+      subject = TemplateValues.new(test_cases: expected_test_cases, abbreviated_commit_hash: nil, version: nil)
       assert_equal expected_test_cases, subject.test_cases
     end
 
     def test_get_binding
-      subject = TemplateValues.new(sha1: nil,version: nil, test_cases: nil)
+      subject = TemplateValues.new(abbreviated_commit_hash: nil,version: nil, test_cases: nil)
```
kotp: What's causing the lack of space in `nil,version`?

Insti: xruby using a `rubocop.yml` with all the useful cops disabled.

kotp: hehe, I think that is allowing it, rather than causing it.  I like adding in that one.  If you would like to run with a "custom" rubocop configuration, you can...

`rubocop -c ~/.sane_rubocop.yml`

Insti: Well then I guess what caused it was me not hitting the spacebar between `,` and `v`.

If it's important (which it seems to be, if you're noticing it in a code review.) it should be checked for by the xruby Rubocop configuration and not rely on a different unmaintained configuration file.

kotp: I am noticing it by sight, rather than by tool, they stand out to me.

The point of me mentioning the custom configuration is not as a solution to this, but in the use of that for individual contributors, for spotting the things that they care about... mentioning it here and hopefully having those things that are deemed important enough to come into the configuration here.

I like how this rubocop change happened here.  Naturally happening, and as needed.

Insti:
Rebased to resolve merge conflict.
This should be ready to merge now. 
These commits should NOT be squashed.

PR #543, opened 2017-01-16
say: add generator

### LINKED ISSUE #530: Add generator for "say"
The exercise `say` should use generated tests (#396).

Canonical data for this exercise has already been created in [exercism/x-common](https://github.com/exercism/x-common) and can be viewed [here](https://github.com/exercism/x-common/blob/master/exercises/say/canonical-data.json).

`Say` does not currently use bookkeeping, so you will need to add a version file at `exercises/say/.meta/.version` and add the following to `exercises/say/example.rb`:

```ruby
module BookKeeping
  VERSION = 1
end
```

To generate tests you will need to create a generator at `lib/say_cases.rb` and a template at `exercises/say/example.tt`.

All generators currently adhere to a common public interface, and must define the following three methods:

- `test_name` - Returns the name of the test (i.e `test_one_equals_one`)
- `workload` - Returns the main syntax for the test. This will vary depending on the test generator and its underlying implementation
- `skipped` - Returns skip syntax (i.e. `skip` or `# skip`)

The `wordy` exercise can be used as an example while building this generator:
- [Canonical Data](https://github.com/exercism/x-common/blob/master/exercises/wordy/canonical-data.json)
- [Tests](https://github.com/exercism/xruby/blob/master/exercises/wordy/wordy_test.rb)
- [Template](https://github.com/exercism/xruby/blob/master/exercises/wordy/example.tt)
- [Generator](https://github.com/exercism/xruby/blob/master/lib/wordy_cases.rb)

For additional information on generated tests, consult the [Generated Test Suites](https://github.com/exercism/xruby/blob/master/README.md#generated-test-suites) section of the README.

As always, if you get stuck, let us know and we'll be happy to help! :smile:

ajwann:
I have the generator working, and I think everything is good to go. Please let me know if there are any changes I need to make.

fixes #530


definemethod:
The build looks like it's faling because it doesn't like how the ```SayTest``` class is defined in exercises/say/say_test.rb. However, the ```WordyTest``` class is defined the same way in exercises/wordy/wordy_test.rb. Thoughts?

exercises/say/say_test.rb
```diff
@@ -1,128 +1,116 @@
...
   def test_twenty
     skip
-    # This really shouldn't be twenty-zero
-    assert_equal 'twenty', Say.new(20).in_english
+    question = '20'
+    assert_equal(twenty, SayProblem.new(question).answer)
   end
 
-  def test_twenty_two
+  def test_twenty-two
```
tommyschaefer: Hi there! Looks like the build is failing because this method (and a few others) have invalid names.

`test_twenty-two` rather than `test_twenty_two`

lib/say_cases.rb
```diff
@@ -0,0 +1,59 @@
...
+  class << self
+    def call(row)
+      row
+    end
+  end
+end
+
+SayCases = proc do |data|
+  JSON.parse(data)['cases'].map.with_index do |row, i|
+    SayCase.new(SayCase::PreProcessor.call(row).merge(index: i))
```
tommyschaefer: Because this exercise doesn't require preprocessing, the `PreProcessor` just returns the provided `row`.

You can remove the `PreProcessor` class and merge the index directly into `row`

exercises/say/say_test.rb
```diff
@@ -1,128 +1,116 @@
...
 require_relative 'say'
 
+# Test data version: e3bd4a2
 class SayTest < Minitest::Test
-  def test_0
-    assert_equal 'zero', Say.new(0).in_english
+  def test_zero
+    # skip
+    question = '0'
+    assert_equal(zero, SayProblem.new(question).answer)
```
tommyschaefer: It looks like the generated tests aren't using a string for an assertion. `zero` rather than `'zero'`

lib/say_cases.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  private
+
+  def indent(size, lines)
+    lines.lines.each_with_object('') { |line, obj| obj << ' ' * size + line }
+  end
+
+  def assertion
+    return error_assertion unless expected
+    return message_assertion if message
```
tommyschaefer: It looks like `message_assertion` isn't being used in this particular exercise.

This check and the method definition can be removed.

exercises/say/say_test.rb
definemethod: Ahh alrighty. I'll fix that.

exercises/say/say_test.rb
```diff
@@ -1,128 +1,116 @@
...
   end
 
   def test_one
     skip
-    assert_equal 'one', Say.new(1).in_english
+    question = '1'
+    assert_equal(one, SayProblem.new(question).answer)
```
tommyschaefer: It looks like the assertions are different from the original.

In this case, its of the form:
```ruby
assert_equal expected, SayProblem.new(input).answer
```

But the original is:
```ruby
assert_equal expected, Say.new(input).in_english
```

exercises/say/say_test.rb
```diff
@@ -1,128 +1,116 @@
...
+    question = '987654321123'
+    assert_equal(nine hundred eighty-seven billion six hundred fifty-four million three hundred twenty-one thousand one hundred twenty-three, SayProblem.new(question).answer)
   end
 
-  def test_1_billion
+  def test_numbers_below_zero_are_out_of_range
     skip
-    assert_equal 'one billion', Say.new(10**9).in_english
+    question = '-1'
+    assert_equal(-1, SayProblem.new(question).answer)
```
tommyschaefer: These tests that assert `-1` should actually be testing that an error is raised.

It looks like this is happening because the `error_assertion` isn't being used in your `SayCase` class.

lib/say_cases.rb
definemethod: Thanks for the sanity check. I should have removed it when I stripped down the ```call``` method.

definemethod:
I made the changes and generated another test. It looks like the expected values are being created as strings this time and the method names don't contain '-' or ',' anymore.

Let me know if you would see any more changes that need to be made, and I can also squash the commits down if you would like.

exercises/say/say_test.rb
```diff
@@ -1,128 +1,116 @@
...
+    # skip
+    question = '0'
+    assert_equal('zero', SayProblem.new(question).answer)
   end
 
   def test_one
     skip
-    assert_equal 'one', Say.new(1).in_english
+    question = '1'
+    assert_equal('one', SayProblem.new(question).answer)
```
tommyschaefer: It looks like the assertions are different from the original.

In this case, its of the form:

```ruby
assert_equal expected, SayProblem.new(input).answer
```

But the original is:

```ruby
assert_equal expected, Say.new(input).in_english
```

exercises/say/say_test.rb
```diff
@@ -1,128 +1,116 @@
...
+    question = '987654321123'
+    assert_equal('nine hundred eighty-seven billion six hundred fifty-four million three hundred twenty-one thousand one hundred twenty-three', SayProblem.new(question).answer)
   end
 
-  def test_1_billion
+  def test_numbers_below_zero_are_out_of_range
     skip
-    assert_equal 'one billion', Say.new(10**9).in_english
+    question = '-1'
+    assert_equal('-1', SayProblem.new(question).answer)
```
tommyschaefer: These tests that assert `-1` should actually be testing that an error is raised.

It looks like this is happening because the `error_assertion` method isn't being used in `SayCase`.

exercises/say/say_test.rb
```diff
@@ -1,128 +1,116 @@
...
-  def test_really_big_number
+  def test_numbers_above_999_999_999_999_are_out_of_range
     skip
-    expected = 'nine hundred eighty-seven billion '
-    expected << 'six hundred fifty-four million '
-    expected << 'three hundred twenty-one thousand '
-    expected << 'one hundred twenty-three'
-    assert_equal expected, Say.new(987_654_321_123).in_english
+    question = '1000000000000'
+    assert_equal('-1', SayProblem.new(question).answer)
```
tommyschaefer: Also asserts `-1` rather than raising an error. See above comment

exercises/say/say_test.rb
```diff
@@ -1,128 +1,116 @@
...
+    question = '1000000000'
+    assert_equal('one billion', SayProblem.new(question).answer)
   end
 
-  def test_1_million_2_thousand_345
+  def test_a_big_number
     skip
-    expected = 'one million two thousand three hundred forty-five'
-    assert_equal expected, Say.new(1_002_345).in_english
+    question = '987654321123'
```
tommyschaefer: It looks like the previous tests have the `question` as a number, rather than a string.

They also format them to make it a bit easier to read the larger numbers. For example: `987_654_321_123`

lib/say_cases.rb
```diff
@@ -0,0 +1,44 @@
+require 'pry'
```
tommyschaefer: Looks like this is what's causing the tests to fail in CI

definemethod: Fail on my part

exercises/say/say_test.rb
definemethod: I don't know of a method that takes an integer and underscore separates it's digits. Is there such a thing? I know that the number 100_000 will be treated the same ass 100000 though.

definemethod: Actually I was able to hack it by calling ```to_s.scan(/.{3}/).join('_')```.

Insti: You can copy one from here: https://github.com/exercism/xruby/blob/master/lib/grains_cases.rb#L16

definemethod: Nice! I never though to reverse the string.

definemethod:
Looks like I finally got the build green! I definitely put Travis CI though it's paces for this PR.

kotp:
@definemethod Go ahead and squash the commits, and do a force push on your branch to update this PR, and then a final review can be done... looking good!

lib/say_cases.rb
```diff
@@ -0,0 +1,46 @@
+class SayCase < OpenStruct
+  def test_name
+    'test_%s' % description.gsub(/[-,]/,'_').downcase.tr(' ', '_')
```
Insti: This is now failing CI checks due to us updating the Rubocop style checks.
There needs to be a space after the comma in this line.

lib/say_cases.rb
```diff
@@ -0,0 +1,46 @@
...
+  def error_assertion
+    [
+      'assert_raises ArgumentError do',
+      indent(2, 'Say.new(question).in_english'),
+      'end',
+    ].join("\n")
+  end
+
+  def underscore_format(number)
+    number.to_s.reverse.gsub(/...(?=.)/,'\&_').reverse
```
Insti: This line needs a space after the comma too.

You can fix both of these automatically by running `bundle exec rubocop -a`

lib/say_cases.rb
```diff
@@ -0,0 +1,46 @@
+class SayCase < OpenStruct
+  def test_name
+    'test_%s' % description.gsub(/[-,]/, '_').downcase.tr(' ', '_')
```
tommyschaefer: This can be simplified to:

```ruby
'test_%s' % description.tr(' ,-', '_').downcase
```

But I'm not sure it's too important

kotp: Please do @tommyschaefer and we can bring it on in.  (Better to fix the window now than to remember to fix it later.)

ajwann:
@tommyschaefer; I made the changes you suggested in 0ccebce. Thanks for your help and feedback on this PR.

It looks like there are other tests in the xruby repo that don't have generators. Would it be alright if I created them and put up some PR's?

tommyschaefer:
@ajwann Awesome! Thank you so much for all of your work on this!

In regards to other generators, absolutely! That would be fantastic! :+1:

@Insti @kotp I'm going to go ahead and merge this since @Insti and I approved this a while ago and it hasn't received any other feedback! Thank you both for helping me review this PR!

Issue #485, opened 2016-11-08
Generator improvements.


Insti:
@petertseng found some things that could be improved while he was building a generator:

From: https://github.com/exercism/xruby/pull/484#issuecomment-259102521

>  you need to touch .version or the generator won't run (increment_version requires that such a file exists). And you need to have an example file too. I tried to generate my tests before working on my example, but increment_version_in_example requires that the example actually exist.

Insti:
@bmulvihill and I are working on improving the generator process, so we'll take these into account.

bmulvihill:
Yeah I actually noticed this too, we could have a simple command to scaffold this stuff when creating a new generator.

ie 
`bin/generate new EXERCISE_NAME`

bmulvihill:
We might even want a different executable to handle something like this to prevent confusion, something like bin/scaffold EXERCISE_NAME

kotp:
Maybe, but look at adding the functionality to take an argument.  This seems so closely tied to generate that it does not make sense to have a new executable.

petertseng:
Another idea that could be considered if y'all really do not want a new command:

Make the two cases mentioned above tolerant of the files not existing.

* In the case of the `.version` file: Create it and write 1 into it.
* In the case of the example file: Either do nothing, or simply print a warning "example file doesn't exist" (since this case should *only* happen when writing the generator for the first time)

That said, something to scaffold the adding of a new exercise with generator seems useful - the lib file, the example.tt file, possibly appending it to config.json if possible (!). Back in the day it used to be more, you would have to add a bin/generate-#{slug}, thanks for getting rid of that step!

kotp:
Prefer to create .version on first use.  That is a great thing... No example file warning is good, as well, a first contributor would be lead to do the needful.

kotp:
Should now use the version number from canonical rather than the sha as is currently done in `bin/generator/git_command.rb`  Should add a `warn('Deprecated - use canonical_version instead.')` maybe?

hilary:
@kotp and I are working on a generator, and have noticed some pain points.

1. The documentation says a generator requires `workload`, however the method is actually `work_load`. 
1. It isn't at all clear how to construct `work_load`

Insti:
I think it should be `workload` it's one word.

Insti:
> The documentation says a generator requires workload, however the method is actually work_load.

What made you think this? Was it some documentation? Another test file?
I'm trying to work out what we need to do to can clear up any confusion.

[...]

PR #549, opened 2017-02-19
etl: added generator

ajwann:
## Description

Created a generator for the etl exercise. The generated test is functionally identical to the original, and the formatting of the hashes is as close as I could get it.

## Motivation and Context

Part of Epic #396

## How Has This Been Tested?

Generator has been ran and test file was successfully generated.

## Types of changes

- ~New feature (non-breaking change which adds functionality)~
- ~Breaking change (fix or feature that would cause existing functionality to change)~

## References and Closures
References #396

## Checklist:

- ~My change requires a change to the documentation.~
- ~My change relies on a pending issue/pull request~
- ~I have updated the documentation accordingly.~
- ~I have added tests to cover my changes.~

```bin/local-status-check``` returns no failures 
```rubocop -D``` returns no failures

_Admin edit: Using strikethrough rather than empty checkboxes, so it doesn't look like an incomplete todo list._

### CODE REVIEW

File: exercises/etl/etl_test.rb [hunk 1]
```diff
@@ -1,27 +1,36 @@
...
     expected = { 'a' => 1 }
 
     assert_equal expected, ETL.transform(old)
   end
 
-  def test_transform_more_values
+  def test_single_score_with_multiple_letters
     skip
-    old = { 1 => %w(A E I O U) }
-    expected = { 'a' => 1, 'e' => 1, 'i' => 1, 'o' => 1, 'u' => 1 }
```
ajwann: The previous test was arbitrarily placing key/value pairs on newlines.
I decided it was easiest to just check if there was more than one key/value pair in a hash 
(by matching on ',') and if so, then printing each key value pair on a separate line.

exercises/etl/etl_test.rb
```diff
@@ -32,27 +41,67 @@ def test_more_keys
...
+  def test_multiple_scores_with_differing_numbers_of_letters
     skip
     old = {
-      1 => %w(A E I O U L N R S T),
-      2 => %w(D G),
-      3 => %w(B C M P),
-      4 => %w(F H V W Y),
-      5 => %w(K),
-      8 => %w(J X),
-      10 => %w(Q Z)
```
ajwann: The previous test was also arbitrarily using the ```%w()``` syntax for arrays. Personally I think it's best if all tests just use one syntax or another, so I went with the standard bracket syntax. I can definitely switch everything to the ```%w()``` syntax though if that's the consensus.

tommyschaefer: > Personally I think it's best if all tests just use one syntax or another

Yes! Agreed :+1:

>  I can definitely switch everything to the %w() syntax though if that's the consensus.

Personally, I think the longer examples are a bit easier to read when using `%w()`, but I'm not sure if everyone else will agree with that. This is likely a question of preference, so a weigh in from @Insti and @kotp would be great. I don't think its much of a big deal though 😄

lib/etl_cases.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def assertion
+    "assert_equal expected, ETL.transform(old)"
+  end
+
+  def integerize_keys(input)
+    input.reduce({}) { |hash, (k, v)| hash[k.to_i] = v; hash }
+  end
+
+  def format_hash(hash)
```
tommyschaefer: I think this method can be simplified a bit. 

Try operating on the hash as a hash, rather than as a string.

exercises/etl/etl_test.rb
kotp: Unsure it was 'arbitrary' as it does save some typing... like 2/3rds for each single character "word", after the first element.  For a single element entry, it is identical in character length.  It was probably a convenience of typing, and something that is idiomatic (even encouraged in the style guide).

Consistency is good though.

lib/etl_cases.rb
ajwann: I created this method to format the hash neatly across multiple lines. Without it, each hash is printed on one line, and those lines are really long. In order to manipulate the hash across these lines I has to convert it to a string to properly add newlines and indentation. I'm not sure how I can achieve the same result without converting it to a string.

lib/etl_cases.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def skipped
+    index.zero? ? '# skip' : 'skip'
+  end
+
+  private
+
+  def indent(size, lines)
+    lines.lines.each_with_object('') { |line, obj| obj << ' ' * size + line }
```
ajwann: Agreed, this looks really strange. I'll fix it.

exercises/etl/etl_test.rb
ajwann: Personally I think that new Ruby users coming from other languages would be able to recognize and understand the bracket syntax faster than the ```%w()``` syntax. But that's just my opinion and I'm happy to change the generator to use only the ```%w()``` syntax.

kotp:
@tommyschaefer Can you pull this locally, and run a `bundle exec generate -f etl` and see if `exercises/etl/etl_test.rb` changes in the `test_bookkeeping` method as far as version goes?

This seems to be unique here, as it does not seem to do the same in Hamming, for example.  It freezes like it should.

lib/etl_cases.rb
kotp: @ajwann I think @tommyschaefer is hinting at something like:

```ruby
  middle = hash.each_pair.with_object('') do |(k, v), obj|
    obj << " :#{k} => #{v},\n "
  end
  " {\n#{middle}}"
```

ajwann:
@kotp @tommyschaefer; sorry to disappear for so long, it's been a busy month. I operated on the hash using ```each_with_object``` as @kotp suggested, and it made life a lot easier.

kotp:
@ajwann It looks like your x-common is now out  of date.  The generator will not build this on the current canonical data.  Can you confirm and fix?

kotp:
@hilary Can you review this, and let us know what needs to be done here?

hilary:
@kotp will do!

hilary:
Minimal required work before merging:

1. first and most important, this branch needs to be rebased onto master.
1. Remove the `EtlCases` loop from `lib/etl_cases.rb`

hilary:
PR carried forward in https://github.com/exercism/xruby/pull/575

hilary:
Closing this PR now that we've carried the work forward. Let's get this generator in :)

@ajwann I've been working on reducing the effort to add a generator. When you recover from this one and get a chance to take a swing at another, I would love to hear whether you think the changes have actually reduced the effort or not.

ajwann:
Thank you @hilary! I'll be starting on another generator soon, I'll let you know how it goes.

hilary:
@ajwann I think the best way to see what is needed to write a generator is to look at the xruby README, which we're keeping up to date.

If you'd like to see what we're planning, get involved in the discussion, etc., we have an issue for the generator changes: exercism/xruby#485

PR #550, opened 2017-02-20
hello world tests via updated x-common data

kotp:
## Description
Uses the x-common data to simplify the hello world introductory exercise and tests.

## Motivation and Context
re: exercism/x-common#520

- [X] All new and existing tests passed.


exercises/hello-world/hello_world_test.rb
```diff
@@ -12,23 +12,11 @@
   exit 1
 end
 
-# Test data version:
-# deb225e Implement canonical dataset for scrabble-score problem (#255)
-
+# Common test data version: 35aa16b
 class HelloWorldTest < Minitest::Test
-  def test_no_name
+  def test_Say_Hi!
```
tommyschaefer: Should this maybe be downcased to `test_say_hi`?

exercises/hello-world/.meta/.version
```diff
@@ -1 +1 @@
-1
\ No newline at end of file
+2
```
tommyschaefer: Does the version need to be bumped since the old implementation is still valid?

kotp: Yes, I believe so.

kotp: Clarification on why I believe it needs to be incremented is that old version would take an argument, this new version will not accept an argument.

kotp:
nb: Generator still uses git commit sha for version, perhaps should use canonical data version instead.  @insti @tommyschaefer ?

Insti:
Generator should be using the sha1 from the canonical data repository though.

exercises/hello-world/.meta/.version
Insti: I like the version increment.

Insti:
> Generator should be using the sha1 from the canonical data repository already.

Now I've caught up a bit more I see you mean the version property that is now present in the canonical-data.json files. I'll need to think about the implications of this. Will continue the discussion in: https://github.com/exercism/xruby/issues/485

PR #553, opened 2017-03-09
Implement generator for beer-song

kytrinyx:
I've submitted a pull request to update the canonical data for the beer-song exercise (https://github.com/exercism/x-common/pull/661). In order to make it easier to pull in those changes once the PR has been merged, I've created a generator for the beer-song exercise.

In addition to implementing the logic for the generator, I ran the `bin/generate beer-song` command to create a fresh implementation.

The structure of the canonical data is a bit different from the implementation that the Ruby track had:

*Before*:

```
BeerSong.new.verse(n)
BeerSong.new.verses(n, m)
BeerSong.new.lyrics
```

*After*:

```
BeerSong.new.verse(n)
BeerSong.new.verses(n, m)
```

In the old version, `lyrics` was a simple indirection, calling `verses` with the full range.

This is a breaking change, and the book-keeping has been updated to reflect this.

As I implemented the generator, I discovered that the documentation was outdated. This has been updated in https://github.com/exercism/xruby/pull/552


kotp:
This may need to be updated after the exercism/x-common#661 is brought in.

kytrinyx:
I've updated this per the newest canonical data.

exercises/beer-song/beer_song_test.rb
```diff
@@ -1,69 +1,86 @@
...
 require_relative 'beer_song'
 
+# Common test data version: 9f3d48a
 class BeerSongTest < Minitest::Test
-  def test_the_first_verse
-    expected = "99 bottles of beer on the wall, 99 bottles of beer.\n" \
-      "Take one down and pass it around, 98 bottles of beer on the wall.\n"
+  def test_first_generic_verse
+    # skip
+    expected = <<-TEXT
```
kotp: Using the `<<-TEXT` means that we can have the ending delimiter indented.

At the moment the ending delimiter is on the 1 column, so this is not needed.  See comment regarding the generator for note.

exercises/beer-song/example.tt
```diff
@@ -0,0 +1,22 @@
...
+require 'minitest/autorun'
+require_relative 'beer_song'
+
+# Common test data version: <%= abbreviated_commit_hash %>
+class BeerSongTest < Minitest::Test<% test_cases.each do |test_case| %>
+  def <%= test_case.test_name %>
+    <%= test_case.skipped %>
+    expected = <<-TEXT
+<%= test_case.expected %>
+TEXT
```
kotp: Because of the `<<-TEXT` we could indent the  ending delimiter to the level of line 11.


```ruby
    <%= test_case.skipped %>
    expected = <<-TEXT
<%= test_case.expected %>
    TEXT
    assert_equal expected, <%= test_case.workload %>
  end
<% end %>
```

The other choice would be to remove the `minus` before the delimiter and leave the ending delimiter as it is.

kytrinyx: Oh, right. Yepp. I don't have strong feelings either way, but I think I'll indent the ending delimiter.

kytrinyx:
I've updated the indentation of the closing token for the heredoc.

I'm using the `property` to decide the name of the function, so as long as that doesn't change over in our canonical data schema discussion, I think we're good here.

behrtam:
I'm just curious ... wouldn't it make more sense now that the canonical test data has a real version ([beer-song/canonical-data.json#L3](https://github.com/exercism/x-common/blob/master/exercises/beer-song/canonical-data.json#L3)) to use that instead of the sha-1 hash of the commit (`# Common test data version: 9f3d48a`)?

kotp:
I think the merge was done before the canonical was finished... but I think using that canonical version now makes sense to reference.

kytrinyx:
Yeah, the canonical version makes much more sense. I didn't think of it.

PR #558, opened 2017-04-16
Standardize hello-world generator.

Insti:
Update the hello-world generator to use the standard generator method names.

## Description
Use the standard `workload` and `skipped` methods rather than the more ambiguous and template dependent `do` and `skipped?` methods.

## How Has This Been Tested?
Regenerated `hello-world_test.rb` which was almost identical to the original version.
### (Minor) Differences
* Change from sigle quotes to double quotes due to use of `inspect`
* Addition of commented out `#skip` call which now matches other generated tests.


kotp:
Thank you @Insti

PR #560, opened 2017-04-21
Update test generator documentation.

Insti:
Update the readme to better reflect the current best practices for test
generators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,8 @@ class ProblemNameCase < OpenStruct
   end
 
   def workload
-    # implement main logic of test here
+    # Example workload:
+    "assert #{expected.inspect}, Problem.call(#{input.inspect})"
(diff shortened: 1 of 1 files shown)


hilary:
I would also suggest using a consistent name for the name method (say that 3 times fast!)

The sample `lib/$PROBLEM_cases.rb` uses `test_name`, as does the list below it, but the sample `example.tt` erb reads `<%= test_case.name %>`. Doh!

A quick scan of the existing generators shows that the consensus name is `name`. 😄

Insti:
Hmm `name` might be the wrong thing to use due to potential conflicts with the json file.
I'll have to check.

Insti:
Addresses some of the issues discussed in: https://github.com/exercism/xruby/issues/485

PR #563, opened 2017-04-21
pig-latin: generate tests using canonical data

hilary:
## Description
* lib/pig_latin_cases.rb
* exercises/pig-latin/example.tt
* added bookkeeping module to exercises/pig-latin/example.rb

## Motivation and Context
* Generate tests with canonical data exercism/x-common#625

## How Has This Been Tested?
`rake test`


exercises/pig-latin/example.tt
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env ruby
+gem 'minitest', '>= 5.0.0'
+require 'minitest/autorun'
+require_relative 'pig_latin'
+
+# Common test data version: <%= abbreviated_commit_hash %>
+class PigLatinTest < Minitest::Test<% test_cases.each do |test_case| %>
```
Insti: Add a newline here to split the class definition from the template code.

lib/pig_latin_cases.rb
```diff
@@ -0,0 +1,22 @@
...
+  def skipped
+    index.zero? ? '# skip' : 'skip'
+  end
+end
+
+PigLatinCases = proc do |data|
+  JSON.parse(data)['cases'].
+    flat_map {|section| section['cases'] }.
+    each_with_index.
+    reduce([]) {|cases, (test, i)| cases << PigLatinCase.new(test.merge('index' => i)) }
```
Insti: This is a bit of a mega code pipeline.
I wonder is it possible to add some intention revealing variable names into the mix.

exercises/pig-latin/example.tt
hilary: Blech. ERB. Adding a newline here causes formatting problems in the generated test file.

kotp: I believe you can use a backslash to avoid that?  There is also something in ERB itself that lets you ignore the new line.

Minor thing though though.

The generated file being exemplar is more important.

kotp:
Thank you @hilary !

PR #564, opened 2017-04-21
ocr-numbers: generate tests from canonical data

hilary:
## Description
* lib/ocr_numbers_cases.rb
* exercises/ocr_numbers/example.tt
* exercises/ocr_numbers/example.rb (had to update it to handle argument errors and the old version was ... wordy)

## Motivation and Context
* Generate tests with canonical data exercism/x-common#625

## How Has This Been Tested?
`rake test`

## Types of changes
Just a note: version 0 of `ocr-numbers` used non-standard naming for the tests and sample solution; both used `OCR` rather than camel casing the problem name. I standardized the naming.

As with pig-latin, I changed the tests so a class is not required.


exercises/ocr-numbers/ocr_numbers_test.rb
```diff
@@ -3,187 +3,112 @@
...
-| |
-|_|
-   
-    NUMBER
-    assert_equal '0', OCR.new(text).convert
+# Common test data version: 80782b6
+class OcrNumbersTest < Minitest::Test
+  def test_recognizes_0
+    # skip
+    assert_equal "0", OcrNumbers.convert(" _ \n| |\n|_|\n   ")
```
Insti: It would be nice to do something like in the previous version that human readably formatted the input data, but that is not a blocker for this PR.

Insti:
Thanks @hilary :heart:

PR #565, opened 2017-04-22
Linked list: add tests for emptied list.

### LINKED ISSUE #562: Linked List exercise tests for emptied list
The tests for the [Linked List exercise](https://github.com/exercism/xruby/tree/master/exercises/linked-list) don't test if, when popping or shifting should leave the list empty, the reverse end of the list is cleared/nulled correctly. I.e. when the list is empty, then both the head and the tail ends should be empty/null.

Forgetting to do that is an easy mistake to make when implementing a doubly linked list (i.e. I made it), and the current test suite doesn't cover that yet.

## Possible Solution

Since the [instructions](https://github.com/exercism/x-common/blob/master/exercises/linked-list/description.md) specifically say that _"`pop` or `shift` will never be called on an 
empty list"_, a possible solution without defining the behaviour for error conditions could be:

```ruby
def test_pop_to_empty
  deque = Deque.new
  deque.push(10)
  assert_equal 10, deque.pop
  deque.push(20)
  assert_equal 20, deque.shift
end

def test_shift_to_empty
  deque = Deque.new
  deque.unshift(10)
  assert_equal 10, deque.shift
  deque.unshift(20)
  assert_equal 20, deque.pop
end
```

What do you think?

kewlar:
Adds tests to test if the opposite end of the Deque is nulled correctly when popping/shifting leaves the list empty.

---

The tests for the [Linked List exercise](https://github.com/exercism/xruby/tree/master/exercises/linked-list) don't yet test if, when popping or shifting should leave the list empty, the reverse end of the list is cleared/nulled correctly. When the list is empty, both the head and the tail ends should be empty/null.

Forgetting to do that is an easy mistake to make when implementing a doubly linked list, and the current test suite doesn't cover that yet.

This PR adds the tests for that, without the need to define the behaviour for error conditions (`pop` or `shift` on an empty list).

---

Closes #562.

### DIFF
--- a/exercises/linked-list/linked_list_test.rb
+++ b/exercises/linked-list/linked_list_test.rb
@@ -53,4 +53,20 @@ class DequeTest < Minitest::Test
     assert_equal 50, deque.pop
     assert_equal 30, deque.shift
   end
+
+  def test_pop_to_empty
+    deque = Deque.new
+    deque.push(10)
+    assert_equal 10, deque.pop
+    deque.push(20)
+    assert_equal 20, deque.shift
+  end
+
+  def test_shift_to_empty
+    deque = Deque.new
+    deque.unshift(10)
+    assert_equal 10, deque.shift
+    deque.unshift(20)
+    assert_equal 20, deque.pop
+  end
 end


Insti:
Nice, thanks.

Insti:
Thanks @kewlar :heart: 

It will be a day or two before the changes appear on the live site,  there is a manual release process, but they will get there eventually.

kewlar:
Awesome :v:

PR #566, opened 2017-04-22
DRY test generators

hilary:
## Motivation and Context
The test generators have a lot of duplicate code. Do we need to discuss why that's not good? I decided to take a pass at DRYing up the test generators before moving on to writing a test generator scaffold generator...

Design principles:
* backwards compatible (while most of the generators I saw don't currently work 😭 ), these change can't break any that do.
* facilitate, don't dictate. Some test generators have different needs. These changes can't make those test generators a nightmare to write.
* take small, bite-sized steps that can be merged one at a time. 

This effort:
* is part of exercism/xruby#485
* is meant to facilitate ruby getting to and staying up to date with canonical data exercism/x-common#625
* takes direction from @Insti 's test generator master design https://github.com/Insti/xruby/pull/1

## Description
I introduced a base class, `ExerciseCase` into `lib/exercise_cases.rb` from which `ProblemNameCase` can inherit. `ExerciseCase` provides public methods `name` and `skipped` as well as a variety of protected helper methods.

I converted three generators I wrote, and one I didn't write, to use `ExerciseCase`. The generated test files remain (almost) the same while the generators exhibit far more consistency. (The only changes to the test files are some small changes in how the descriptions are processed to produce the test names.

I updated the README to reflect the changed workflow.

## How Has This Been Tested?
I can regenerate the tests with no changes except for the small changes noted above.
The tests pass under `rake test`.

## Next Steps
* There's no need for the person writing a generator to write the `ProblemNameCases` logic. We can inspect and extract the cases from the JSON programmatically.
* Fall back to a default template for exercises which don't provide a customized template. The four exercises I have converted thus far can all use the same template and I'm sure that many, many other exercises can as well. I would much rather not generate a template by default, as we should be actively discouraging putting logic into views.


lib/exercise_cases.rb
```diff
@@ -1,2 +1,30 @@
 require 'ostruct'
 require 'json'
+
+class ExerciseCase < OpenStruct
+  def name
+    'test_%s' % description.gsub(/[^\w ?!]/, '').gsub(/[- ]/, '_')
```
kotp: This change causes the test names to change by not presenting hyphenated words as individual words, but rather as shown in this diff:

```plain
-  def test_large_distance_in_off_by_one_strand
+  def test_large_distance_in_offbyone_strand
```

I would like to place the change back in, due to things like "non_unique" being acceptable and readable, even if it technically isn't a separate word, and to avoid "off-by-one" becoming "offbyone".

kotp:
The design principles are very much in line with what we want to do, I think.

> while most of the generators I saw don't currently work

These places would definitely benefit from being named.

Insti:
Firstly: Awesome PR description :heart: thanks.

> Design principles:
>   backwards compatible

This is not a requirement, feel free to break as much stuff as you like.
There are a significant number of already broken generators (as you have discovered.)
Once we work out what the "true" generation process is we can update all the old generators.

> I introduced a base class, ExerciseCase into `lib/exercise_cases.rb` 

Excellent.

> I converted [four] generators [..] to use ExerciseCase. 

Excellent.

> I updated the README to reflect the changed workflow.

Excellent.

lib/exercise_cases.rb
```diff
@@ -1,2 +1,30 @@
 require 'ostruct'
```
Insti: This file should be in `lib/generator'

lib/exercise_cases.rb
```diff
@@ -1,2 +1,30 @@
...
 require 'json'
+
+class ExerciseCase < OpenStruct
```
Insti: Not an issue for this PR, but `ExerciseCase` should not be an `OpenStruct` and the canonical data should be kept separate from the methods of this class. 

See: [my imagined version of `ExerciseCase`](https://github.com/Insti/xruby/blob/1e54c30f13cbe584a37a0b3bcd636eb17c8b7f39/lib/exercise_testcase.rb) for an example.

lib/luhn_cases.rb
```diff
@@ -1,22 +1,8 @@
...
-
-  def skipped
-    index.zero? ? '# skip' : 'skip'
-  end
-
-  private
-
-  def assertion
-    expected ? 'assert' : 'refute'
+    %Q(#{assert_or_refute} Luhn.valid?(#{input.inspect}))
```
Insti: I like that all these need to do now is implement `workload`.

Insti:
All these TODOs probably deserve their own issue.

## TODO
>    extract the cases from the JSON programmatically.
>    Fall back to a default template for exercises which don't provide a customized template. 

Yes and Yes.

Now some of the things from the grand design in my head:

Note: "problem" in all these examples is the variable exercise name.

* Generator problem_cases.rb should live in `exercises/problem/.meta/generator/` 
* Templates should live in `problem/.meta/generator/` be called `test_file_template.erb` (or similar). Once they are under .meta they don't need to contain the word `example`
* There definitely should be a default `test_file_template.erb` - probably in `/lib/generator`
* ExerciseCase should not be an OpenStruct and the canonical data should be kept separate from the methods of this class.
* canonical_data version should be added to the test file - probably immediately to the left of the abbreviated commit hash.
* `.version` should be updated to be the canonical_data version (e.g. `2.1.0`) plus an x-ruby version. (`3`) e.g.: `2.1.0.3`

Some of this will require changes to the bin/generator code.


By creating the new `modern` generators under .meta/generator it will be easy to track which exercises have been upgraded as the `lib/problem_cases.rb` are deleted with the end goal being all generators living under `.meta/generator` and lib being mostly empty.

Everything needed to create a generator for an exercise will be in `exercises/problem/.meta/generator/`

Insti:
For reference, here is my proof of concept for this from a few months ago: https://github.com/Insti/xruby/pull/1

README.md
```diff
@@ -114,54 +114,50 @@ the exercise, which lives in the x-common repository.
...
-class ProblemNameCase < OpenStruct
-  def name
-    'test_%s' % description.gsub(/[ -]/, '_')
-  end
+class ProblemNameCase < ExerciseCase
 
   def workload
     # Example workload:
-    "assert #{expected.inspect}, Problem.call(#{input.inspect})"
+    "#{assert_or_refute} Problem.call(#{input.inspect})"
```
Insti: Can this be just `assert` it doesn't REALLY matter if the final test file contains `refute`

lib/exercise_cases.rb
```diff
@@ -1,2 +1,30 @@
...
+    index.zero? ? '# skip' : 'skip'
+  end
+
+  protected
+
+  def raises_error?
+    expected.to_i == -1
+  end
+
+  def assert_or_refute
```
Insti: Can this be just `assert` it doesn't REALLY matter if the final test file contains `refute`

Insti: Future:
It may be possible to add logic into the `assert` method that works out the correct assert to use including `assert_equal` or `assert_error`

Insti:
I think that's all for now.
I'm happy for this PR to be merged as-is, and have further work be continued in other ones.

lib/exercise_cases.rb
hilary: Will do (in the morning!)

lib/exercise_cases.rb
hilary: Indeed! I'll go ahead and move it (also in the morning).

lib/exercise_cases.rb
kotp: I think it is a separate commit, so it may be a simple process.  That was a great idea to separate it out like that!

Insti: I'm not sure if this is premature:

We will need a `underscore_format(string)`\* method that we can have a bunch of unit tests for so we can keep track of all the weird cases we encounter.

So this line becomes:
`'test_' + underscore_format(description)`

\* name and whether it should be a `String` refinement TBD.

exercises/ocr-numbers/ocr_numbers_test.rb
```diff
@@ -5,87 +5,87 @@
...
   end
 
-  def test_recognizes_1
+  def test_Recognizes_1
     skip
     assert_equal "1", OcrNumbers.convert("   \n  |\n  |\n   ")
   end
 
-  def test_unreadable_but_correctly_sized_inputs_return_?
+  def test_Unreadable_but_correctly_sized_inputs_return_?
```
tommyschaefer: I think `ExerciseCase#name` should probably also downcase the description.

Insti: What to do about punctuation characters is another issue.

But this one is probably a job for a pre-processor, since it's problem dependent.
`s/"?"/question mark string/`

hilary: Heh. I had `downcase` in there at one point, then took it out. Shoulda trusted my first instincts on that one 😄

lib/pig_latin_cases.rb
```diff
@@ -1,5 +1,3 @@
-require 'exercise_cases'
```
Insti: Not entirely sure I agree with removing this line.
Sure `require_all` is awesome, but there's only one dependency and it's nice to be able to see where `ExerciseCase` comes from.

test/generator/string/underscore_test.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class UnderscoreTest < MiniTest::Test
+  using Underscore
+
+  # at present, we're downcasing everything, including TLAs
+  def test_mixed_case
+    assert_equal 'A string with TLA'.underscore, 'a_string_with_tla'
+  end
+
+  def hyphenated_text
```
Insti: `def test_hyphenated_text`

lib/luhn_cases.rb
```diff
@@ -1,8 +1,6 @@
-require 'exercise_cases'
-
 class LuhnCase < ExerciseCase
   def workload
-    %Q(#{assert_or_refute} Luhn.valid?(#{input.inspect}))
+    "#{assert} Luhn.valid?(#{input.inspect})"
```
Insti: It's a good idea to keep your commits atomic. 
This has nothing to do with underscores and should be in its own commit.

It is a discipline that needs to be trained. Commit early, commit often.

lib/ocr_numbers_cases.rb
```diff
@@ -1,5 +1,3 @@
-require 'exercise_cases'
```
Insti: Should have been another separate commit.

lib/pig_latin_cases.rb
kotp: And it simply doesn't break anything... if that dependency goes away, no problem, it is a documentation value at this point.

exercises/hamming/example.tt
```diff
@@ -4,14 +4,16 @@ require 'minitest/autorun'
...
 # Common test data version: <%= abbreviated_commit_hash %>
-class HammingTest < Minitest::Test<% test_cases.each do |test_case| %>
+class HammingTest < Minitest::Test
+<% test_cases.each do |test_case| %>
   def <%= test_case.name %>
-    <%= test_case.skipped %><% if test_case.raises_error? %>
-    assert_raises(ArgumentError) { <%= test_case.workload %> }<% else %>
-    assert_equal <%= test_case.expected %>, <%= test_case.workload %><% end %>
+    <%= test_case.skipped %>
+    <%= test_case.workload %>
```
kotp: Very nice.

lib/ocr_numbers_cases.rb
```diff
@@ -1,22 +1,12 @@
...
+class OcrNumbersCase < ExerciseCase
   def workload
-    if expected == -1
-      "assert_raises(ArgumentError) { #{test_case} }"
+    if raises_error?
+      assert_raises(ArgumentError) { test_case }
     else
-      "assert_equal #{expected.inspect}, #{test_case}"
+      assert_equal { test_case }
     end
```
kotp: Definitely a better place for this.

hilary:
Putting this here so we don't forget:

We currently create `exercises/$PROBLEM/.meta/.version` and fill it with `1` if it doesn't exist. Do we create the `.meta` directory if needed? If not, we should. Once we do, we should remove the instruction to do so from the README.

PR #567, opened 2017-04-24
generators: convert generators to use auto extract

hilary:
## Motivation and Context
I converted all generators to use auto extraction (exercism/xruby#568) so we don't need to maintain support for the old `proc` extraction technique.

## Description
(Commits https://github.com/exercism/xruby/pull/567/commits/19d34af4b2a278799ae2e304debd836b5020e950 and https://github.com/exercism/xruby/pull/567/commits/e839613707269888a2d64578b9074b352eead781 are part of exercism/xruby#568)

* generators I wrote myself:
  * Luhn
  * OCR numbers
  * Pig Latin
* others
  * Acronym (updated)
  * All your base
  * Alphametics (generator was broken)
  * Anagram
  * Binary (generator was broken)
  * Beer Song
  * Bowling (generator was broken)
  * Bracket Push
  * Clock
  * Connect
  * Custom Set
  * Difference of squares
  * Dominoes
  * Gigasecond (updated, generator was broken)
  * Grains (generator was broken)
  * Hamming
  * Hello World
  * Isogram
  * Largest Series Product (updated)
  * Leap (updated)
  * Nth prime
  * Pangram
  * Queen attack (generator was broken)
  * Raindrops
  * RNA transcription
  * Roman numerals
  * Run length encoding (updated, generator was broken)
  * Say
  * Sieve
  * Tournament (generator was broken)
  * Transpose
  * Triangle (generator was broken)
  * Two Bucket (generator was broken)
  * Word Count
  * Wordy

## How Has This Been Tested?
In all cases, `rake test` passes.
In all cases, regenerating the tests produces either the same tests (when the data has not changed) or the desired tests (for updated data)

## Types of changes
Updates and fixes

## References and Closures
* exercism/xruby#566
* exercism/xruby#568
* exercism/xruby#485
* https://github.com/Insti/xruby/pull/1

## Checklist:
- ~~My change requires a change to the documentation~~
- [x] My change relies on a pending issue/pull request
- ~~I have updated the documentation accordingly.~~
- ~~I have added tests to cover my changes.~~


hilary:
Broken generators not yet fixed (tracking here):
* bowling
* grains
* tournament
* triangle
* two-bucket

lib/pangram_cases.rb
```diff
@@ -1,15 +1,11 @@
...
-    'test_%s' % description.downcase.tr_s(" -'", '_').sub(/_$/, '')
-  end
-
+class PangramCase < ExerciseCase
   def workload
     [
     "phrase = '#{input}'",
     "    result = Pangram.pangram?(phrase)",
-    "    #{assertion} result, \"#{message}\""
+    "    #{assert} result, \"#{message}\""
```
Insti: Can do `message.inspect` here and get rid of the `\"`s

exercises/acronym/acronym_test.rb
```diff
@@ -54,6 +59,6 @@ def test_hyphenated
 
   def test_bookkeeping
     skip
-    assert_equal 2, BookKeeping::VERSION
+    assert_equal 3, BookKeeping::VERSION
```
Insti: Not sure the changes here warrant a version bump.

hilary: It added a test, hence the version bump.

exercises/anagram/anagram_test.rb
```diff
@@ -137,6 +104,6 @@ def test_capital_word_is_not_own_anagram
 
   def test_bookkeeping
     skip
-    assert_equal 2, BookKeeping::VERSION
+    assert_equal 3, BookKeeping::VERSION
```
Insti: unnecessary version bump

exercises/acronym/acronym_test.rb
Insti: ok, I missed that

lib/pangram_cases.rb
hilary: Doesn't work here (at least straight substitution doesn't). The interpolation in `message` would also need adjusting.

exercises/anagram/anagram_test.rb
hilary: There are changed tests. Shouldn't that mean the version is updated?

lib/pangram_cases.rb
Insti: you're right. :+1:

exercises/anagram/anagram_test.rb
Insti: Which tests changed?

The commit that does the version update should just include the bits of the file that caused that version upgrade.

These are being hidden by the rest of the noise in the commit due to formatting changes.

exercises/anagram/anagram_test.rb
```diff
@@ -3,119 +3,86 @@
...
+    assert_equal [], Anagram.new('galea').match(["eagle"]).sort
   end
 
-  def test_detects_multiple_anagrams
+  def test_detects_two_anagrams
     skip
-    detector = Anagram.new('master')
-    anagrams = detector.match(["stream", "pigeon", "maters"])
-    assert_equal ["maters", "stream"], anagrams.sort
+    assert_equal ["maters", "stream"], Anagram.new('master').match(["stream", "pigeon", "maters"]).sort
```
Insti: I strongly prefer the previous 3-line version of these tests.

exercises/anagram/anagram_test.rb
hilary: Ah, you are right. In this case, just the names changed.

exercises/anagram/anagram_test.rb
hilary: Interesting. I strongly prefer the one line version. With the three line version, I have to mentally assemble the test; with the one line version, it is all right there.

I go to multiple lines on tests that are too complex to mentally wrangle by reading a single line (like the tests I wrote for `Generator::CaseValues`. But if I can do it in one or two lines, I prefer it.

How would you feel about a two line version?

    anagrams = Anagram.new('master').match(["stream", "pigeon", "maters"])
    assert_equal ["maters", "stream"], anagrams.sort

Insti: It is worse than both 3 and 1 line versions.

hilary: ROFL! 

The usual 3 line pattern for tests is 

    expected = ...
    actual = ...
    assert_equal expected, actual

These 3 liners cause cognitive friction for me because they are 3 liners that violate that pattern.

Insti: 4 would probably be best since you can use another line for `expected`

The main issue here is that we want to prioritize educational value over expert comprehension. 

1 line:
```ruby
assert_equal ["gallery", "largely", "regally"], Anagram.new('allergy').match(["gallery", "ballerina", "regally", "clergy", "largely", "leading"]).sort
```
At my day job I would be fine with the 1 line version as everyone knows the test framework and can maintain a consistent test style, and we are not trying to work out how to implement a solution by looking at the tests.

On Exercism I want to make sure that things are clear to people who may not be as familiar with Minitest and even Ruby. 

Multi line:
```ruby
    detector = Anagram.new('allergy')
    anagrams = detector.match(["gallery", "ballerina", "regally", "clergy", "largely", "leading"])
    expected = ["gallery", "largely", "regally"]
    assert_equal expected, anagrams.sort
```

The multi line version uses variable naming to reveals the intention of the individual elements of the one line assertion. Note that the values used for `detector`,`anagrams` and `expected` are all vary by test.
This provides the information one needs to derive the code to pass the test in a clear and consistent manner.

hilary: Makes sense to me! I'll change it.

Insti:
Big thanks @hilary for all your work on this! :heart: :star: :+1:

PR #568, opened 2017-04-24
generators: Auto extract cases

hilary:
## Motivation and Context
While working on exercism/xruby#566, I noticed that a number of the existing generators would not run because the loop which extracts cases from the data did not match the current structure. I also noticed a lot of consistency in the loops for both failing and successful generators. Automatically extracting cases from the JSON feels like a big win.

## Description
Added `Generator::CaseValues` (in `lib/generator/case_values.rb`, which extracts the cases from the canonical data. It uses a custom extractor if such an extractor exists, otherwise it extracts the cases automagically.

## How Has This Been Tested?
* I wrote tests for `Generator::CaseValues` for both custom extractor and not custom extractor and for both simple and complex canonical data. 
* I wrote a test for a public helper method I introduced to `Generator::Files::GeneratorCases`. 
* The existing developer tests pass.

## Types of changes
No breaking changes.
New feature: auto extraction of cases.

## References and Closures
* exercism/xruby#566
* exercism/xruby#485
* https://github.com/Insti/xruby/pull/1

## Checklist:
- [x] My change requires a change to the documentation.
- ~~My change relies on a pending issue/pull request~~
- [x] I have updated the documentation accordingly.


lib/generator/template_values.rb
```diff
@@ -31,8 +31,40 @@ def cases_require_name
       Files::GeneratorCases.filename(exercise_name)
     end
 
+    def test_cases
```
Insti: There is a LOT of untested logic here. That scares me.

lib/generator/template_values.rb
```diff
@@ -31,8 +31,40 @@ def cases_require_name
...
+    end
+
+    def extract_test_cases(data = JSON.parse(canonical_data.to_s)['cases'])
+      data.flat_map do |entry|
+        entry.key?('cases') ? extract_test_cases(entry['cases']) : entry
+      end
+    end
+
+    def test_cases_proc?
+      Object.const_defined?(test_cases_procname)
```
Insti: Use the method you already have: `test_cases_proc.exists?`

test/generator/files/track_files_test.rb
```diff
@@ -25,6 +25,10 @@ def test_filename
         assert_equal 'two_parter_cases', GeneratorCases.filename(exercise_name)
       end
 
+      def test_class_name
+        assert_equal 'TwoParterCase', GeneratorCases.class_name('two-parter')
+      end
```
Insti: This is a nice test

lib/generator/files/track_files.rb
```diff
@@ -14,8 +14,12 @@ def filename(exercise_name)
         "#{exercise_name.tr('-', '_')}_cases"
       end
 
+      def class_name(exercise_name)
+        filename(exercise_name)[0..-2].split('_').map(&:capitalize).join
+      end
+
       def proc_name(exercise_name)
-        filename(exercise_name).split('_').map(&:capitalize).join
+        "#{class_name(exercise_name)}s"
```
Insti: If you're just adding an 's' do `class_name(exercise_name) + 's'`

Insti:
I'd like to see an example of how `class_name` and `proc_name` are used as part of this PR.

lib/generator/template_values.rb
```diff
@@ -31,8 +31,40 @@ def cases_require_name
       Files::GeneratorCases.filename(exercise_name)
     end
 
+    def test_cases
+      if test_cases_proc?
+        test_cases_proc.call(canonical_data.to_s)
+      else
+        test_cases_with_index
+      end
```
Insti: There's another object hiding in here.
Replace all this with:
```
test_cases_foobar.call(canonical_data.to_s)
```
Then work out what test_cases_foobar is.

lib/generator/template_values.rb
hilary: `test_cases_proc` returns `NameError` if the constant is not defined. I dislike rescuing an error in this situation, but would prefer to check to see if it is there.

lib/generator/template_values.rb
Insti: This will also help solve your "bunch of untested private methods" problem.

hilary:
`proc_name` is an existing method. I'm preserving it for backwards compatibility. Once we've converted we can ditch it.

lib/generator/template_values.rb
```diff
@@ -31,8 +31,40 @@ def cases_require_name
...
+      Object.const_defined?(test_cases_procname)
+    end
+
     def test_cases_proc
-      Object.const_get(Files::GeneratorCases.proc_name(exercise_name))
+      Object.const_get(test_cases_procname)
+    end
+
+    def test_case_class
+      Object.const_get(Files::GeneratorCases.class_name(exercise_name))
```
hilary: @Insti here is where `class_name` is used.

lib/generator/template_values.rb
hilary: 💯  Win, win, win. I'll refactor and dig that object out tomorrow.

lib/generator/template_values.rb
hilary: Me too!

lib/generator/template_values.rb
Insti: It was desirable before, for the whole thing to blow up if the test case class did not exist. Since that's no longer the case it suggests to me that `test_cases_proc` needs changing to something like:

```
def test_cases_proc
  Object.const_get(test_cases_procname) if Object.const_defined?(test_cases_procname)
end
```

(This may be moot due to upcoming object extraction.)

lib/generator/template_values.rb
Insti: so `proc_name` is never used externally and should be `private`?

hilary:
Refactoring to split apart the adapter and the two solutions

test/generator/template_values_test.rb
```diff
@@ -56,4 +56,115 @@ def teardown
...
+		"expected": 2
+	}
+	]
+}
+TEXT
+      )
+      simple_canonical_data
+    end
+
+    def complex_canonical_data
```
Insti: These (simple|complex) can live in `/fixtures` somewhere.

test/generator/template_values_test.rb
```diff
@@ -56,4 +56,115 @@ def teardown
       $LOAD_PATH.delete 'test/fixtures/xruby/lib'
     end
   end
+
+  class CaseValuesTest < Minitest::Test
+    def setup
+      $LOAD_PATH.unshift 'test/fixtures/xruby/lib'
```
Insti: Modifying the load path for every test should not be necessary.
But there might be something else to do so that everything still works.

test/generator/template_values_test.rb
```diff
@@ -56,4 +56,115 @@ def teardown
...
+    end
+
+    def simple_canonical_data
+      simple_canonical_data = Minitest::Mock.new
```
Insti: Why is this a `Mock`?

test/generator/template_values_test.rb
hilary: I was following the existing pattern.

Insti: I probably should fix that then :blush:

test/generator/template_values_test.rb
Insti: We discussed this in gitter and have decided to try a data section at the end of the test file. 
The main thing is removing the big heredoc from the middle of the test file.

test/generator/case_values_auto_extractor_test.rb
```diff
@@ -0,0 +1,116 @@
...
+TEXT
+        )
+        simple_canonical_data
+      end
+
+      def complex_canonical_data
+        complex_canonical_data = Minitest::Mock.new
+        complex_canonical_data.expect(
+          :to_s,
+          <<-TEXT
```
Insti: Were you going to move this data out of the tests?

test/generator/case_values_auto_extractor_test.rb
```diff
@@ -0,0 +1,116 @@
+require_relative '../test_helper'
```
Insti: `case_values_auto_extractor_test.rb` has too many `_auto_extractor`s in the filename

test/generator/case_values_auto_extractor_test.rb
hilary: They're at the bottom of the file now. 

Where I'd really like to put these is in a module I could include.

Insti: In which case why not use a fixture?

test/generator/template_values_test.rb
```diff
@@ -56,4 +56,5 @@ def teardown
       $LOAD_PATH.delete 'test/fixtures/xruby/lib'
     end
   end
+
```
Insti: An unnecessary blank line has snuck in.

test/generator/case_values_auto_extractor_test.rb
```diff
@@ -0,0 +1,118 @@
...
+
+module Generator
+  module CaseValues
+    class AutoExtractorTest < Minitest::Test
+
+      def setup
+        $LOAD_PATH.unshift 'test/fixtures/xruby/lib'
+      end
+
+      def test_simple_auto_extraction
```
Insti: I believe this test is redundant.

test/generator/case_values_auto_extractor_test.rb
hilary: A module can easily be included in multiple places, but (may be/is ?) easier to be surgical than a fixture.

I guess the question is, what are you suggesting? That I create fixtures for each of these cases, in which case I will end up duplicating data or reusing fixtures across tests, neither of which is desirable.

While data is non-executable, being able to manipulate the data via code makes the test suite far more manageable.

PR #570, opened 2017-04-27
Fix "exercise_cases" require path

Insti:
Updates all `lib/*_cases.rb` files to use the correct `exercise_cases.rb` file which is in `lib/generator` and removes the other one from `lib`, which is no longer used.

I have also added the require line to the files that were missing it.

Generation works even without the `require`, due to `generator/exercise_cases.rb` being required by the generator, but it is a good idea to make explicit the dependency within *_cases.rb file.

This also fixes the bug that `exercise` appeared in the list of available generators even though there was no `exercise` problem or test case generator for it.

### DIFF
--- a/lib/acronym_cases.rb
+++ b/lib/acronym_cases.rb
@@ -1,4 +1,4 @@
-require 'exercise_cases'
+require 'generator/exercise_cases'
 
 class AcronymCase < ExerciseCase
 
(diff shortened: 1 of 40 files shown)


hilary:
Makes a lot more sense than leaving in `require exercise_cases`, which I thought an odd request at the time :)

PR #575, opened 2017-04-28
Carry #549: Add generator for etl

hilary:
## Description
Carry @ajwann's valuable work forward so we can get it in :)

See original PR for description, motivation and context.

## How Has This Been Tested?
rake test

## Types of changes

- ~~New feature (non-breaking change which adds functionality)~~
- ~~Breaking change (fix or feature that would cause existing functionality to change)~~


lib/etl_cases.rb
```diff
@@ -0,0 +1,34 @@
...
+      "    expected = #{format_hash(expected)}\n",
+      indent(4, assertion),
+    ].join("\n")
+  end
+
+  private
+
+  def indent(size, text)
+    text.lines.each_with_object('') { |line, obj| obj << ' ' * size + line }
+  end
```
Insti: `ExerciseCases` needs some magic indenting code in it.

exercises/etl/etl_test.rb
```diff
@@ -3,56 +3,114 @@
...
-class TransformTest < Minitest::Test
-  def test_transform_one_value
-    old = { 1 => ['A'] }
-    expected = { 'a' => 1 }
+# Common test data version: ca9ed58
+class EtlTest < Minitest::Test
+  def test_a_single_letter
+    # skip
+    old = {
+       '1' => ["A"],
```
tommyschaefer: It looks like the generated test use strings rather than numbers (`'1'` vs `1`)

exercises/etl/etl_test.rb
```diff
@@ -3,56 +3,114 @@
...
+    }
+    expected = {
+       'a' => '1',
```
tommyschaefer: Same as above, except for the value rather than the key

lib/etl_cases.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def integerize_keys(input)
+    input.reduce({}) { |hash, (k, v)| hash[k.to_i] = v; hash }
+  end
+
+  def format_hash(hash)
+    middle = hash.each_pair.with_object('') do |(k, v), obj|
+      value = v.class == Array ? v : "'#{v}'"
+      obj << "      '#{k}' => #{value},\n "
```
tommyschaefer: When given a hash with integerized keys from https://github.com/exercism/xruby/pull/575/files#diff-7ccee8bf882363700677e1780f2665e4R6, the key is still wrapped in strings, negating `#integerize_keys`.

Insti: good spot @tommyschaefer :+1:

lib/etl_cases.rb
hilary: It has some magic indenting code, `indent_lines`. 

I've started converting `#workload` methods over to using it when I have a reason to rewrite a `#workload` method.

hilary:
Nice catch on the keys, @tommyschaefer . I'll get a chance to look at that after I get some sleep :)

PR #582, opened 2017-04-30
Probot: Changes "wontfix" to "stale" from default

kotp:
Change will have Probot label stale issues with "stale" rather than "wontfix"


Insti:
Does the "stale" label need to exist? It doesn't seem to right now.

kotp:
Yes, for this to work, the stale label would need to work.  I had not done that yet, since this in the process of being reviewed.

PR #584, opened 2017-04-30
Revert "Simplify test for MetadataFiles#canonical_data"

Insti:
This reverts commit 4f572e8e4bd3913f530439fdf4d3bfcc144d9aa6.

This change worked on Ruby 2.2 but stopped working from ruby 2.3 due to
changes in the way OpenStruct works.

See also:

http://stackoverflow.com/questions/39278864/openstruct-issue-with-ruby-2-3-1
minimal example: https://gist.github.com/anonymous/162dd5466b23331e0e31b891ac61d337

### DIFF
--- a/test/generator/files/metadata_files_test.rb
+++ b/test/generator/files/metadata_files_test.rb
@@ -7,9 +7,17 @@ module Generator
         track: 'test/fixtures/xruby'
       )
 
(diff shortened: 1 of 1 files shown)


hilary:
Yep, stable tests are a good thing!

Issue #576, opened 2017-04-28
Difference of squares: template has a single hardcoded test


hilary:
## Expected Behavior
The data for the test should be in canonical data. Hard-coding it into the view is a smell.

## Possible Solution
I would vote for deleting the test.

Alternatively, several other exercises have a need for a custom section inserted after the generated tests that isn't simply not putting a test into canonical data. Once we decide how to handle those sections, we could leverage that technique to clean up this smell. But I would rather not encourage avoiding canonical data.

## Context
Other than this single test, difference of squares uses the default template. I wouldn't want to encourage customizing the template just to add additional tests without adding the data to canonical data.

Insti:
Link to file being discussed: https://github.com/exercism/xruby/blob/master/exercises/difference-of-squares/example.tt

Insti:
Yes delete the test.

How would you even write code that caused it to fail?

Tests all\* should be in the canonical-data.


\* exceptions apply for tricky and Ruby specific tests.

Insti:
I was wondering why this test was added: 

Here is the commit that added the test: https://github.com/exercism/xruby/commit/22127fad27b9ae8e4466f1e68599d80ae81836b0

> Since the API calls for a `Squares` object to be instantiated with a
specific max value, it seems to be implied that the instance methods of
a given instance should have the same result on subsequent calls.
>
> While not common, I have seen more than one submission that didn't have
this property.

It is implementation specific so doesn't belong in canonical-data, maybe we do need it?

Insti:
> maybe we do need it?

I'd argue we'd be better off changing the implementation to something more appropriate.
There's no reason why (and plenty of reasons why not) `Squares` should be an instantiable object.

kotp:
Definitely a delete the test vote from me.

hilary:
Sounds like we have decided to delete the test (yay!)

Insti:
I'm changing my vote to keep the test.

 (but I think I'm outvoted.)

Remove it when the implementation change makes it redundant.

hilary:
Ok, so what we want is to do two things:

1. change the implementation in `#workload` so it doesn't specify an instantiable object
1. remove the hardcoded test

tommyschaefer:
> Remove it when the implementation change makes it redundant.

I'm okay with this too. It's a toss-up for me 😄

kotp:
[...]

PR #585, opened 2017-04-30
generator: move existing custom test templates

hilary:
Going through the existing custom test templates and:

* converting the exercise to use the default test template when possible (half, lol)
* moving the custom test template to `exercises/<exercise_name>/.meta/generator/test_template.erb` otherwise (other half)

For more context, see the generator improvements issue: exercism/xruby#485

Closes exercism/xruby#576


exercises/tournament/tournament_test.rb
```diff
@@ -5,65 +5,64 @@
...
     # skip
     input = <<-INPUT.gsub(/^ */, '')
-      Allegoric Alaskans;Blithering Badgers;win
-      Devastating Donkeys;Courageous Californians;draw
-      Devastating Donkeys;Allegoric Alaskans;win
-      Courageous Californians;Blithering Badgers;loss
-      Blithering Badgers;Devastating Donkeys;loss
-      Allegoric Alaskans;Courageous Californians;win
-    INPUT
+Allegoric Alaskans;Blithering Badgers;win
```
tommyschaefer: This doesn't apply in our case since we run CI on ruby `2.1.2` and only require ruby `2.1+`, but I just learned about a cool `2.3+` ruby feature this week!

If you declare your heredoc with `<<~` (`~` rather than `-`), it does this:

```
The indentation of the least-indented line will be removed from each line of the content.
```

Sorry if you'd already seen this, I just got really excited because heredoc indentation has always tripped me up!

exercises/all-your-base/all_your_base_test.rb
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env ruby
+gem 'minitest', '>= 5.0.0'
```
tommyschaefer: It looks like this may not be needed, since it ran without it in the previous version of the generated test.

Insti: It isn't needed: https://github.com/exercism/xruby/issues/587

exercises/tournament/tournament_test.rb
kotp: I like that feature as well.

Insti: Yeah, looking forward to '<<~'

But these lines should still be indented.

exercises/transpose/transpose_test.rb
```diff
@@ -9,24 +9,24 @@ def test_empty_string
     # skip
     input = <<-INPUT.gsub(/^ {6}/, '')
 
-    INPUT
+INPUT
     actual = Transpose.transpose(input)
     expected = <<-EXPECTED.gsub(/^ {6}/, '')
 
-    EXPECTED
+EXPECTED
```
Insti: More heredoc indentation issues.

lib/generator/exercise_cases.rb
```diff
@@ -25,6 +25,23 @@ def indent_lines(code, depth, separator = "\n")
...
   end
 
+  # use to indent multi line workloads with blank
+  # lines
+  #   indent_text(4, lines.join("\n"))
+  def indent_text(depth, text)
+    text.lines.reduce do |obj, line|
+      obj << (line == "\n" ? line : ' ' * depth + line)
+    end
+  end
```
Insti: I'd like to see the `exercise_cases` changes in a separate PR and have this one be just about moving the template files.

exercises/tournament/tournament_test.rb
hilary: Why should these lines be indented? 

Yes, they were indented in the old version, but ...

We don't indent the heredoc for beer-song, why are we indenting these lines?

Looking at the README, there's no inherent reason for the lines in the heredoc to be indented. The only reason it is indented is to `indent` the test. We wouldn't expect a real problem to change the way its data was configured just so the test appeared correctly indented!

exercises/transpose/transpose_test.rb
hilary: same response as above

lib/generator/exercise_cases.rb
hilary: I wrote the methods in exercise_cases as I worked my way through moving the template files. 

I could artificially go back and separate out the two, but it feels, well, artificial?

exercises/tournament/tournament_test.rb
Insti: > Why should these lines be indented?

It's probably just consistent indent fetish.
Why should any lines be indented?

I guess it's ok if you feel strongly they shouldn't be indented.
In which case remove from the line above the `gsub` that removes the whitespace.

> We wouldn't expect a real problem to change the way its data was configured just so the test appeared correctly indented!

There's no change to the indentation used as an argument to the problem, just its presentation within the tests.

lib/generator/exercise_cases.rb
Insti: This one I do feel strongly about. You wouldn't need to change this PR at all, just extract the `exercise_cases` changes and we apply them first.
I can do it if you don't want to.

hilary: I'll separate them out.

hilary:
ok, this now has only the moves and changes.

PR #586, opened 2017-04-30
generators: move existing generators to .meta/generator/

hilary:
Now that exercism/xruby#581 is merged, move all the existing generators under their exercise directory.

Another generator improvement brought to you by issue exercism/xruby#485


Insti:
Yay.

PR #588, opened 2017-05-01
run-length-encoding: Namespace tests for encode/decode

iamchrissmith:
## Why?
1) This provides added clarity for which tests are running/failing/passing.
2) This fixes a name collision bug between:
`def test_empty_string` for `encode` on line `10`
and 
`def test_empty_string` for `decode` on line `50`

## What?
I added `encode_` or `decode_` as a prefix for the test names that did not already specify which method they were testing.

## How Has This Been Tested?
I used this version to complete the exercism locally

## Types of changes
- [ ] Enhancement (non-breaking change which adds functionality)
- [x] Gardening (code and/or documentation cleanup)

## Checklist:

- [x] My change relies on a pending issue/pull request
- [ ] I have added tests to cover my changes.

### DIFF
--- a/exercises/run-length-encoding/run_length_encoding_test.rb
+++ b/exercises/run-length-encoding/run_length_encoding_test.rb
@@ -5,7 +5,7 @@ require_relative 'run_length_encoding'
 
 # Common test data version: 503a57a
 class RunLengthEncodingTest < Minitest::Test
-  def test_empty_string
+  def test_encode_empty_string
     # skip
     input = ''
     output = ''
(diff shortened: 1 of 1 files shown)


Insti:
Thanks for finding this and submitting a PR @iamchrissmith.

However, this is one of our generated test suites. (See the README: https://github.com/exercism/xruby#generated-test-suites) so these changes will be overwritten next time the test case generation script is run!

If you're willing to try and update the test generator for this problem, the place to make these changes would be in the run_length_encoding_cases.rb file: https://github.com/exercism/xruby/blob/master/exercises/run-length-encoding/.meta/generator/run_length_encoding_cases.rb

You will probably need to provide a custom `name` method that overrides the one provided in the base class: https://github.com/exercism/xruby/blob/master/lib/generator/exercise_cases.rb

This is all a bit in flux at the moment and this process is not as well documented as we would like as we are in the process of overhauling the generator code, so let us know if there is anything confusing or any improvements that would make your life easier.

We can usually be found in https://gitter.im/exercism/xruby if you have any questions.

Insti:
The data that the tests are generated comes from: https://github.com/exercism/x-common/blob/master/exercises/run-length-encoding/canonical-data.json. maybe the `property` property would be helpful. It should already be accessible as a method on 
`RunLengthEncodingCase`

kotp:
Good catch on the duplicate method name, @iamchrissmith !

Insti:
Thanks for bringing this to our attention @iamchrissmith , I've fixed this by updating the test generator in #593. The updated tests should make their way to the live site in a day or two.

PR #591, opened 2017-05-02
generators: Exercise cases helper methods

hilary:
Separating out the changes to ExerciseCase from moving the test templates.... 

This PR goes before exercism/xruby#585

I also:
* moved `ExerciseCase` under the `Generator` module (and gave thanks for `sed`)
* moved the assert helper methods out into their own module, for ease of testing and because `ExerciseCase` was becoming bloated.
* added test coverage for `ExerciseCase`


Insti:
Travis is currently broken, tests pass when run locally.

PR #592, opened 2017-05-03
run-length-encoding: Remove redundant test

jwworth:
Because there are two tests with the same name (see line 9), Minitest incorrectly infers that the test is skipped, not that it is passing or failing, when the user first runs the test suite.

### DIFF
--- a/exercises/run-length-encoding/run_length_encoding_test.rb
+++ b/exercises/run-length-encoding/run_length_encoding_test.rb
@@ -47,13 +47,6 @@ class RunLengthEncodingTest < Minitest::Test
     assert_equal output, RunLengthEncoding.encode(input)
   end
 
-  def test_empty_string
-    skip
-    input = ''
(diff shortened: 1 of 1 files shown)


Insti:
Thanks @jwworth , this seems to be a duplicate of: https://github.com/exercism/xruby/pull/588

Insti:
While we appreciate your contribution @jwworth :heart:, I'm going to close this pull request.

The test is not actually a duplicate, since one is a test for the `encode` method and the other is a test for the `decode` method.

The solution proposed by #588 will also fix this problem.

These tests come from a generated test file, and changes to the generated file will be overwritten when the generator is re-run. (as described in #558)

Please continue to submit issues/PRs for problems you find as they help make Exercism better.

Thanks.

jwworth:
Makes sense! Thanks for the explanation and for maintaining this great project.

PR #593, opened 2017-05-03
run-length-encoding: Add property to generated tests

Insti:
Resolves #588 

* Update the test generator to add `encode_` or `decode_` as a prefix to all tests depending on the method being tested.
This fixes a name collision bug between:
def `test_empty_string` for encode
and
def `test_empty_string` for decode

* Regenerated tests.


Insti:
Test version number is not incremented, due to there being no changes in the content of the  tests.

PR #594, opened 2017-05-03
generator: Rename `exercise_cases`

Insti:
The class implemented in these files is called `ExerciseCase`.
* Update the filename to match.
* Update reference in tests
* Update references in exercises
* Update README.


hilary:
happiness! 💯

PR #595, opened 2017-05-03
Readme: Update default generator example.

Insti:
Add some missing elements:
require statement
`Generator` namespace


Insti:
I was trying to make one, so thought I should read the instructions about how to do it.

hilary:
What did you think about the instructions overall?

PR #605, opened 2017-05-03
generator: support example solutions in .meta/solutions

hilary:
As part of moving generator bits under `.meta`, we're moving `example.rb` under `.meta/solutions`. This PR is the first step: adding support for example solutions as `.meta/solutions/<exercise_name>.rb` while retaining support for the current location.

Tests are included for both old and new locations.

Supporting both old and new introduces a bit of mess that will get cleaned up when support for the old location is removed. I'd rather live with the mess than refactor to get an elegant solution that would be around for a day or two...


test/generator/files/track_files_test.rb
```diff
@@ -13,7 +13,7 @@ def initialize
           @paths = FixturePaths
           @exercise_name = 'alpha'
         end
-        attr_reader :paths, :exercise_name
+        attr_accessor :paths, :exercise_name
```
Insti: Why has this changed?

Insti: I'd rather exercise name modification being done via constructor argument rather than `attr_accessor`

PR #614, opened 2017-05-08
generator: disable old example loc

hilary:
Last step (woohoo!) in moving everything needed to generate a test suite (other than canonical data) under `.meta`. For more info, see Generator Improvements, exercism/xruby#485

Changes:
* Removed support for legacy example solution location
* Cleaned up a fixture file we missed at an earlier stage
* A little bit of cleanup in TrackFiles so it doesn't sprawl quite so much.

In addition to running the test suite, I tested the changes by regenerated a sample of the test suites and checking that nothing changed.


lib/generator/files/track_files.rb
```diff
@@ -47,19 +49,6 @@ def default_tests_template_filename
       def tests_template_filename
         'test_template.erb'
       end
-
-      def example_file
-        File.exist?(File.join(exercise_path, example_filename)) ? example_filename : legacy_example_filename
-      end
-
-      def example_filename
```
Insti: I'd like these methods back please.

test/generator/files/track_files_test.rb
```diff
@@ -24,15 +24,7 @@ def test_tests_version
...
-        expected_filename = FixturePaths.track + '/exercises/alpha/.meta/solutions/alpha.rb'
-        assert_equal expected_filename, subject.example_solution.filename
-      end
-
-      def test_legacy_example_solution
-        subject = TestTrackFiles.new
-        subject.exercise_name = 'beta'
-        expected_filename = FixturePaths.track + '/exercises/beta/example.rb'
-        assert_equal expected_filename, subject.example_solution.filename
+        assert_instance_of ExampleSolutionFile, subject.example_solution
```
Insti: This no longer tests the filename is correct.

lib/generator/files/track_files.rb
```diff
@@ -49,9 +51,6 @@ def default_tests_template_filename
       def tests_template_filename
         'test_template.erb'
       end
-      def minitest_tests_filename
```
Insti: I'd like this method back please.

test/generator/files/track_files_test.rb
hilary: Right. I reverted it back to the pattern that these tests used to follow now that we no longer have multiple filename patterns.

lib/generator/files/track_files.rb
hilary: done

PR #615, opened 2017-05-08
generator: remove requirement for test case to know its index

hilary:
Part of the Generator Improvements: exercism/xruby#485

Changes:
* Convert the default `test_template.erb` to include the index in the iterator (with test changes, etc)
* Update the six custom test templates in the same way
* Stop injecting index as a case value
* (and cleaned up a fixture/test that was leftover from the cases proc logic)

Testing:
* `rake test`
* regenerated test suites for all six exercises with custom test templates, no changes
* regenerated test suites for a sample of other exercises, no changes


Insti:
One shot sounds fine if that will be easiest. We have much better knowledge of all the generators right now.

Insti:
Looks good, works. Merge when ready. :+1:

PR #617, opened 2017-05-09
robot-name: Updated based on clarified description.

Insti:
The robot name description [has been updated](https://github.com/exercism/x-common/pull/765)  to clarify that name uniqueness is a requirement.

> The names must be random: they should not follow a predictable sequence.
Random names means a risk of collisions. Your solution must ensure that
every existing robot has a unique name.

This PR updates the tests and example solution to match this.

Because the new problem is significantly harder than the old version it has been moved down in the order of problems.
 
A HINTS.md file has been added to specify the `forget` method used to aid in testing.

### DIFF
--- a/config.json
+++ b/config.json
@@ -61,12 +61,6 @@
       ]
     },
     {
-      "slug": "robot-name",
-      "difficulty": 1,
-      "topics": [
-      ]
-    },
-    {
       "slug": "nth-prime",
       "difficulty": 1,
       "topics": [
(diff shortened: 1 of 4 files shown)


hilary:
I'm concerned that the new version requires a specific approach. It's a good approach, but nonetheless a specific approach. I don't see `Robot.forget` as a product requirement in the same way that `robot.name` and `robot.reset` are.

kotp:
@remcopeereboom may be a good resource for this exercise as well.

Insti:
> I'm concerned that the new version requires a specific approach.

Can you suggest a way to test this that does not require it?

hilary:
I've been thinking about it, and everything I come up with ends up being *more* prescriptive, not less.

This update is a big step in the right direction, so I'm in favor.

👍

PR #618, opened 2017-05-09
generator: update location of example solution in README

hilary:
We missed updating the README in the process of sorting out the needed changes to configlet, etc.


hilary:
@Insti, @kotp  I'm going to merge this now as it's only changing the README and the README is currently wrong (something on the internet is wrong!). Feel free to ask for changes even though I've merged.

PR #619, opened 2017-05-09
generator: separate option parsing from generator class construction

hilary:
I started work on adding the `--all` option and quickly realized that `command_line.rb` needed some love. So I decided to take a step back and do the refactor in the current version of `command_line.rb` (with the current tests, etc) in a separate PR.

I have not yet made changes to the existing tests. I'd like to get some feedback before proceeding.

### DIFF
--- a/lib/generator/command_line.rb
+++ b/lib/generator/command_line.rb
@@ -10,8 +10,8 @@ module Generator
 
     def parse(args)
       parser = GeneratorOptparser.new(args, @paths)
-      @options = parser.parse_options
-      generator if parser.valid?
+      @options = parser.options
+      generator if parser.options_valid?
     end
 
     private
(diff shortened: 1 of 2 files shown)


lib/generator/command_line.rb
```diff
@@ -1,26 +1,21 @@
...
       @paths = paths
-      @options = default_options
     end
 
     def parse(args)
-      parse_options(args)
-      generator if valid?
+      parser = GeneratorOptparser.new(args, @paths)
+      @options = parser.parse_options
+      generator if parser.valid?
```
Insti: very minor: We want to know if the options are valid, not if the parser is valid, so this reads oddly to me.

hilary: I'll rename :)

hilary: In fact, the flow needs to be a bit different...

kotp:
Very nice.

PR #620, opened 2017-05-10
generator: generate --all

hilary:
Now we're getting to the fun stuff in the Generator Improvements exercism/xruby#485

Add an `--all` option to `bin/generate`.

Notes:
* I decided that `--all` implies `--freeze`, as updating the xruby version of a generator is something that a contributor working on a specific generator does. (We probably have that default backwards...)
* In a future enhancement, we might consider checking whether the local copy of x-common is up to date and issuing a warning when regenerating all if it isn't.
* Code coverage is 100% on everything except for `lib/generator.rb`. I'm wrestling with the mocking and stubbing involved with testing `GenerateAllTests#call` via minitest.


lib/generator/files/generator_cases.rb
```diff
@@ -4,7 +4,7 @@ module GeneratorCases
       module_function
 
       def available(track_path)
-        generator_glob = File.join(track_path, "{lib,exercises}", '**', '*_cases.rb')
+        generator_glob = File.join(track_path, "exercises", '**', '*_cases.rb')
```
Insti: This probably maybe could possibly refer to the full correct path and only glob the `*_cases.rb`

lib/generator/command_line.rb
```diff
@@ -9,33 +7,23 @@ def initialize(paths)
     end
 
     def parse(args)
-      parser = GeneratorOptparser.new(args, @paths)
+      parser = GeneratorOptparser.new(args, paths)
       @options = parser.options
-      generator if parser.options_valid?
+      generate if parser.options_valid?
```
Insti: generate is a verb, generator is a noun.

hilary: So it is 😄  

The change is an artifact of me tugging at the code repeatedly until I arrived at what you see. Looking at it now, I don't like either. `generator if parser.options_valid?` is lousy english. `generate if parser.options_valid?` is worse, as it is misleading. Maybe `#build_generator`?

lib/generator/files/generator_cases.rb
hilary: pass in exercise_path rather than track_path? Not sure what you have in mind...

lib/generator/command_line.rb
Insti: Maybe adding an explicit return.
`return generator if parser.options_valid?`

lib/generator/files/generator_cases.rb
Insti: Oh right, I forgot about that variable bit in the middle, what about:

`File.join(track_path, "exercises", '*','.meta','generator', '*_cases.rb')`

lib/generator.rb
```diff
@@ -11,6 +11,26 @@ def initialize(track:, metadata:)
     end
   end
 
+  # A generator has to have a #call method, that's the only requirement.
+  # It doesn't have to be a RepositoryDelegator
```
Insti: A generator needs to be `RepositoryDelegator`, but `GenerateAllTests` is a meta generator, so _it_ doesn't need one. I suspect the term "generator" is too overloaded.

lib/generator/generator_factory.rb
```diff
@@ -0,0 +1,38 @@
+require 'logger'
+
+module Generator
+  class GeneratorFactory
+    attr_reader :paths, :exercise_name, :verbose, :freeze
+
+    def initialize(paths:, exercise_name:, verbose: false, freeze: false)
```
Insti: paths + exercise_name is a Repository
verbose + freeze are options

PR #622, opened 2017-05-10
[WIP] Refactor exercise case

hilary:
A first pass at "ExerciseCase should not be an OpenStruct and the canonical data should be kept separate from the methods of this class." from Generator Improvements exercism/xruby#485.

* ExerciseCase is no longer an OpenStruct
* canonical data is injected into an ExerciseCase as case_data

More changes need to be made, in the various *_cases.rb. Identifying what changes need to be made will be simple once exercism/xruby#620 is merged 😄 

Todo:
* rebase onto master
* run `bin/generate --all` to figure out what else needs to be updated and update
* for bonus points, use an OpenStruct rather than a hash for the data
* for bonus points, write tests to catch what got caught by running `bin/generate --all` in this case

### DIFF
--- a/exercises/wordy/.meta/generator/wordy_cases.rb
+++ b/exercises/wordy/.meta/generator/wordy_cases.rb
@@ -4,7 +4,7 @@ class WordyCase < Generator::ExerciseCase
 
   def workload
     [
-      "question = '#{input}'",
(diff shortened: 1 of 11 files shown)


Insti:
Superseded by #648

Issue #627, opened 2017-05-12
Generator: split `GeneratorCases` into `Cases` and `Case`


hilary:
We've got instance and class methods mixed into a module full of class methods, which is driving us batty. Discovered during exercism/xruby#625

PR #632, opened 2017-05-12
generator: rename test generators to *_case

hilary:
This is a quick PR to cleanup a leftover discrepancy. The test generators were still named `<exercise_name>_cases.rb` even though they now contain a single class each name `<ExerciseName>Case`.

I did this PR now as it makes the larger refactor I'm working on, for exercism/xruby#627, cleaner.


sed
```diff
@@ -0,0 +1 @@
+foo_cases s/cases/case/g
```
Insti: This file should not be here.

Insti:
I'd rather have PR dependency chains than PRs merged without review.

hilary:
It's something I expect to do very rarely if at all... especially as even what seems the simplest PR tends to have mistakes like the non-file file that slipped in.

Asking you to review it after the fact was my backstop 😄  I figured you would catch anything that slipped through.

PR #633, opened 2017-05-13
generator: refactor generate cases

### LINKED ISSUE #627: Generator: split `GeneratorCases` into `Cases` and `Case`

hilary:
There's a `Case` object (with a `test_class` and a `filepath`) lurking in here, but I'm not sure it's worth extracting into its own class until we tackle the larger composition issues.

This PR is intended to prep us for composition rather than including, or at least not get in the way.

### DIFF
--- a/lib/generator/files/generator_cases.rb
+++ b/lib/generator/files/generator_cases.rb
@@ -10,16 +10,10 @@ module Generator
           filename(exercise_name_or_slug).split('_').map(&:capitalize).join
         end
 
-        def source_filepath(track_path, slug)
-          path = meta_generator_path(track_path, slug)
-          filename = filename(slug) + '.rb'
-          File.join(path, filename)
-        end
-
         private
 
         def cases_filepaths(track_path)
-          generator_glob = File.join(meta_generator_path(track_path, '*'), '*_case.rb')
+          generator_glob = File.join(track_path, 'exercises', '*', '.meta', 'generator', '*_case.rb')
           Dir.glob(generator_glob, File::FNM_DOTMATCH)
         end
 
@@ -29,10 +23,6 @@ module Generator
 
         def filename(exercise_name_or_slug)
           "#{exercise_name_or_slug.tr('-', '_')}_case"
-        end
-
-        def meta_generator_path(track_path, slug)
-          File.join(track_path, 'exercises', slug, '.meta', 'generator')
         end
       end
     end
--- a/lib/generator/files/generator_cases.rb
+++ b/lib/generator/files/generator_cases.rb
@@ -6,10 +6,6 @@ module Generator
           cases_filepaths(track_path).map { |filepath| slugify(filepath) }.sort
         end
 
-        def class_name(exercise_name_or_slug)
-          filename(exercise_name_or_slug).split('_').map(&:capitalize).join
-        end
-
         private
 
         def cases_filepaths(track_path)
@@ -18,11 +14,7 @@ module Generator
         end
 
         def slugify(filepath)
-          %r{([^/]*)_case\.rb$}.match(filepath).captures[0].tr('_', '-')
-        end
-
-        def filename(exercise_name_or_slug)
-          "#{exercise_name_or_slug.tr('-', '_')}_case"
+          File.basename(filepath, '_case.rb').tr('_', '-')
         end
       end
     end
--- a/lib/generator/files/track_files.rb
+++ b/lib/generator/files/track_files.rb
@@ -56,7 +56,7 @@ module Generator
       end
 
       def cases_filename
-        "#{exercise_name}_case.rb"
+        "#{test_case_name}.rb"
       end
 
       def tests_template_absolute_filename
--- a/lib/generator/files/track_files.rb
+++ b/lib/generator/files/track_files.rb
@@ -17,8 +17,8 @@ module Generator
         ExampleSolutionFile.new(filename: File.join(solutions_path, example_filename))
       end
 
-      def cases_filepath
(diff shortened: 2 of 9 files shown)


lib/generator/files/generator_cases.rb
```diff
@@ -10,16 +10,10 @@ def class_name(exercise_name_or_slug)
...
-          path = meta_generator_path(track_path, slug)
-          filename = filename(slug) + '.rb'
-          File.join(path, filename)
-        end
-
         private
 
         def cases_filepaths(track_path)
-          generator_glob = File.join(meta_generator_path(track_path, '*'), '*_case.rb')
+          generator_glob = File.join(track_path, 'exercises', '*', '.meta', 'generator', '*_case.rb')
```
Insti: I really don't like inlining this method.

lib/generator/files/track_files.rb
```diff
@@ -17,6 +17,10 @@ def example_solution
         ExampleSolutionFile.new(filename: File.join(solutions_path, example_filename))
       end
 
+      def cases_filepath
```
Insti: this is not a cases_filepath it's the file, so should be `cases`

lib/generator/files/generator_cases.rb
```diff
@@ -6,10 +6,6 @@ def available(track_path)
           cases_filepaths(track_path).map { |filepath| slugify(filepath) }.sort
         end
 
-        def class_name(exercise_name_or_slug)
-          filename(exercise_name_or_slug).split('_').map(&:capitalize).join
-        end
```
Insti: The number of files this change effects concerns me.

lib/generator/repository.rb
```diff
@@ -35,6 +35,10 @@ def create_tests_file
     def exercise_name
       @exercise_name ||= slug.tr('-', '_')
     end
+
+    def test_case_name
+      exercise_name + '_case'
```
Insti: This is the wrong place for this method.
If it matters what the filename is it should somewhere under Generator::Files

lib/generator/template_values.rb
```diff
@@ -44,10 +44,14 @@ def extract
 
     def extractor
         CaseValues::Extractor.new(
-          case_class: Object.const_get(Files::GeneratorCases.class_name(slug))
+          case_class: Object.const_get(case_class_name)
```
Insti: should be `case.class_name`

lib/generator/template_values.rb
```diff
@@ -44,10 +44,14 @@ def extract
...
         )
     end
 
+    def case_class_name
+      test_case_name.split('_').map(&:capitalize).join
```
Insti: Does not belong here, should be in `Case`

sed
```diff
@@ -1 +0,0 @@
-foo_cases s/cases/case/g
```
Insti: indeed.

lib/generator/files/generator_cases.rb
```diff
@@ -6,33 +6,15 @@ def available(track_path)
...
-          %r{([^/]*)_case\.rb$}.match(filepath).captures[0].tr('_', '-')
-        end
-
-        def filename(exercise_name_or_slug)
-          "#{exercise_name_or_slug.tr('-', '_')}_case"
-        end
-
-        def meta_generator_path(track_path, slug)
-          File.join(track_path, 'exercises', slug, '.meta', 'generator')
+          File.basename(filepath, '_case.rb').tr('_', '-')
```
Insti: Nice use of basename 👍

lib/generator/template_values.rb
Insti: case being a keyword is a bit annoying.

lib/generator/files/track_files.rb
```diff
@@ -91,5 +103,7 @@ def generate(template:, values:)
 
     class TestsTemplateFile < Readable
     end
+
+    class CaseFile < Readable; end
```
Insti: There's no reason to only use one line for case definitions.

Insti:
My comments here are "things I thought while reading through this" rather than specific requests for changes.

lib/generator/template_values.rb
```diff
@@ -44,12 +44,16 @@ def extract
```
Insti: (github: why can't I comment on unchanged lines.)

`load cases_load_name` can become `case.filename`

lib/generator/files/generator_cases.rb
hilary: Inlining this falls under the 'sometimes it gets uglier on the way to getting cleaner'

lib/generator/repository.rb
hilary: Isn't that true for `#exercise_name` as well? What's the difference between the two?

hilary:
After discussion with @Insti (all of this will link back to this PR)
* extracting a few of the changes here into standalone PRs 
* doing a new refactor (`Repository` => `Exercise`)
* extracting Case into its own class
* revisiting GeneratorCases

lib/generator/repository.rb
Insti: Yes it is.

Differences: 
* `exercise_name` is a property of the slug so it's less obvious where to put it.
* I didn't notice it was in the wrong place when I reviewed the PR that added it.

Similarities:
* They are both something that relate to Files/Filenames so should be under Generator::Files

hilary:
closing this as we've gone in a different direction

PR #637, opened 2017-05-13
generator: rename Repository model

hilary:
Step 1 in separating out the Repository and Exercise models

(see also exercism/xruby#633)

### DIFF
--- a/lib/generator.rb
+++ b/lib/generator.rb
@@ -13,14 +13,14 @@ module Generator
 
   # This contains the order for updating/generating the files. (Strategy pattern).
   # Doesn't update the version information.
-  class GenerateTests < RepositoryDelegator
+  class GenerateTests < ExerciseDelegator
     def call
       create_tests_file
     end
   end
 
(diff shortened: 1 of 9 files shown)


Insti:
In: https://github.com/exercism/xruby/blob/master/lib/generator/files.rb#L52
There is a pre-existing module called `Exercise`.

Which is used by: 
lib/generator/files/track_files.rb:      include Exercise
lib/generator/files/metadata_files.rb:      include Exercise

To cause track_files and metadata files to emit `NotImplementedError`s if these methods aren't implemented.

I Suggest:

Delete this module completely and copy the unimplemented methods directly into TrackFiles and MetadataFiles classes rather than including.
Code duplication is a minor consideration here.

hilary:
I remembered about the pre-existing `Exercise` module, too, but that can be done in a separate PR. There's no inherent need to do it in the same one....

Insti:
Removing it completely also works 👍

PR #638, opened 2017-05-14
generator: extract repository model

hilary:
second step in separating out the Repository and Exercise models. The first step is in exercism/xruby#637 (this PR builds on that one). All of this came out of exercism/xruby#633.

My goal is this PR was to create the new version of Repository and inject Repository into Exercise without engaging in collateral refactoring(s). Aka, one step at a time.

So, for example:
* the case object still needs to get extracted
* Exercise uses delegation more than it probably should
* I just moved some of the old includes over to Repository, and left another in Exercise.

Why have all the fun myself ;)

### DIFF
--- a/lib/generator/command_line.rb
+++ b/lib/generator/command_line.rb
@@ -16,15 +16,15 @@ module Generator
     attr_reader :paths
 
     def generators
-      exercises.map { |slug| generator(repository(slug)) }
+      exercises.map { |slug| generator(exercise(slug)) }
     end
 
     def exercises
       @options[:all] ? Files::GeneratorCases.available(paths.track) : [@options[:slug]]
     end
 
-    def generator(repository)
-      generator_class.new(repository)
+    def generator(exercise)
+      generator_class.new(exercise)
     end
 
     def generator_class
@@ -35,9 +35,9 @@ module Generator
       @options[:freeze] || @options[:all]
     end
 
-    def repository(slug)
-      LoggingRepository.new(
-        repository: Repository.new(paths: paths, slug: slug),
+    def exercise(slug)
+      LoggingExercise.new(
+        exercise: Exercise.new(paths: paths, slug: slug),
         logger: logger
       )
     end
--- a/lib/generator/command_line.rb
+++ b/lib/generator/command_line.rb
@@ -16,7 +16,7 @@ module Generator
     attr_reader :paths
 
     def generators
-      exercises.map { |slug| generator(exercise(slug)) }
+      exercises.map { |slug| generator(exercise(repository(slug))) }
     end
 
     def exercises
@@ -35,11 +35,15 @@ module Generator
       @options[:freeze] || @options[:all]
     end
 
-    def exercise(slug)
+    def exercise(repository)
       LoggingExercise.new(
-        exercise: Exercise.new(paths: paths, slug: slug),
+        exercise: Exercise.new(repository: repository),
         logger: logger
       )
+    end
+
+    def repository(slug)
+      Repository.new(paths: paths, slug: slug)
     end
 
     def logger
--- a/lib/generator/files/track_files.rb
+++ b/lib/generator/files/track_files.rb
@@ -36,7 +36,7 @@ module Generator
       end
 
       def minitest_tests_filename
-        "#{exercise_name}_test.rb"
+        "#{name}_test.rb"
       end
 
       def version_filename
@@ -44,7 +44,7 @@ module Generator
       end
 
       def example_filename
-        "#{exercise_name}.rb"
+        "#{name}.rb"
       end
 
       def tests_template_absolute_filename
--- a/lib/generator/repository.rb
+++ b/lib/generator/repository.rb
@@ -13,7 +13,12 @@ module Generator
       @name ||= slug.tr('-', '_')
     end
 
-    def cases_load_name
+    # these two should get refactored somewhere together later
+    def case_class_name
+      Files::GeneratorCases.class_name(slug)
+    end
+
+    def case_load_name
       Files::GeneratorCases.source_filepath(paths.track, slug)
     end
   end
--- a/lib/generator/repository.rb
+++ b/lib/generator/repository.rb
@@ -12,5 +12,9 @@ module Generator
     def name
       @name ||= slug.tr('-', '_')
     end
+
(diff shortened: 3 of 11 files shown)


lib/generator/files/track_files.rb
```diff
@@ -36,15 +36,15 @@ def solutions_path
       end
 
       def minitest_tests_filename
-        "#{exercise_name}_test.rb"
+        "#{name}_test.rb"
```
Insti: This is not right.
Track files should not know anything about Exercise.

lib/generator/files/track_files.rb
```diff
@@ -36,15 +36,15 @@ def solutions_path
...
       end
 
       def version_filename
         '.version'
       end
 
       def example_filename
-        "#{exercise_name}.rb"
+        "#{name}.rb"
```
Insti: See above.

test/generator/exercise_test.rb
```diff
@@ -95,9 +95,9 @@ def test_bookkeeping
       Object.send(:remove_const, :AlphaCase)
     end
 
-    def test_exercise_name
+    def test_name
       subject = Exercise.new(paths: FixturePaths, slug: 'alpha-beta')
-      assert_equal 'alpha_beta', subject.exercise_name
+      assert_equal 'alpha_beta', subject.name
```
Insti: I REALLY don't like this being a part of `Exercise` now.
(but yeah, refactor later.)

test/generator/files/track_files_test.rb
```diff
@@ -12,9 +12,9 @@ class TestTrackFiles
         def initialize
           @paths = FixturePaths
           @slug = 'alpha-beta'
-          @exercise_name = 'alpha_beta'
+          @name = 'alpha_beta'
```
Insti: should still be exercise_name and should not exist.

test/generator/exercise_test.rb
```diff
@@ -83,21 +88,27 @@ def test_bookkeeping
...
+       mock_minitest_file = Minitest::Mock.new.expect(:generate, expected_content,
+         [template: mock_template_file, values: mock_template_values])
+
+       mock_repository.expect :minitest_tests, mock_minitest_file
+
+       assert_equal expected_content, subject.build_tests
+
+       mock_repository.verify
+       mock_minitest_file.verify
+       mock_template_file.verify
```
Insti: There are too many things being asserted/verified in all the tests that have changed in this commit.
Refactor later.

test/generator/exercise_test.rb
```diff
@@ -43,73 +43,9 @@ def test_update_example_solution
...
-       assert_equal expected_content, subject.build_tests
-
-       mock_repository.verify
-       mock_minitest_file.verify
-       mock_template_file.verify
-     end
-    end
+#   will replace later
+#    def test_build_tests
+#    end
```
Insti: If it's temporary, I'd rather rename rather than delete.
Renaming to 'temporarily_disabled_test_build_tests` will prevent it from running, but still retain all the information without having to a) remember it was there, b) find it in the git history.

lib/generator/command_line.rb
```diff
@@ -35,13 +35,17 @@ def freeze?
...
+    def exercise(repository)
       LoggingExercise.new(
-        exercise: Exercise.new(paths: paths, slug: slug),
+        exercise: Exercise.new(repository: repository),
         logger: logger
       )
     end
 
+    def repository(slug)
+      Repository.new(paths: paths, slug: slug)
```
Insti: Bookmark

test/generator/exercise_test.rb
```diff
@@ -2,35 +2,43 @@
...
+      repository = Repository.new(paths: FixturePaths, slug: 'alpha')
+      # Verify iniital condition from fixture file
+      assert_equal 1, repository.tests_version.to_i
+
+      mock_file = Minitest::Mock.new.expect :write, '2'.length, [2]
+      subject = Exercise.new(repository: repository)
+      File.stub(:open, true, mock_file) do
+        assert_equal 2, subject.update_tests_version
+      end
+      mock_file.verify
```
Insti: Too much going on in these tests.
They should be able to mock repository now and not going all the way down to File.
That belongs in Repository tests.

test/generator/template_values_test.rb
```diff
@@ -71,30 +71,10 @@ def paths
       include TemplateValuesFactory
     end
 
-    class ClassBasedTestTemplateValuesFactory
+    class ClassBasedTestTemplateValuesFactory < TestTemplateValuesFactory
```
Insti: Don't DRY tests.

test/generator/exercise_test.rb
```diff
@@ -39,21 +39,65 @@ def test_update_example_solution
       mock_file.verify
     end
 
+    def test_create_tests_file
```
Insti: Ok it's back, good. Still think the rename is preferable.
Also leads to quieter diffs.

Insti:
There appears to still be some confusion about the responsibilities of Exercise and Repository.

So the last 5 commits: https://github.com/exercism/xruby/pull/638/files/2124b5712c6eff904b02427b2151b28035d1d5cd..e9421970140d8a69c7f1455d36c05eb9086d8346

I think are heading in the wrong direction.

(I will write more on this later.)

lib/generator/repository.rb
```diff
@@ -0,0 +1,16 @@
+module Generator
+  class Repository
+    include Files::TrackFiles
+    include Files::MetadataFiles
+
+    def initialize(paths:, slug:)
```
Insti: Slug should not be a property of Repository.

lib/generator/repository.rb
```diff
@@ -0,0 +1,16 @@
...
+      @paths = paths
+      @slug = slug
+    end
+    attr_reader :paths, :slug
+
+    def name
```
Insti: I'm not sure:
a) if this should be public.
b) if this is the right name.

test/generator/exercise_test.rb
hilary: I actually had that version of these tests, but ran into issues with the mocking. I may try a second version...

hilary:
I'm going to take another thwack at this, splitting into three models....

PR #640, opened 2017-05-14
[WIP] - Separate canonical data methods from ExerciseCase

Insti:
> ExerciseCase should not be an OpenStruct and the canonical data should be kept separate from the methods of this class.

from Generator Improvements exercism/xruby#485.

Completes #622 to a usable state.

Outstanding issue: Update all the *_cases files to use the `canonical.` version.
These can easily be identified by emitting a warning whenever the forward method of ExerciseCase is called.

### DIFF
--- a/lib/generator/exercise_case.rb
+++ b/lib/generator/exercise_case.rb
@@ -6,13 +6,13 @@ module Generator
     include CaseHelpers
     include Assertion
 
-    attr_reader :case_data
-    def initialize(case_data:)
-      @case_data = case_data
+    attr_reader :canonical
+    def initialize(canonical:)
+      @canonical = canonical
     end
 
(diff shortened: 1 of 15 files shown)


lib/generator/exercise_case.rb
```diff
@@ -18,5 +19,17 @@ def name
     def skipped(index)
       index.zero? ? '# skip' : 'skip'
     end
+
+    private
+
+    def setup_forwards(data)
+      data.to_h.keys.each do |key|
+        send(:define_singleton_method, key) { forward(key) }
+      end
```
Insti: `method_missing` is a much nicer way to handle this.

Insti:
Superseded by #648

PR #641, opened 2017-05-14
[WIP] generator: refactor implementation, part 1

hilary:
I did this PR in two parts. The first part (this one) separates out the Exercise model.

The second part (two different PRs) separates out the Repository model in two different ways. One PR creates a Repository which is instantiated with paths only. The second PR creates a Repository which is instantiated with paths and exercise.

### DIFF
--- a/lib/generator.rb
+++ b/lib/generator.rb
@@ -15,7 +15,7 @@ module Generator
   # Doesn't update the version information.
   class GenerateTests < ImplementationDelegator
     def call
-      create_tests_file
+      build_tests
     end
   end
 
@@ -24,7 +24,7 @@ module Generator
     def call
       update_tests_version
       update_example_solution
-      create_tests_file
+      build_tests
     end
   end
 end
--- a/lib/generator/files/generator_cases.rb
+++ b/lib/generator/files/generator_cases.rb
@@ -8,10 +8,6 @@ module Generator
           end.sort
         end
 
-        def class_name(exercise_name_or_slug)
-          filename(exercise_name_or_slug).split('_').map(&:capitalize).join
-        end
-
         def source_filepath(track_path, slug)
           path = meta_generator_path(track_path, slug)
           filename = filename(slug) + '.rb'
--- a/lib/generator/files/generator_cases.rb
+++ b/lib/generator/files/generator_cases.rb
@@ -3,7 +3,9 @@ module Generator
     module GeneratorCases
       class << self
         def available(track_path)
-          cases_filepaths(track_path).map { |filepath| slugify(filepath) }.sort
+          filepaths(track_path).map do |filepath|
+            %r{#{track_path}/exercises/([-a-z]+)/}.match(filepath)[1]
+          end.sort
         end
 
         def class_name(exercise_name_or_slug)
@@ -18,13 +20,8 @@ module Generator
 
         private
 
-        def cases_filepaths(track_path)
-          generator_glob = File.join(meta_generator_path(track_path, '*'), '*_case.rb')
-          Dir.glob(generator_glob, File::FNM_DOTMATCH)
-        end
-
-        def slugify(filepath)
-          File.basename(filepath, '_case.rb').tr('_', '-')
+        def filepaths(track_path)
+          Dir.glob(meta_generator_path(track_path, '*'))
         end
 
         def filename(exercise_name_or_slug)
--- a/lib/generator/template_values.rb
+++ b/lib/generator/template_values.rb
@@ -26,16 +26,12 @@ module Generator
         abbreviated_commit_hash: canonical_data.abbreviated_commit_hash,
         canonical_data_version: canonical_data.version,
         version: version,
-        exercise_name: slug_underscore,
+        exercise_name: exercise.name,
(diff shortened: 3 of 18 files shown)


test/generator/implementation_test.rb
```diff
@@ -12,11 +12,6 @@ def test_version
       assert_equal 1, subject.version
     end
 
-    def test_slug
-      subject = Implementation.new(paths: FixturePaths, slug: 'alpha')
-      assert_equal 'alpha', subject.slug
-    end
```
Insti: This is not a test of core ruby functionality, it is a test that Implementation responds to slug.

lib/generator/underscore.rb
```diff
@@ -4,6 +4,10 @@ module Underscore
       def underscore
         downcase.gsub(/[- ]/, '_').gsub(/[^\w?]/, '')
       end
+
+      def camel_case
+        underscore.split('_').map(&:capitalize).join
+      end
```
Insti: * I'm sceptical of any change that is an "add" as part of a 'refactor' commit.
* This change makes the class name `Underscore` inaccurate.

lib/generator/files/generator_cases.rb
```diff
@@ -3,7 +3,9 @@ module Files
     module GeneratorCases
       class << self
         def available(track_path)
-          cases_filepaths(track_path).map { |filepath| slugify(filepath) }.sort
```
Insti: I strongly disagree with this change. 

It is true that available does not need to know exactly how a test case is generated, and you could achieve this by abstracting out the availability criteria, but since that is all this class does that seems excessive.

> because that might go away or get changed. 

This is a case of YAGNI and we can worry about that when it happens.

See also: https://github.com/exercism/xruby/pull/646 which aims to give new generator implementers helpful advice.

test/generator/exercise_test.rb
```diff
@@ -0,0 +1,20 @@
...
+module Generator
+  class ExerciseTest < Minitest::Test
+    def test_slug
+      exercise = Exercise.new(slug: 'alpha')
+      assert_equal 'alpha', exercise.slug
+    end
+
+    def test_name
+      exercise = Exercise.new(slug: 'alpha-beta')
+      assert_equal 'alpha_beta', exercise.name
```
Insti: Yay for tests.
I'm not sure `name` is the right name for this property, but it is consistent with what we've been using.

test/generator/template_values_test.rb
```diff
@@ -71,30 +71,10 @@ def paths
       include TemplateValuesFactory
     end
 
-    class ClassBasedTestTemplateValuesFactory
+    class ClassBasedTestTemplateValuesFactory < TestTemplateValuesFactory
```
Insti: You like tests to be way more DRY than I do.
I'm not sure this is a good idea.
(But it's not wrong.)

test/generator/template_values_test.rb
```diff
@@ -71,17 +71,6 @@ def paths
       include TemplateValuesFactory
     end
 
-    class ClassBasedTestTemplateValuesFactory < TestTemplateValuesFactory
```
Insti: It goes away anyway. Yay.

test/generator/template_values_test.rb
```diff
@@ -8,36 +8,12 @@ def setup
       }
     end
 
-    def test_abbreviated_commit_hash
```
Insti: Again, not core ruby functionality.
That it is implemented using an attr_reader is irrelevant.

Request: omit this commit.

lib/generator/template_values.rb
```diff
@@ -15,8 +17,12 @@ def get_binding
       binding
     end
 
+    def exercise_name
+      exercise.name
+    end
+
     def exercise_name_camel
-      exercise_name.split('_').map(&:capitalize).join
+      exercise.name.camel_case
```
Insti: `slug` is the canonical value for problem identity.
This should be `exercise.slug.camel_case`
But either way it's a Demeter violation, so should be just `exercise.classname`

`exercise_name_camel` is a slightly misleading method name here given it's usage is to determine the test class name used in the template.

lib/generator/template_values.rb
```diff
@@ -3,20 +3,27 @@ module Generator
   class TemplateValues
```
Insti: The original intention of `TemplateValues` was that it it should be a dumb container and would only contain the template values available to the erb file. The `initialize` method provided one place where it was possible to see exactly all the properties that would be available to the erb. 

It is the responsibility of the `TemplateValuesFactory` to construct the necessary values.

If that is the goal, then there is probably a nicer way of implementing this but I'm not sure this commit is heading in that direction.

lib/generator.rb
```diff
@@ -15,7 +15,7 @@ def initialize(track:, metadata:)
   # Doesn't update the version information.
   class GenerateTests < ImplementationDelegator
     def call
-      create_tests_file
```
Insti: I like this 👍

Insti:
I've extracted the best parts from this PR in: #677 and will continue the refactoring from there.

PR #642, opened 2017-05-14
[WIP] generator: part 2 of refactor implementation model

hilary:
part 2 of exercism/xruby#641.

This version instantiates Repository with paths only.


Insti:
This is not really what I had in mind, I guess I just communicated my idea poorly.

I agree with you that this is not good.
Thanks for doing it though :heart:

PR #583, opened 2017-04-30
space-age: Add generator

ajwann:
## Why?

space-age exercise needs a generator

## What?

added generator for space-age

## How Has This Been Tested?

rake test:space-age
rubocop -D

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or enhancement that would cause existing functionality to change)
- [ ] Gardening (code and/or documentation cleanup)

## References and Closures

references #396

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [ ] My change relies on a pending issue/pull request
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #646, opened 2017-05-15
Be more helpful when generator can not be found

### LINKED ISSUE #644: generator: more helpful error message when not found.
In #583 there was a problem with the generator not being found:

```
$ bin/generate space-age
A generator does not currently exist for space-age!
```

It would be nice if this error message gave some more information about where it expected to find the files it needs.

Insti:
Resolves #644 

In #583 there was a problem with the generator not being found:

```
$ bin/generate space-age
A generator does not currently exist for space-age!
```

This patch adds a bit more information to the error message about where the generator expected to find the file it needs.

```
A generator does not currently exist for space-age!
Expecting it to be at: <user_specific_path>/exercises/space-age/.meta/generator/space_age_case.rb
```

### DIFF
--- a/lib/generator/command_line/generator_optparser.rb
+++ b/lib/generator/command_line/generator_optparser.rb
@@ -83,9 +83,9 @@ module Generator
 
     def validate_cases
       return true if available_generators.include?(options[:slug])
-      error_message = "A generator does not currently exist for #{options[:slug]}!"
+      warning = "A generator does not currently exist for #{options[:slug]}!"
       expected_locations = "Expecting it to be at: #{Files::GeneratorCases.source_filepath(@paths.track, options[:slug])}"
-      $stderr.puts [error_message, expected_locations].join("\n")
+      $stderr.puts [warning, expected_locations].join("\n")
       false
     end
   end
--- a/lib/generator/command_line/generator_optparser.rb
+++ b/lib/generator/command_line/generator_optparser.rb
@@ -83,7 +83,9 @@ module Generator
 
     def validate_cases
       return true if available_generators.include?(options[:slug])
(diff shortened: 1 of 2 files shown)


lib/generator/command_line/generator_optparser.rb
```diff
@@ -83,7 +83,9 @@ def validate_exercise
 
     def validate_cases
       return true if available_generators.include?(options[:slug])
-      $stderr.puts "A generator does not currently exist for #{options[:slug]}!"
+      error_message = "A generator does not currently exist for #{options[:slug]}!"
+      expected_locations = "Expecting it to be at: #{Files::GeneratorCases.source_filepath(@paths.track, options[:slug])}"
+      $stderr.puts [error_message, expected_locations].join("\n")
```
kotp: This isn't raising so `error_message` should probably be simply `message`.

kotp:
Looks good to me, other than the variable name.  I looked for where it was raising, and couldn't find it, so it caused me some extra "investigation" to realize that it was not going to do that... the name leads me to search, without it, I know it is just a message with no further expectations, other than perhaps output somewhere.

lib/generator/command_line/generator_optparser.rb
```diff
@@ -83,7 +83,9 @@ def validate_exercise
 
     def validate_cases
       return true if available_generators.include?(options[:slug])
-      $stderr.puts "A generator does not currently exist for #{options[:slug]}!"
+      warning = "A generator does not currently exist for #{options[:slug]}!"
+      expected_location = "Expecting it to be at: #{Files::GeneratorCases.source_filepath(@paths.track, options[:slug])}"
```
kotp: Good catch

kotp:
Thanks @Insti

PR #647, opened 2017-05-15
Add missing skip in trinary_test

nathanbwright:
I noticed that the trinary test didn't automatically skip the book keeping test.

This pull request fixes that.


Insti:
Thanks @nathanbwright :heart:

PR #648, opened 2017-05-18
Separate canonical data methods from ExerciseCase

Insti:
> ExerciseCase should not be an OpenStruct and the canonical data should be kept separate from the methods of this class.

from Generator Improvements exercism/xruby#485.

Supersedes #640 (Closed) 
Depends on #650 (Closed/Merged)
Completes #622 (Closed) to a usable state.

PR #650, opened 2017-05-18
Hilary's ExerciseCase refactorings

Insti:
These refactorings are not directly relevant to separating the canonical data methods from ExerciseCase so I've separated them from the changes required for #648


Insti:
FYI:
I created a separate branch for these because I was trying to work out how to more elegantly build on PRs, and the answer seems to be branches. 
I've not updated anything else because I don't want to re-do the PRs I've already done.

PR #652, opened 2017-05-21
Test Fixnum#underscore

Insti:
https://github.com/exercism/xruby/pull/583 Added a refinement to `Fixnum` to convert integers into their underscored string representation.

 This PR adds a unit test to bring us back up to 100% test coverage.

### DIFF
--- a/test/generator/underscore_test.rb
+++ b/test/generator/underscore_test.rb
@@ -21,7 +21,7 @@ module Generator
     end
 
     def test_fixnum_underscore
-      assert_equal '1_000_000', 1_000_000.underscore
+      assert_equal '1_000_000', 1000000.underscore
     end
   end
 end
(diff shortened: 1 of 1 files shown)


test/generator/underscore_test.rb
```diff
@@ -19,5 +19,9 @@ def test_question_mark
         'unreadable_but_correctly_sized_inputs_return_?'
       )
     end
+
+    def test_fixnum_underscore
+      assert_equal '1_000_000', 1_000_000.underscore
```
kotp: The tests should test from a "raw" number, such as `1000000` rather than one with underscores, otherwise why would the call be needed?  (Yes, it is needed to also convert it to a string... but this is not clear in the name, especially since the number already has underscores that it is being called on.)

PR #656, opened 2017-05-24
Add the "first patch" labels to exempt from stale

kotp:
@tejasbuane had the idea of keeping the these accepted work units to be
kept from becoming stale.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -8,6 +8,8 @@ daysUntilClose: 7
 exemptLabels:
   - pinned
   - security
+  - good first patch
+  - first-timers only
 # Label to use when marking as stale
 staleLabel: stale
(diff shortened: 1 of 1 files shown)


Insti:
Is this a good idea? 
If they are marked stale by probot we should review them to see if they are still relevant.

kotp:
That is why it is here.  Unsure if it is a good idea.  But all first_time style issues should be reviewed periodically anyway, as complications from other developments could move it from an 'easier' change to something that is no longer appropriate for first time pull requests.

Insti:
Ok, well I vote no.
Let's have probot remind us about the things we might not have looked at for a while.

kotp:
That works... And Probot should definitely help with that.

PR #657, opened 2017-05-25
bob: added generator

ajwann:
references #396


exercises/bob/.meta/generator/test_template.erb
```diff
@@ -0,0 +1,28 @@
...
+
+# Common test data version: <%= canonical_data_version %> <%= abbreviated_commit_hash %>
+class <%= exercise_name_camel %>Test < Minitest::Test
+  def bob
+    Bob
+  end
+
+  def feedback(text)
+    "Bob hears #{text.inspect}, and.."
+  end
```
Insti: This really belongs in the generator, the feedback should be expanded before it gets added to the test.

exercises/bob/.meta/generator/test_template.erb
```diff
@@ -0,0 +1,28 @@
...
+
+<% test_cases.each_with_index do |test_case, idx| %>
```
Insti: Use `index` instead of `idx`

exercises/bob/.meta/generator/test_template.erb
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env ruby
+require 'minitest/autorun'
+require_relative '<%= exercise_name %>'
+
+# Common test data version: <%= canonical_data_version %> <%= abbreviated_commit_hash %>
+class <%= exercise_name_camel %>Test < Minitest::Test
+  def bob
+    Bob
+  end
```
Insti: What is the point of this?

Insti:
Looking good @ajwann, you're getting good at these.

exercises/bob/.meta/generator/test_template.erb
kotp: I agree with @insti use either `index` or use `i`.  I prefer `i` in block argument lists.

exercises/bob/.meta/generator/test_template.erb
ajwann: I honestly just tried to reproduce the test pretty much in it's entirety. I agree this method is especially useless. I'm happy to remove it forever.

exercises/bob/.meta/generator/test_template.erb
ajwann: I'll move it where it belongs. I also don't think we need to call ```.inspect``` if we are string interpolating it in the generator.

exercises/bob/.meta/generator/test_template.erb
ajwann: I'm totally fine with either. Index is more explicit, so I'll go with that.

ajwann:
@Insti I made changes based on you and @kotp's feedback. I also got rid of the call to ```.inpsect``` altogether when interpolating the input in the ```workload``` method. Unless I'm missing something it's not necessary anymore.

exercises/bob/bob_test.rb
```diff
@@ -2,133 +2,184 @@
...
+    skip
+    remark = "Okay if like my  spacebar  quite a bit?   "
+    assert_equal 'Sure.', Bob.hey(remark), "Bob hears Okay if like my  spacebar  quite a bit?   , and.."
+  end
+
+  def test_other_whitespace
+    skip
+    remark = "
+
 	"
```
kotp: Someones newlines are not being managed in their git configuration?

exercises/bob/bob_test.rb
```diff
@@ -2,133 +2,184 @@
...
+    assert_equal 'Fine. Be that way!', Bob.hey(remark), "Bob hears           , and.."
   end
 
-  def test_alternate_silences
+  def test_alternate_silence
     skip
-    remark = "\t" * rand(1..10)
-    assert_equal 'Fine. Be that way!', bob.hey(remark), feedback(remark)
+    remark = "										"
+    assert_equal 'Fine. Be that way!', Bob.hey(remark), "Bob hears 										, and.."
```
Insti: The things that Bob hears should be quoted.
`%Q` can be good for this to help avoid needing to quote quotes.

Insti:
The `inspect`s are essential in this exercise due to all the wacky whitespace characters present in the canonical data.

exercises/bob/bob_test.rb
Insti: There's an explicit "\n\r"(!) in the test data.

kotp: I honestly think that working with the various newlines that we have to deal with is an important thing... but why in this exercise.  Not appropriate to press back here though... should be addressed in x-common.

ajwann:
I think we are good to go here.

exercises/bob/bob_test.rb
```diff
@@ -2,133 +2,184 @@
...
+    assert_equal 'Fine. Be that way!', Bob.hey(remark), %Q(Bob hears           , and..)
   end
 
-  def test_alternate_silences
+  def test_alternate_silence
     skip
-    remark = "\t" * rand(1..10)
-    assert_equal 'Fine. Be that way!', bob.hey(remark), feedback(remark)
+    remark = %Q(										)
+    assert_equal 'Fine. Be that way!', Bob.hey(remark), %Q(Bob hears 										, and..)
```
Insti: Still needs quotes around the string Bob hears. (input)

exercises/bob/.meta/generator/bob_case.rb
```diff
@@ -0,0 +1,9 @@
+require 'generator/exercise_case'
+
+class BobCase < Generator::ExerciseCase
+  def workload
+    indent_lines(["remark = %Q(#{input})",
```
Insti: Just use `inspect` here rather than `%Q`

exercises/bob/.meta/generator/bob_case.rb
```diff
@@ -0,0 +1,9 @@
...
+      "assert_equal '#{expected}', Bob.hey(remark), %Q(Bob hears #{input}, and..)"
```
Insti: `%Q` is right to use here though.

```ruby
"assert_equal '#{expected}', Bob.hey(remark), %Q(Bob hears #{input.inspect}, and..)"
```

exercises/bob/bob_test.rb
```diff
@@ -2,133 +2,184 @@
...
+    skip
+    remark = %Q(Okay if like my  spacebar  quite a bit?   )
+    assert_equal 'Sure.', Bob.hey(remark), %Q(Bob hears Okay if like my  spacebar  quite a bit?   , and..)
+  end
+
+  def test_other_whitespace
+    skip
+    remark = %Q(
+
 	)
```
Insti: Using inspect should also make the silly whitespace characters usefully visible.

exercises/bob/.meta/generator/bob_case.rb
ajwann: @Insti Adding ```inspect``` here actually causes several tests to close the %Q() call early,because those test cases contain literal ```)``` in the actual values. This causes a syntax error on line 101 of the generated test.

If I do something like ```'Bob hears #{input.inspect}, and..'``` then I get the same issue because certain output contains ```'```. 

I'm basically trapped in a big catch 22 of string interpolation.

kotp: :eyes:

I will look soon if @insti doesn't beat me to it.

kotp: Try doing the same with the `%Q` but instead use `%\Q` to escape it.  That should reconstruct the executable line, but ignore it in place.

Let me know.

Insti: %Q can use different delimiters.
https://ruby-doc.org/core-2.2.0/doc/syntax/literals_rdoc.html#label-Percent+Strings

Insti: Would this work?
```ruby
  def workload
    indent_lines(["remark = #{input.inspect}",
      "assert_equal '#{expected}', Bob.hey(remark), %q{Bob hears #{input.inspect}, and..}"
    ], 4)
  end
```

Insti: lowercase `q` is better since we don't want it interpolating the string. (the second time)

ajwann: @Insti using ```%q``` with curly brackets insted of parens did the trick. I had no idea you could use alternative delimiters, thanks!

ajwann:
The calls to ```inspect``` are back and the whitespace characters now show up when the test is generated.

exercises/bob/.meta/generator/test_template.erb
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env ruby
+require 'minitest/autorun'
+require_relative '<%= exercise_name %>'
+
+# Common test data version: <%= canonical_data_version %> <%= abbreviated_commit_hash %>
+class <%= exercise_name_camel %>Test < Minitest::Test
```
Insti: This looks just like the default test template, are there any changes or can it be deleted?

ajwann: No changes, It got copied when I copied the ```generator``` directory. I'll get rid of it.

Insti:
Thanks for all your work on this @ajwann :heart: ⭐️

PR #658, opened 2017-05-25
sum-of-multiples: added generator

ajwann:
references #396


exercises/sum-of-multiples/sum_of_multiples_test.rb
```diff
@@ -2,48 +2,87 @@
...
-  def test_configurable_43_47_to_10000
+  def test_multiples_of_5_or_25_up_to_51
     skip
-    assert_equal 2_203_160, SumOfMultiples.new(43, 47).to(10_000)
+    assert_equal 275, SumOfMultiples.new(5, 25).to(51)
+  end
+
+  def test_multiples_of_43_or_47_up_to_10000
+    skip
+    assert_equal 2203160, SumOfMultiples.new(43, 47).to(10000)
```
Insti: It would be good to use number underscoring with big numbers.

exercises/sum-of-multiples/.meta/generator/sum_of_multiples_case.rb
```diff
@@ -0,0 +1,8 @@
+require 'generator/exercise_case'
+
+class SumOfMultiplesCase < Generator::ExerciseCase
+  def workload
+    indent_lines(["assert_equal #{expected}, SumOfMultiples.new" \
+                  "(#{factors.join(', ')}).to(#{limit})"] , 4)
```
Insti: 1) I like to try really hard to avoid line continuations like this.
2) Did you know `ExerciseCase` has an `assert_equal` helper method?

exercises/sum-of-multiples/sum_of_multiples_test.rb
```diff
@@ -2,48 +2,87 @@
 require 'minitest/autorun'
 require_relative 'sum_of_multiples'
 
-class SumTest < Minitest::Test
-  def test_sum_to_1
+# Common test data version: 1.0.0 72b1496
+class SumOfMultiplesTest < Minitest::Test
+  def test_multiples_of_3_or_5_up_to_1
+    # skip
     assert_equal 0, SumOfMultiples.new(3, 5).to(1)
```
Insti: Probably unrelated to generator, but would it be interesting to allow `SumOfMultiples` to return something more Enumerator-like? 
`SumOfMultiples.new(3,5).take(1)`
cc: @kotp

Insti:
Nice 👍

exercises/sum-of-multiples/sum_of_multiples_test.rb
Insti: I'm really liking the idea of changing the method name from `to` to `take`

exercises/sum-of-multiples/.meta/generator/sum_of_multiples_case.rb
ajwann: Nice! I'm a big fan of the new ```assert_equal``` method accepting a block.

ajwann:
@Insti I was able to add underscoring at the cost of using the ```assert_equal``` helper. See my comment above for more details. I'm happy to tweak the string concatenation in my ```workload``` method to make it avoid the line continuation if you'd like.

exercises/sum-of-multiples/.meta/generator/sum_of_multiples_case.rb
Insti: Hmm interesting. Thanks.

The whole assertion helper thing is quite new and has many unexplored edge cases.

Creating an issue about this might be a good idea.

Insti:
Thanks @ajwann The other changes look good.

>  I'm happy to tweak the string concatenation in my workload method to make it avoid the line continuation if you'd like.

This would be super 👍

exercises/sum-of-multiples/.meta/solutions/sum_of_multiples.rb
```diff
@@ -1,10 +1,14 @@
...
+end
+
 class SumOfMultiples
   attr_reader :multiples
   def initialize(*multiples)
     @multiples = multiples
   end
 
-  def to(limit)
+  def take(limit)
```
kotp: What inspired the method name change?

Take a limit is a different idea potentially than to a limit.

The "to" implies an exclusive limit, while take suggests an inclusive limit.  An example of this in English is "go to the wall" which implies do not go all the way into the wall, or on the wall, just approach it as close as you can.  The Readme is clear that we are to exclude the limit.

Insti: https://github.com/exercism/xruby/pull/658#discussion_r118641355

Insti: Although I had forgotten that the limit was excluded, so the name change may be inappropriate.

Insti: `take` is also not good since we're not actually taking an array, we just want the sum.

I think I've convinced myself that the method name change was a bad idea.

ajwann: Haha sounds good. I'll add the line break and *take* out the new method name. 😂

ajwann:
@Insti @kotp, I think we are good to go here. I renamed ```take``` back to ```to``` and got rid of the line continuation.

exercises/sum-of-multiples/.meta/generator/sum_of_multiples_case.rb
```diff
@@ -0,0 +1,11 @@
+require 'generator/exercise_case'
+
+class SumOfMultiplesCase < Generator::ExerciseCase
+  using Generator::Underscore
+
+  def workload
+    assert_expected = "assert_equal #{expected.underscore}"
+    value = "SumOfMultiples.new(#{factors.join(', ')}).to(#{limit})"
+    indent_lines(["#{assert_expected}, #{value}"], 4)
```
Insti: Feedback rather than a change request:

The grouping here is logically confusing as in `assert_expected` you have only 2/3 of a valid `assert_equal` call. 
`value` is good.
constructing the string at the end is ok due to your variable naming, but confusing because `assert_expected` looks like a minitest method but it isn't one.


I would have done this:
```
result = "SumOfMultiples.new(#{factors.join(', ')}).to(#{limit})"
indent_lines(["assert_equal #{expected.underscore}, #{result}"], 4)
```

kotp:
Bringing it in, other than the [mentioned code](https://github.com/exercism/xruby/pull/658#discussion_r120541553) by @Insti it looks good to me.  We can change that up if it ends up being an irritant.  I can see it being that way, on the next round.  I can also see the `underscore` name changing on the same line at some point to `snake_case` instead, as I find that name a little bit meh, too.  Just not irritated enough to change it.

kotp:
Thank you @insti and @ajwann good job as always!

PR #660, opened 2017-05-28
Update license information

kytrinyx:
We don't need the duplication, especially now that the GitHub interface shows the
license information on the main page of the repository when it can be detected
directly from the LICENSE file.

This also updates the LICENSE to reassign the copyright to the Exercism legal entity (which I finally got my act together and created).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
 
-Copyright (c) 2014 Katrina Owen, _@kytrinyx.com
+Copyright (c) 2017 Exercism, Inc
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/README.md
+++ b/README.md
@@ -252,11 +252,6 @@ For an in-depth discussion of how exercism language tracks and exercises work,
 please see the
 [contributing guide](https://github.com/exercism/x-api/blob/master/CONTRIBUTING.md#the-exercise-data)
 
-## License
-
-The MIT License (MIT)
-
-Copyright (c) 2014 Katrina Owen, _@kytrinyx.com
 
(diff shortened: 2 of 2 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
 
-Copyright (c) 2014 Katrina Owen, _@kytrinyx.com
+Copyright (c) 2017 Exercism, Inc
```
kotp: The `(c)` form of **&#169;** is used when the typsetting equipment does not have it available, and is even then not necessarily recognized legally as indicating copyright notice.  We have the typesetting available and should use the officially/legally recognized symbol, or `copr.` or `copyright`.

My suggestion then is to in the word, and leave out the `(c)`.

In case of concerns for ASCII characters, this is character `0169`.

kytrinyx: I have taken this directly from https://choosealicense.com/licenses/mit/ which has both the word "copyright" and the "(c)".

kytrinyx: According to the wikipedia page, the approximation (c) appears to be acceptable.
https://en.wikipedia.org/wiki/Copyright_symbol#Digital_representation

kotp: Upper case C then?  Since 1989 and for those countries honoring the convention, none are required, it appears.

kotp: It was the only thing I saw... IANAL but that was the only thing I had noticed (a recent attorney's note had mentioned the non-legal distinction of the `(c)` but I can not find that note at the moment, and it is likely based on some outdated things, apparently.)

kytrinyx: Since this is what they're using on the choosealicense website, I'd prefer to keep it the way they did it rather than to alter the case.

kotp: Good solid thing to do, I think.

PR #664, opened 2017-05-31
Isogram Exercise improvements: change method name/add failure messages

deohtee:
## Why?

- Improves the Isogram Exercise by adding failure messages and simplifying the method name 
- Helps bring the exercise up to standard

## What?

- Added failure messages to tell you what value was expected and what input was tested
- Shortened the method name from `is_isogram?` to `isogram?`
- Changed from version `3` to version  #`4`

## How Has This Been Tested?

- By generating the test for this exercise by running `bin/generate -f isogram' and then
testing the generated test. 

## Screenshots (if appropriate):
screen shot 2017-05-31 at 11 53 27 am

## Types of changes

- n/a Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (non-breaking change which adds functionality)
- [x] Breaking change (fix or enhancement that would cause existing functionality to change)
- n/a Gardening (code and/or documentation cleanup)

**Picked breaking change too because the method name changed so if you went back and ran the 
version 3 solutions against this version they would fail.**

## References and Closures

references #475

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] My change relies on a pending issue/pull request
- n/a My change requires a change to the documentation.
- n/a I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #669, opened 2017-06-08
isogram: update solution to pass current tests.

Insti:
#664 updated the `isogram` tests but not the example solution.
For some reason Travis CI, which would have detected this, did not run on that PR.

This patch updates the example solution to pass the updated tests.

### DIFF
--- a/exercises/isogram/.meta/solutions/isogram.rb
+++ b/exercises/isogram/.meta/solutions/isogram.rb
@@ -1,9 +1,9 @@
 module BookKeeping
-  VERSION = 3
+  VERSION = 4
 end
 
 class Isogram
-  def self.is_isogram?(str)
+  def self.isogram?(str)
     letters = str.downcase.gsub(/[[:punct:]]| /, '').chars
     letters == letters.uniq
   end


exercises/isogram/.meta/solutions/isogram.rb
```diff
@@ -1,9 +1,9 @@
 module BookKeeping
-  VERSION = 3
+  VERSION = 4
 end
 
 class Isogram
-  def self.is_isogram?(str)
+  def self.isogram?(str)
```
kotp: Do use "string" as a name.   Better use phrase or word or whatever the most general "language" thing is that is wanted, rather than a programming thing.

Yet, even without that change, the example is not shown actively, so it would be acceptable to the "the problem is solvable in this language" goal.

deohtee:
Sorry I should have caught that as well, that's my fault.

kotp:
@deohtee are you running the git hooks when you develop these?

exercises/isogram/.meta/solutions/isogram.rb
Insti: What?

kotp: The argument name.

Insti: Oh, you were commenting on an aspect of the example solution that has not been changed as part of this PR.

This PR is just trying to make the existing example pass the current tests, if you want to improve the example solution this can be done in another PR.

deohtee:
I am not. Is there any documentation on it? I tried looking for some just now but couldn't find any ☹️

Insti:
@kotp?

The git hooks are not necessary to trigger Travis CI - only for local checks.

Insti:
@deohtee the git hook documentation is here: https://github.com/exercism/xruby#pull-requests

kotp:
Right... the hooks locally though can help contributors confirm the tests...

PR #670, opened 2017-06-09
Add book keeping to Bob problem test

BigChief45:
The Bob problem is missing Book Keeping.

### DIFF
--- a/exercises/bob/bob_test.rb
+++ b/exercises/bob/bob_test.rb
@@ -131,4 +131,25 @@ Does this cryogenic chamber make me look fat?
 no)
     assert_equal 'Whatever.', bob.hey(remark), feedback(remark)
   end
+
+  # Problems in exercism evolve over time, as we find better ways to ask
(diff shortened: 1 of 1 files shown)


Insti:
Thanks @BigChief45 but I am sorry to say that we will not be merging this change.

`BookKeeping` is only included when tests have a test generator.  

See: https://github.com/exercism/xruby#anatomy-of-an-exercise for more information about test generators. (Sorry that the use of `BookKeeping` is not documented.)

Normally I would suggest you try making a test generator, but there is already an almost complete pull request to add one for this exercise: https://github.com/exercism/xruby/pull/657

Thanks for the contribution. Please make another if you find anything else that could be improved. :heart:

BigChief45:
@Insti No problem, thanks for the explanation.

Insti:
@BigChief45 if you wanted to, you could help update the documentation so that the next person does not encounter the same problem as you did: See issue #671.

PR #677, opened 2017-06-13
generator: refactor implementation, part 1 highlights.

Insti:
In #641 - generator: refactor implementation, part 1, Hilary started some generator refactorings.  Progress on this seems to have stopped.

In order to continue the refactoring I have extracted from the uncontroversial changes from the initial PR. 

I have further built upon these changes to continue the refactoring. (To follow in further PRs.)

### DIFF
--- a/test/generator/files/track_files_test.rb
+++ b/test/generator/files/track_files_test.rb
@@ -11,10 +11,9 @@ module Generator
       class TestTrackFiles
         def initialize
           @paths = FixturePaths
-          @slug = 'alpha-beta'
-          @exercise_name = 'alpha_beta'
+          @exercise = Exercise.new(slug: 'alpha-beta')
         end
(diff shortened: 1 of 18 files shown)


test/generator/files/track_files_test.rb
```diff
@@ -44,9 +43,9 @@ def test_tests_template
       class TestTrackFilesUseDefault
         def initialize
           @paths = FixturePaths
-          @slug = 'notemplate'
+          @exercise = Exercise.new(slug: 'notemplate')
```
kotp: Is there room here to use `'no template'` or if necessary `'no_template'`?

Insti: Probably. It would need to be `no-template`. This will need to happen later though as I'm trying not to mess with the pre-existing commits.

PR #643, opened 2017-05-14
[WIP] generator: part 2 (version b) of refactor implementation model

hilary:
The second version of part 2 of exercism/xruby#641

This version instantiates Repository with paths and exercise.

PR #678, opened 2017-06-13
generator: refactor implementation, part 2 - extract Repository

Insti:
Extract a `Repository` class that takes `paths` and `slug` as constructor arguments.

Repository in this case is an implementation of the Repository pattern which acts as an interface to all the data that an `Implementation` needs rather than anything related to Git repositories.

Continues the work done in #677, 
Supersedes  #643

### DIFF
--- a/test/generator/files/track_files_test.rb
+++ b/test/generator/files/track_files_test.rb
@@ -39,6 +39,11 @@ module Generator
         assert_equal expected_filename, subject.tests_template.filename
       end
 
+      def test_test_case
+        subject = TestTrackFiles.new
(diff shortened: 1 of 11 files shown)


Insti:
Why is Rubocop complaining, I it was disabled. :/

test/generator/files/track_files_test.rb
```diff
@@ -43,9 +43,9 @@ def test_tests_template
       class TestTrackFilesUseDefault
         def initialize
           @paths = FixturePaths
-          @exercise = Exercise.new(slug: 'notemplate')
+          @slug = 'no-template'
```
Insti: Added a hyphen so we don't need to worry about whether it's an `emplate` or not.

PR #682, opened 2017-06-20
hello-world/hamming: Separate GETTING_STARTED and RUNNING_TESTS instructions.

### LINKED ISSUE #681: hello_world: GETTING_STARTED.md references tests that aren't there.
There is one test in hello_world_test.rb. GETTING_STARTED.md makes repeated references to three tests, two of which are skipped.

## Expected Behavior

Expected behavior is that the tests in hello_world_test.rb support the steps listed in GETTING_STARTED.md

## Current Behavior

Currently, GETTING_STARTED.md references code that is not there, and steps that are not possible to complete.

## Possible Solution

__Option 1__
Add two tests to hello_world_test.rb if you want to demonstrate how to un-skip tests, and run multiple tests.

This is probably the best option given the number of references to three tests in GETTING_STARTED.md, and the fact that the learner will be expected to know how to do this in future exercise.

__Option 2__
Change GETTING_STARTED.md so all references to three tests are removed, or rewritten to reference 1 test:

+ line 52 currently reads `3 runs, 0 assertions, 0 failures, 1 errors, 2 skips`. It could be changed to `1 runs, 0 assertions, 0 failures, 1 errors, 0 skips`

+ remove lines 77 - 84

+ remove lines 132 - 144

and then add instructions for how to un-skip a test in the next exercise?

chrisman:
+ Edit GETTING_STARTED.md to remove references to multiple tests since
there is only one test in the suite.

+ Add RUNNING_TESTS.md, which includes instructions for running multiple
tests.

closes #681

### DIFF
--- a/exercises/hamming/RUNNING_TESTS.md
+++ b/exercises/hamming/RUNNING_TESTS.md
@@ -1,0 +1,56 @@
+# Running Tests
+
+In order to complete this exercise, and all of the subsequent exercises, you
+will need to pass multiple tests.
+
+## Understanding Test Results
+
+After you have created and saved `hamming.rb`, run the test suite. You should
+see output like the following:
+
+    # Running:
+
+    SSSSSSESSSSSSSSS
+
+    Finished in 0.002593s, 6170.4588 runs/s, 0.0000 assertions/s.
+
+      1) Error:
+    HammingTest#test_identical_strands:
+    NameError: uninitialized constant HammingTest::Hamming
+    Did you mean?  HammingTest
+        ../hamming/hamming_test.rb:9:in `test_identical_strands'
+
+    16 runs, 0 assertions, 0 failures, 1 errors, 15 skips
+
+    You have skipped tests. Run with --verbose for details.
+
+
+The letters `SSSSSSESSSSSSSSS` show that there are sixteen tests altogether,
+that one of them has an error (`E`), and that the rest of them are skipped (all
+the `S` letters).
+
+The tests are run in randomized order, which will cause the letters to display
+in random order as well.
+
+The goal is to have all passing tests, which will show in two places:
+
+1. `SSSSSSESSSSSSSSS` will become a series dots: `................`, 
+
+2. The line at the bottom will read `16 runs, 0 assertions, 0 failures, 0
+   errors, 0 skips`.
+
+## Passing Tests
+
+Write enough code to change the Error to Failure, and finally to Passing.
+
+Open `hamming_test.rb`, and find the word "skip". All but the first test start
+with "skip", which tells Minitest to ignore the test. This is so that you don't
+have to deal with all the failures at once.
+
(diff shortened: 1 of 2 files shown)


Insti:
This looks great @chrisman 👍 thanks.

I changed the subject because it's nice to be able to  know what the PR is about without having to read a separate issue. Moved the closes reference to the body of the PR so that the referenced issue gets turned into a link.

exercises/hamming/RUNNING_TESTS.md
```diff
@@ -0,0 +1,56 @@
...
+    16 runs, 0 assertions, 0 failures, 1 errors, 15 skips
+
+    You have skipped tests. Run with --verbose for details.
+
+
+The letters `SSSSSSESSSSSSSSS` show that there are sixteen tests altogether,
+that one of them has an error (`E`), and that the rest of them are skipped (all
+the `S` letters).
+
+The tests are run in randomized order, which will cause the letters to display
```
kotp: `run` should be `ran`

Insti: I think 'run' is correct here. :/

kotp: Then perhaps the `are` should be removed?  `The tests run in randomized order` it is true.  "They are ran in randomized order" also.  Depends on the tense.

Insti: I don't have time for an English grammar discussion right now (I'd need to do more research), but both `The tests run in randomized order` and `The tests are run in randomized order` sound correct, and I prefer the second version.

`They are ran in randomized order` sounds really wrong.
`They ran in randomized order` would be ok, but means something different.

This may all just be due to regional differences.

How would you treat: `they will be run in randomized order`?

chrisman: I prefer `The tests are run in randomized order`. It sounds most correct.

It is also vetted. This is the exact phrasing, copied and pasted, from the original GETTING_STARTED.md. So FWIW, it has been reviewed and accepted once already.

kotp: `They are ran in randomized order` is correct but does sound wrong, I believe it is the least used.  `They are run in randomized order` was missing, and I believe is the most correct, describing present and future use.

chrisman: 👍 so it sounds as though we are all in agreement that `The tests are run in randomized order` does not need to be changed?

kotp: `The tests run in randomized order` is probably preferred.  The "most correct" was in reference to the list given by @insti.

It could still be changed, but of course "need" is much too strong a word here.

kotp:
This is awesome work.  The consideration of the one area that may be changed is something that can easily and quickly be done, after this is in.

Good discussion on a tricky trivial English language nuance.

PR #684, opened 2017-06-30
dominoes: Update generator and test cases

Insti:
Fixed the generator which broke when the expected value key was changed in the canonical data.

Regenerated test cases based on updated canonical data.

No change to the content of the tests so no BookKeeping::VERSION increment is required.


kotp:
Thanks Insti

PR #691, opened 2017-07-17
Fix phone-number test method name.

dkinzer:
The test method names are referencing area and exchange codes that start
with 2_9, but the actual inputs start with 2_2 and 4_5 respectively.

This change updates the name to reflect the actual inputs.

REF #598

### DIFF
--- a/exercises/phone-number/phone_number_test.rb
+++ b/exercises/phone-number/phone_number_test.rb
@@ -53,12 +53,12 @@ class PhoneNumberTest < Minitest::Test
     assert_nil PhoneNumber.clean("123-@:!-7890")
   end
 
-  def test_invalid_if_area_code_does_not_start_with_2_9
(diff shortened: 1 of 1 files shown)


Insti:
Thanks for your PR @dkinzer.

The test is not checking for numbers that start with 2 and 9 but rather numbers that are not in the range 2 to 9 (so basically 0 or 1) it may be that there is a better description for this test.

In which case the place to change that would be in the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json) from which these tests are generated.

Or it may be the responsibility of the [test generator](
https://github.com/exercism/ruby/blob/master/exercises/phone-number/.meta/generator/phone_number_case.rb) to ensure that the test name is correct. Currently it replaces the `-` with an `_` which is not helpful in this case.

Insti:
Thanks @dkinzer, you definitely did the right thing creating this PR :heart:

PR #697, opened 2017-07-28
refinements: removed from individual cases

### LINKED ISSUE #645: generator: refinements and lexical scope.
context: https://github.com/exercism/xruby/pull/583#pullrequestreview-37758023

Make the underscoring helper methods available to <problem>_cases without them having to remember to include 'using Generator::Underscore'

(This may involve not using refinements.)

ajwann:
fixes #645

### DIFF
--- a/lib/generator/exercise_case.rb
+++ b/lib/generator/exercise_case.rb
@@ -27,5 +27,20 @@ module Generator
     def respond_to?(sym, include_private = false)
       canonical.respond_to?(sym) || super
     end
+
+    protected
+
+    def literal(number)
+      number.underscore
+    end
+
+    def underscore(string)
+      string.underscore
+    end
+
+    def camel_case(string)
+      string.camel_case
+    end
+
   end
 end
--- a/exercises/collatz-conjecture/.meta/generator/collatz_conjecture_case.rb
+++ b/exercises/collatz-conjecture/.meta/generator/collatz_conjecture_case.rb
@@ -1,7 +1,6 @@
 require 'generator/exercise_case'
 
 class CollatzConjectureCase < Generator::ExerciseCase
-  using Generator::Underscore
 
   def workload
     case expected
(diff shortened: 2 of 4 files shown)


lib/generator/exercise_case.rb
```diff
@@ -27,5 +27,16 @@ def method_missing(sym, *args, &block)
     def respond_to?(sym, include_private = false)
       canonical.respond_to?(sym) || super
     end
+
+    protected
+
+    def underscore(target)
+      target.underscore
```
ajwann: I think it's best if we still use the refinement within ```ExerciseCase```. If we don't, we will have to define two separate ```underscore``` methods, one for ```Fixnum``` and one for ```String```, or define a single ```underscore``` method that switches on the type of the argument, but I really dislike that option.

Insti: I agree that switching on type is undesirable.

Insti:
Getting rid of the refinements in the <problem>Case classes is good 👍 and I think what you have here is mergable.

I'm not sure continuing to use the function with refinements the way we are here is a good idea. Underscoring a string and underscoring an Integer are two fundamentally different operations performed for different reasons, yet they are called using the same function.

Perhaps two different methods is the right way to go, but naming them based on `why` rather than `what`.

@kotp had some thoughts about this (that I can't find to link to right now.)

ajwann:
I agree that string underscore and integer underscore are very different operations, and I'm in favor of creating the ```literal(number)``` and ```underscore(string)``` methods.

Insti:
>  I'm in favor of creating the `literal(number)` and `underscore(string)` methods.

Is this something you want to do as part of this PR?

ajwann:
@Insti sure, no problem! I can make the change right now actually.

ajwann:
@Insti I think this is  good to, but let me know if you want any more changes.

Insti:
Looks good, thanks @ajwann

PR #698, opened 2017-07-28
Check for VERSION on BookKeeping module in binary_search_tree

### LINKED ISSUE #696: Binary search tree exercise checks for version on Bst class instead of Bookkeeping module
The [binary search tree exercise test](https://github.com/exercism/ruby/blob/5fc6cd30eabc9d90a8cef98fa95c29ba43b0b1dc/exercises/binary-search-tree/binary_search_tree_test.rb#L112) checks for `Bst::VERSION` instead of `BookKeeping::VERSION ` like the other exercises.

Not the biggest deal, but it's probably worth fixing if/when #513 is addressed.

Ryan1729:
Fixes #696

### DIFF
--- a/exercises/binary-search-tree/.meta/solutions/binary_search_tree.rb
+++ b/exercises/binary-search-tree/.meta/solutions/binary_search_tree.rb
@@ -1,6 +1,4 @@
 class Bst
-  VERSION = 1
-
   attr_reader :data, :left, :right
   def initialize(data)
     @data = data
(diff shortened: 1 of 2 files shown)


Insti:
Thanks @Ryan1729 !

This has failed the continuous integration test because the example solution also needs to be updated: https://github.com/exercism/ruby/blob/master/exercises/binary-search-tree/.meta/solutions/binary_search_tree.rb#L2

If you can also update that, then everything should be fine.

kotp:
Thank you @Ryan1729

PR #704, opened 2017-08-19
config.json: Add blurb, fix Gitter name

Insti:
The `blurb` is required by Nextercism to give a short overview of the language.

This text is taken from the front page of https://www.ruby-lang.org/en/


config.json
```diff
@@ -1,6 +1,7 @@
 {
   "language": "Ruby",
   "active": true,
+  "blurb" : "Ruby is a dynamic, open source programming language with a focus on simplicity and productivity. It has an elegant syntax that is natural to read and easy to write.",
   "gitter": "xruby",
```
kotp: Is this still `xruby` for the gitter key?

Insti: is the gitter key still used?

PR #709, opened 2017-08-27
hamming: Regenerate tests.

Insti:
BookKeeping version NOT updated, due to the canonical data changes not
changing the meaning of any of the tests.
They were just a minor re-ordering.


Insti:
These changes were made after #708 but are not dependent on it.

PR #714, opened 2017-08-27
RNA Transcription: Correct abbreviations in user-facing descriptions

kotp:
re: problem-specifications#838


Insti:
Those 'correctly_handles' are redundant :(

kotp:
That is definitely the case, about the redundancy of information in the titles.

PR #721, opened 2017-09-09
Autogenerate README based on description.md

rudresh4:
## Why?
According to the docs, README are auto-generated but that is not the case: https://github.com/exercism/ruby#readmes

Manually copy-pasting README from problem-specifications can lead to errors. This creates consistency across all READMEs. 

## What?
README is generated by combining:
1) https://github.com/exercism/problem-specifications/blob/master/exercises/accumulate/description.md
2) https://github.com/exercism/ruby/blob/master/docs/EXERCISE_README_INSERT.md

## How Has This Been Tested?
rake test
./bin/generate

## Types of changes
- [x] Add build_readme to generator
- [x] Add unit tests for generator
- [x] Run generator across all tests to create uniform READMEs

## Checklist:

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -3,7 +3,7 @@
 require_relative '../lib/helper'
 require 'generator'
 
-paths = Generator::Paths.new(track: EXERCISM_RUBY_ROOT, metadata: METADATA_REPOSITORY_PATH)
+paths = Generator::Paths.new(track: EXERCISM_RUBY_ROOT, docs: EXERCISM_RUBY_DOCS, metadata: METADATA_REPOSITORY_PATH)
 
 generators = Generator::CommandLine.new(paths).parse(ARGV)
 exit 1 unless generators
--- a/exercises/acronym/README.md
+++ b/exercises/acronym/README.md
@@ -1,5 +1,3 @@
-# Acronym
-
 Convert a phrase to its acronym.
 
 Techies love their TLA (Three Letter Acronyms)!
--- a/exercises/acronym/README.md
+++ b/exercises/acronym/README.md
@@ -32,10 +32,5 @@ To include color from the command line:
 
     ruby -r minitest/pride hello_world_test.rb
 
-
-## Source
-
-Julien Vanier [https://github.com/monkbroc](https://github.com/monkbroc)
-
 ## Submitting Incomplete Solutions
 It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/all-your-base/README.md
+++ b/exercises/all-your-base/README.md
@@ -1,4 +1,4 @@
-# All Your Base
+# All_your_base
 
 Convert a number, represented as a sequence of digits in one base, to any other base.
 
@@ -54,8 +54,6 @@ directory. For example, if the test suite is called
 To include color from the command line:
 
     ruby -r minitest/pride hello_world_test.rb
-
-
 
 ## Submitting Incomplete Solutions
 It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/all-your-base/README.md
+++ b/exercises/all-your-base/README.md
@@ -1,5 +1,3 @@
-# All_your_base
-
 Convert a number, represented as a sequence of digits in one base, to any other base.
 
 Implement general base conversion. Given a number in base **a**,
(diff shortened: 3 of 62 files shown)


exercises/acronym/README.md
```diff
@@ -1,5 +1,3 @@
-# Acronym
```
Insti: Why have all the titles disappeared?

Insti:
A tool to generate the readmes is good 👍 


There was some talk of having configlet generate READMEs, I'm not sure how far that idea progressed. You might like to search the discussions in /meta /configlet /discussions
(there are a lot of repositories!)

Insti:
I'd like to see better separation of commits here to make this easier to review.
at least:
1 commit for the README generator.
1 commit for the regeneration of all the exercise readmes.
(plus whatever else is relevant, the more atomic commits the better basically.)

bin/generate
```diff
@@ -3,7 +3,7 @@
 require_relative '../lib/helper'
 require 'generator'
 
-paths = Generator::Paths.new(track: EXERCISM_RUBY_ROOT, metadata: METADATA_REPOSITORY_PATH)
+paths = Generator::Paths.new(track: EXERCISM_RUBY_ROOT, docs: EXERCISM_RUBY_DOCS, metadata: METADATA_REPOSITORY_PATH)
```
Insti: `docs:` is not an appropriate argument here, the docs path should be derived from the `track: EXERCISM_RUBY_ROOT`

exercises/all-your-base/README.md
```diff
@@ -1,4 +1,4 @@
-# All Your Base
+# All_your_base
```
Insti: The underscores should not be here.

exercises/change/README.md
```diff
@@ -0,0 +1,44 @@
+# Change
```
Insti: Why is this showing as completely new?

exercises/acronym/README.md
rudresh4: The titles disappeared because the original problem-specfications/description.md don't have it in them.

Accumulate README from problem-specfications:
https://github.com/exercism/problem-specifications/blob/master/exercises/accumulate/description.md

Ruby's Accumulate README:
https://github.com/exercism/ruby/blob/master/exercises/accumulate/README.md

I was thinking about adding the titles into the description.md of problem-specifications for every problem so that when generator gets run the titles don't disappear.

Insti: Please have a read through this thread an the associated issues/PRs:
https://github.com/exercism/problem-specifications/issues/767

rudresh4:
Okay, turns out that Configlet actually does the same thing that I was trying to do but it expects the user to run it manually. IE, READMEs are properly generated using:

./configlet generate . --only=sublist

I think a better approach is to add this command to run "confliget generate" command with Travis CI and automatically have READMEs pushed back to repository?

Places to change: https://github.com/exercism/ruby/blob/master/.travis.yml
Upload files after build is successful: https://gist.github.com/willprice/e07efd73fb7f13f917ea

Thoughts?

exercises/acronym/README.md
Insti: And: https://github.com/Insti/trackler/blob/master/lib/trackler/implementation.rb#L87
For the code that used to (/still does) generate the readmes.

Insti:
> I think a better approach is to add this command to run "confliget generate" command with Travis CI and automatically have READMEs pushed back to repository?

This is premature I think. How about making the ruby track generate script call out to configlet to regenerate the readme?

rudresh4:
> How about making the ruby track generate script call out to configlet to regenerate the readme?

Yeah, that seems more reasonable

exercises/acronym/README.md
Insti: Things are a bit more complicated than normal right now as we're trying to make sure that the exercises are compatible with both the existing exercism.io site and the new https://v2.exercism.io/ (nextercism) site that have slightly different requirements.
It is nextercism that requires that the track have local README.md files.

Insti:
If possible do this as 2 pull requests.
1 that just adds the necessary generator changes.
and another that regenerates all the readmes. (You should be able to do this locally without having to have the generator changes merged.)

This will make things easier to review/test/merge.

Insti:
On a philosophical level, I personally don't think having configlet generate the readmes is the right thing to do, but that's how things seem to be, so we'll go with that for now.

rudresh4:
Calling configlet during generate: https://github.com/exercism/ruby/pull/722

PR #722, opened 2017-09-09
Invoke configlet to create README during generate process.

rudresh4:
## Why?
According to the docs, README are auto-generated but that is not the case: https://github.com/exercism/ruby#readmes

Manually copy-pasting README from problem-specifications can lead to errors. This creates consistency across all READMEs. 

## What?
README is generated by calling configlet command
`
./configlet generate . --only=acronym
`

## How Has This Been Tested?

1. Removal of README inside of directory
`
rm exercises/acronym/README.md 
`
2. Run bin/generate acronym -f
3. Verify README is generated for acronym

## Types of changes
- [x] Add configlet invocation to generator

## Checklist:

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -3,18 +3,20 @@
 require_relative '../lib/helper'
 require 'generator'
 
-paths = Generator::Paths.new(track: EXERCISM_RUBY_ROOT, metadata: METADATA_REPOSITORY_PATH)
-
-generators = Generator::CommandLine.new(paths).parse(ARGV)
-exit 1 unless generators
-
-generators.map(&:call)
-
 configletPath = "#{EXERCISM_RUBY_ROOT}/bin/configlet"
 unless File.file?(configletPath)
   raise Exception.new("Unable to find configlet, fetch using bin/fetch_configlet")
 end
 
+paths = Generator::Paths.new(track: EXERCISM_RUBY_ROOT, metadata: METADATA_REPOSITORY_PATH)
+
+generators = Generator::CommandLine.new(paths).parse(ARGV)
+exit 1 unless generators
+
 generators.each do |generator|
+  # Generate tests
+  generator.call
+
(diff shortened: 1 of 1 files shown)


Insti:
> According to the docs, README are auto-generated but that is not the case: https://github.com/exercism/ruby#readmes

They ARE generated on the existing `exercism.io` site, they will not be on the new `v2.exercism.io` site.

bin/generate
```diff
@@ -9,3 +9,12 @@ generators = Generator::CommandLine.new(paths).parse(ARGV)
 exit 1 unless generators
 
 generators.map(&:call)
+
+configletPath = "#{EXERCISM_RUBY_ROOT}/bin/configlet"
+unless File.file?(configletPath)
+  raise Exception.new("Unable to find configlet, fetch using bin/fetch_configlet")
+end
```
Insti: Should the configlet availability check happen earlier in the process?

bin/generate
```diff
@@ -9,3 +9,12 @@ generators = Generator::CommandLine.new(paths).parse(ARGV)
...
+
+generators.each do |generator|
+  system("#{configletPath} generate #{EXERCISM_RUBY_ROOT} --only=#{generator.exercise.slug}")
+end
```
Insti: I'm not entirely sure if this is the right way to be approaching this problem, I'll have to think about it for a while.

Is README generation related to test suite generation?
Will we always want to do both at the same time?
Should the code for doing this live inside the `Generator` with the other 3 things that it does already?

Insti: Do we want to be able to ONLY regenerate the readme for a problem?
Does this need to be controlled by command line arguments to bin/generate?

Insti:
Configlet readme generation documentation: https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/readmes.md

PR #727, opened 2017-09-25
Fix template fails with configlet v3.5.1

dkinzer:
The latest version of configlet throws an error when the template
includes `{{- -}}`

```
-> 1 error occurred:

* template: readme:3: illegal number syntax: "-"
``

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -1,13 +1,13 @@
 # {{ .Spec.Name }}
 
-{{ .Spec.Description -}}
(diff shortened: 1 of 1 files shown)


kytrinyx:
Just a quick note here to say that we're troubleshooting the issue over in exercism/configlet#82. I'm unable to reproduce the problem on my machine, so now we're narrowing down variables.

Please hold off on merging this until we're able to determine what is causing the issue.

dkinzer:
This is not an issue as long as we use a version of configlet compiled with go1.9 (or higher).

PR #729, opened 2017-09-27
guide user to create proper class and method via class instantiation in test file

thefonso:
## Why?

It is more natural to read the test with an instantiated class name. This helps guide the student into class-with-this-name contains method-with-this-name. Otherwise they will become lost.
## What?

simple really. for each
    assert_equal 'I', 1.to_roman
changed to
    assert_equal 'I', Numeric.new(1).to_roman
## How Has This Been Tested?

I ran this modification against my own solution code....https://github.com/thefonso/exercism-solutions-ruby/blob/master/roman-numerals/roman_numerals.rb
## Types of changes

- [x] Enhancement (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/exercises/roman-numerals/roman_numerals_test.rb
+++ b/exercises/roman-numerals/roman_numerals_test.rb
@@ -9,70 +9,87 @@ class RomanNumeralsTest < Minitest::Test
   end
 
   def test_2
+    skip
     assert_equal 'II', Numeric.new(2).to_roman
   end
 
   def test_3
+    skip
     assert_equal 'III', Numeric.new(3).to_roman
   end
 
   def test_4
+    skip
     assert_equal 'IV', Numeric.new(4).to_roman
   end
 
   def test_5
+    skip
     assert_equal 'V', Numeric.new(5).to_roman
   end
 
   def test_6
+    skip
     assert_equal 'VI', Numeric.new(6).to_roman
   end
 
   def test_9
+    skip
     assert_equal 'IX', Numeric.new(9).to_roman
   end
 
   def test_27
+    skip
     assert_equal 'XXVII', Numeric.new(27).to_roman
   end
 
   def test_48
+    skip
     assert_equal 'XLVIII', Numeric.new(48).to_roman
   end
 
   def test_59
+    skip
     assert_equal 'LIX', Numeric.new(59).to_roman
   end
 
   def test_93
+    skip
     assert_equal 'XCIII', Numeric.new(93).to_roman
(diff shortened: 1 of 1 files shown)


Insti:
Thanks for the PR @thefonso. I'll have a closer look at this when I get a chance.

Insti:
> It is more natural to read the test with an instantiated class name. This helps guide the student into class-with-this-name contains method-with-this-name. Otherwise they will become lost.

Was this something you encountered? Part of this exercise is about learning that you can extend the integer class by adding new methods to it.

thefonso:
Yes it has been something I encountered. I see where you are thinking regarding the integer class. However, perhaps it can be made clearer by explicitly saying Integer(1). (Or perhaps explicitly state this as a hint in a comment in the test file) Otherwise the student can become lost. Cheers.

Insti:
Updating the Readme with some hints or instructions about this is a good idea.

Another option is to move the exercise further down the list so it is encountered later.
Currently it is 9th(?) which is still quite soon to be introducing class extension.

What do you think @kotp?

Do you feel like you could contribute an update to the readme @thefonso?

Insti:
I had a look through http://exercism.io/stats/ruby to see if there was a big drop off between sieve and roman-numerals that might indicate that people are having trouble and giving up.

But my only conclusion was 'it's hard to tell'

kotp:
This is also related in some ways to `change` exercise as well.  I think that the tests indicate that we can call the method on a Numeric, but I am not sure that I would want to write a test to reference Numeric, because, what if I reasonably would want to exclude Float?  Having the tests written where `1.to_roman` definitely leaves the discussion open... And this has more value than too strict of tests.  The tests should strictly enforce a result, but not the "how", if that makes sense.

The current test suggests an Integer, which is a good suggestion.  It does so without spelling it out, and for our purposes, I think that is a good thing.  Using Numeric says "We don't care if you use a Float or Complex" which may not be the case or Integer or whatever kinds of "numeric" types there are included in that class structure.

That said, yes, anything that clarifies the "what" without directing the "how" is a good thing, and this likely belongs in the Readme, if people are having difficulties understanding.

Insti:
I don't think changing the tests is the right way to address this problem.
Extending/Refining the `Integer` class is an appropriate way to solve this problem in Ruby.

The thing to do would be to update the README to explain this is possible and give hints as to how to implement it.

Insti:
Thanks for bringing up this issue @thefonso 👍

PR #732, opened 2017-10-05
generator: Remove underscore in test method name when canonical data has whitespace in description

vinmaster:
## Why?
Remove unnecessary underscore in test method name to make it more readable

Issue: https://github.com/exercism/ruby/issues/668

## What?
Modify Underscore module to clean up leading/trailing whitespace

## How Has This Been Tested?
Updated `exercise_case_test.rb` to account for this case

## Types of changes

- [x] Enhancement (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/lib/generator/underscore.rb
+++ b/lib/generator/underscore.rb
@@ -2,7 +2,7 @@ module Generator
   module Underscore
     refine String do
       def underscore
-        strip.downcase.gsub(/[- ]/, '_').gsub(/[^\w?]/, '')
+        downcase.gsub(/[- ]/, '_').gsub(/[^\w?]/, '')
       end
 
       def camel_case
--- a/lib/generator/underscore.rb
+++ b/lib/generator/underscore.rb
@@ -2,7 +2,7 @@ module Generator
   module Underscore
     refine String do
       def underscore
-        downcase.gsub(/[- ]/, '_').gsub(/[^\w?]/, '')
+        strip.downcase.gsub(/[- ]/, '_').gsub(/[^\w?]/, '')
       end
 
       def camel_case
--- a/test/generator/exercise_case_test.rb
+++ b/test/generator/exercise_case_test.rb
@@ -7,6 +7,16 @@ module Generator
       assert_equal 'test_foo', subject.name
     end
 
+    def test_name_with_trailing_whitespace
+      subject = ExerciseCase.new(canonical: OpenStruct.new(description: 'foo '))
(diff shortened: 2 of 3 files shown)


test/generator/exercise_case_test.rb
```diff
@@ -7,6 +7,16 @@ def test_name
...
 
+    def test_name_with_trailing_whitespace
+      subject = ExerciseCase.new(canonical: OpenStruct.new(description: 'foo '))
+      assert_equal 'test_foo', subject.name
+    end
+
+    def test_name_with_leading_whitespace
+      subject = ExerciseCase.new(canonical: OpenStruct.new(description: ' foo'))
+      assert_equal 'test_foo', subject.name
+    end
```
Insti: These tests are good 👍

lib/generator/underscore.rb
```diff
@@ -2,7 +2,7 @@ module Generator
   module Underscore
     refine String do
       def underscore
-        downcase.gsub(/[- ]/, '_').gsub(/[^\w?]/, '')
+        strip.downcase.gsub(/[- ]/, '_').gsub(/[^\w?]/, '')
```
Insti: This solution works, but I'm not sure this is the right place to be `strip`ping the string.

I would expect `' something '.underscore` to be `'_something_'` (this may be an unreasonable expectation.)

Is there somewhere further up the call stack where the `strip` could happen?

Insti:
Thanks @vinmaster :heart:
This looks good and solves the issue, but I wonder if the change might be better in a different place. (The answer might be no.) Can you have a look and see where else you might be able to make the change?

lib/generator/underscore.rb
vinmaster: Thanks for reviewing my PR

I was thinking about using `strip` in `exercise_case.rb` instead, but I see that `underscore` method also `downcase` the string. I thought `strip` as part of cleaning up before converting to underscore like how `downcase` was

Insti: I think it would be better in `exercise_case.rb`

Your reasoning about where to put it is fine. Do you have any objection to moving it to  the `ExerciseCase` class?

I would also prefer that the `downcase` wasn't there, but you didn't put that in and it's not related to the current task, so I wasn't suggesting you take it out at this point.

vinmaster:
Changes made. Please have a look 😄

Insti:
Thanks @vinmaster ❤️

PR #737, opened 2017-10-15
generator: replace --freeze with --update

### LINKED ISSUE #631: generator: replace --freeze with --update
(@Insti please fill in as you wish)

There's a long history to how we ended up with `bin/generate` defaulting to updating the xruby version of an exercise, with a `--freeze` option to prevent that from happening, but what matters is that the default usage doesn't match our current needs.

The issue came to the forefront when we added the `--all` option. 

What to do: replace the `--freeze` option with an `--update` option.

* not updating becomes the default
* adding new behaviors becomes much simpler

Make sure to clean up the `--all` logic to be sure it is still correct.

carolkng:
## Why?
fixes #631 

## What?
Replaces the --freeze and -f options with --update and -u.
Updates default --all behaviour to not update version.
Alternative implementation could involve *adding* an update option, instead of replacing the freeze option entirely.

## Testing
`rake test:dev`

## Types of changes

- [x] Breaking change (fix or enhancement that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,8 +108,8 @@ From within the ruby directory, run the following command:
 
     bin/generate --update <slug> 
 
-To regenerate tests without the updated canonical data, run the command
-without the `--update` option.
+Leaving out the --update option will cause the BookKeeping version number to remain the same.
+This can be useful when testing generators.
 
 #### Changing a Generated Exercise
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,10 @@ tree -L 1 ~/code/exercism
 
 From within the ruby directory, run the following command:
 
-    bin/generate <slug>
+    bin/generate --update <slug> 
+
+To regenerate tests without the updated canonical data, run the command
+without the `--update` option.
 
 #### Changing a Generated Exercise
 
--- a/lib/generator/command_line.rb
+++ b/lib/generator/command_line.rb
@@ -29,11 +29,11 @@ module Generator
     end
 
     def generator_class
-      freeze? ? GenerateTests : UpdateVersionAndGenerateTests
+      update? ? UpdateVersionAndGenerateTests : GenerateTests
     end
 
-    def freeze?
-      @options[:freeze]
+    def update?
+      @options[:update]
     end
 
     def implementation(slug)
--- a/lib/generator/command_line.rb
+++ b/lib/generator/command_line.rb
@@ -33,7 +33,7 @@ module Generator
     end
 
     def freeze?
-      @options[:freeze] || @options[:all]
+      @options[:freeze]
     end
 
     def implementation(slug)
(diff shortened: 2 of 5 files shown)


Insti:
This is great! Thanks.

lib/generator/command_line/generator_optparser.rb
```diff
@@ -3,7 +3,7 @@
 module Generator
   class GeneratorOptparser
     DEFAULT_OPTIONS = {
-      freeze: false,
+      freeze: true,
```
Insti: Since `:freeze` is no longer an option, it would be good to be able to use `:update` as the options hash key.

lib/generator/command_line.rb
```diff
@@ -33,7 +33,7 @@ def generator_class
     end
 
     def freeze?
-      @options[:freeze] || @options[:all]
+      @options[:freeze]
```
Insti: This method could be renamed to `update?`

Insti:
We also need to remember to add instructions to use `--update` to the readme:

https://github.com/exercism/ruby#regenerating-a-test-suite

Insti:
Thanks @carolkng this looks really good, I would be happy to merge it as it is.

I have a few minor suggestions about removing references to freeze which you could implement if you are willing and have the time.

carolkng:
re: adding to the documentation

How should the documentation be extended, in this case? The current documentation suggests running `/bin/generate` is all that is necessary, and doesn't mention any of the CLI options. Would the change be to mention all of them (verbose, update, etc.)?

Insti:
Re: Latest changes - These look great! many thanks :heart:

Re: Adding to the documentation

I this case I would suggest just adding the `--update` flag to the instruction in the README

If you wanted to add a new section that described the generator tool and its usage/options you are welcome to open another PR for that.

Insti:
> Does a user usually expect to have the xruby version of exercise updated whenever they regenerate test suites?

Sometimes you want to, sometimes you don't want to.
Usually when you're testing, or making a new test suite you don't want the version incrementing every time which is why the default not to update is good.
When you are updating the tests "for real" because the canonical-data has changed for example, you do want the version to update, so using the `--update` flag is appropriate there. The README seems to be documenting the case where you do want to update the version.

README.md
```diff
@@ -106,7 +106,10 @@ tree -L 1 ~/code/exercism
 
 From within the ruby directory, run the following command:
 
-    bin/generate <slug>
+    bin/generate --update <slug> 
+
+To regenerate tests without the updated canonical data, run the command
+without the `--update` option.
```
Insti: This sentence is misleading.

The `--update` flag causes the BookKeeping version number to increment.

How about something like: 

"Leaving out the `--update` option will cause the BookKeeping version number to remain the same.
This can be useful when testing generators."

Insti:
Great.
I've tested this works for me locally, so I think we can merge.

Insti:
Thanks for all your work on this @carolkng :heart:

PR #740, opened 2017-10-18
two-fer: Implement exercise

architapatelis:
## Why?

To add another exercise to Ruby track.

## What?

Add exercise two-fer. This is work-in-progress PR, as suggested by documentation.

## Types of changes

- [x ] Enhancement (non-breaking change which adds functionality)

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.


Insti:
Thanks @architapatelis let us know if you have any questions. Either here, or in the gitter chatroom: https://gitter.im/exercism/ruby

exercises/two-fer/.meta/generator/two_fer_case.rb
```diff
@@ -0,0 +1,9 @@
+require 'generator/exercise_case'
+
+class TwoFerCase < Generator::ExerciseCase
+
+  def workload
+    assert_equal { "TwoFer.two_fer" }
```
Insti: This generator does not generate the test cases.

See: https://github.com/exercism/ruby#generated-test-suites

exercises/two-fer/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+    gem install minitest
+
+If you would like color output, you can `require 'minitest/pride'` in
+the test file, or note the alternative instruction, below, for running
+the test file.
+
+In order to run the test, you can run the test file from the exercise
+directory. For example, if the test suite is called
+`hello_world_test.rb`, you can run the following command:
```
Insti: We know what the test suite is called, so we should not use hello_world as an example here.

Insti: See: https://github.com/exercism/ruby/pull/748/files#diff-fbef0ff69fea6a3ef4b865b66b037dd5R55

config.json
```diff
@@ -16,6 +16,16 @@
       ]
     },
     {
+      "uuid": "b02a4214-0f1a-9480-2132-2ea3d5073dbfbec0aa4",
+      "slug": "two-fer",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+
```
Insti: Adding some topics would be good, see #748 for an example.

architapatelis:
@Insti thanks for the feedback, I am working on creating the generator. I was looking at the `pig-latin` example, that exercise passes a `phrase` parameter to the `translate` method - https://github.com/exercism/ruby/blob/master/exercises/pig-latin/.meta/generator/pig_latin_case.rb

the `two-fer` exercise passes a `name` parameter to `two_fer` method, so for the `workload` method should I be writing something like this:

```ruby
class TwoFerCase < Generator::ExerciseCase
  def workload
    assert_equal { "TwoFer.two_fer(#{input.inspect})" }
  end
end
```

I also looked at the `nth-prime` where a number `n` is passed to the `nth` method -  https://github.com/exercism/ruby/blob/master/exercises/nth-prime/.meta/generator/nth_prime_case.rb

that example, uses `"Prime.nth(#{input})"` what's the difference between `input.inspect` and `input`?

Insti:
>  I am working on creating the generator.

Great.

`inspect` is a ruby method. Try it out in irb.

```
$ irb
2.4.1 :001 > a = 'hello'
 => "hello" 
2.4.1 :002 > puts a
hello
 => nil 
2.4.1 :003 > puts a.inspect
"hello"
 => nil 
```

It's useful when your output might contain quotes, as it will escape them.
For two-fer none of the canonical-data inputs currently contain quotes, so use whichever method you are most comfortable with.

Insti:
Numbers need to not have quotes around them which is why nth-prime doesn't use `inspect`.

architapatelis:
@Insti thanks for the explanation. I am new to ruby, and coding in general so sorry for the trouble. I think I have correct code for `workload` now.

cemeng:
:wave: @architapatelis - I had a similar PR for this exercise (https://github.com/exercism/ruby/pull/748) - I will try to help you with this PR (by giving suggestions I guess) seeing that you are new to Ruby :) - keep up the good work

architapatelis:
@cemeng I think I have made all the corrections requested. If I need to make more modifications I will reach out to you. thanks! for lending a helping hand :)

exercises/two-fer/two_fer_test.rb
```diff
@@ -0,0 +1,41 @@
+require 'minitest/autorun'
+require_relative 'two_fer'
+
+class TwoFerTest < Minitest::Test
+  def  test_empty
```
cemeng: there is an extra space here :)

exercises/two-fer/two_fer_test.rb
```diff
@@ -0,0 +1,41 @@
...
+    #skip
+    assert_equal "One for you, one for me.", TwoFer.two_fer
+  end
+
+  def test_alice
+    skip
```
cemeng: I think none of these tests should be skipped? as they are good tests. what do you think @Insti?

exercises/two-fer/.meta/solutions/two_fer.rb
```diff
@@ -0,0 +1,13 @@
...
+end
+
+class TwoFer
+  def self.two_fer(name=nil)
+    if name == nil
+      return "One for you, one for me."
+    else
+      return "One for #{name}, one for me."
+    end
+  end
```
cemeng: I wonder if we can write more idiomatic ruby for this. For example you don't need ```return```. 
Ruby method will return the last evaluated value by default. So the code above is the same as:
```
def self.two_for(name = nil)
   if name == nil
      "One for you, one for me"
  else 
     "One for #{name}, one for me"
  end 
end
```

But! Ruby is so awesome and expressive - you don't even need to check for nil explicitly,  nil, empty string, false is ```falsy``` in ruby, conversely any other values are ```truthy```.  
```
if !name 
    "One for you, one for me"
else
    "One for #{name}, one for me"
end
```

```if !condition``` statement usually is written as ```unless condition``` in ruby too 

Sorry is a long winded comment - this is just style comment of course - your code is correct functionally. 

I am not sure if exercism worry about idiomatic ruby - what do you think @Insti ?

File: exercises/two-fer/two_fer_test.rb (hunk 5 above)
Comment (Insti) [reply exercises/two-fer/two_fer_test.rb]: The `skip`s should be there, they are part of the standard test cases and are added by the test generator. The skip just means that that test won't be run for now, but should be enabled later. This is so the student is only presented with one failing test to make pass at a time.


See: http://exercism.io/languages/ruby/tests
And: https://github.com/exercism/ruby/issues/753

File: exercises/two-fer/.meta/solutions/two_fer.rb (hunk 6 above)
Comment (Insti) [reply exercises/two-fer/.meta/solutions/two_fer.rb]: I'm currently not too concerned about having idiomatic Ruby in the example solution, as long as it passes the tests.

Your suggestions above are certainly valid.

If I were solving this I would eliminate the conditional completely and write it as:

```ruby
def self.two_for(name = 'you')
  "One for #{name}, one for me"
end
```

The best place to discuss this sort of thing is on the Exercism site itself. 😄

Insti:
This is looking good @architapatelis I will give it another full review when I get a chance.

config.json
```diff
@@ -16,6 +16,17 @@
...
+      "uuid": "b02a4214-0f1a-9480-2132-2ea3d5073dbfbec0aa4",
+      "slug": "two-fer",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Control-flow (conditionals)",
+        "Strings"
+      ]
+    },
```
Insti: New exercises should be added at the end of the list please.
Before the deprecated entry for 'octal'

exercises/two-fer/two_fer_test.rb
```diff
@@ -0,0 +1,41 @@
+require 'minitest/autorun'
+require_relative 'two_fer'
+
+class TwoFerTest < Minitest::Test
+  def test_empty
```
Insti: This test file should be the one generated by the test generator you have written.

You can create the file by running: `bin/generate two-fer` from your Exercism ruby directory.

Make sure you have the `problem-specifications` repository checked out.

See: https://github.com/exercism/ruby#generated-test-suites

Insti:
There should also be a `exercises/two-fer/.meta/.version` file as part of this pull request.

exercises/two-fer/two_fer_test.rb
```diff
@@ -0,0 +1,42 @@
+require 'minitest/autorun'
+require_relative 'two_fer'
+
+# Common test data version: 1.1.0 c080bdf
+class TwoFerTest < Minitest::Test
+  def test_no_name_given
+    # skip
+    assert_equal "One for you, one for me.", TwoFer.two_fer
```
Insti: When I run the generator for this problem it gives me this line: 

`assert_equal "One for you, one for me.", TwoFer.two_fer(nil)`

The generator should generate the test file without it requiring any further editing.

Insti:
All the other changes you've made have been great 👍 Things are heading in the right direction. :star:

architapatelis:
@Insti thanks for all the help and being so patient, I didn't realize that the generator was editing the test code. I believe the problem has been fixed, I ran the generator command, and didn't notice any further editing.

Insti:
Thanks for sticking with this @architapatelis :heart:
I hope you learned something useful along the way.

It will take a day or so for the exercise to appear on the main Exercisim.io site.

Thanks again.

cemeng:
Awesome work @architapatelis :1st_place_medal:

PR #745, opened 2017-10-18
pig-latin: regenerate tests

Insti:
Regenerate tests based on updated canonical data.

BookKeeping version incremented:

Additional tests for handling words containing 'y' added.


Insti:
Need to update the example solution to pass the new tests.

PR #746, opened 2017-10-18
two-bucket: regenerate tests

Insti:
Regenerate tests based on updated canonical data.

BookKeeping version incremented:

The expected solution for a test case changed.


Insti:
Need to update the example solution to pass the new tests.

exercises/two-bucket/.meta/solutions/two_bucket.rb
```diff
@@ -68,6 +72,22 @@ def start_from_second
...
+    else
+      fill_second_bucket
+    end
+  end
+
+  def other_bucket_matches_goal
+    if @start_bucket == 'one'
+      return @second_size == @goal
+    else
+      return @first_size == @goal
```
kotp: Redundant `return` used here.  The last statement of the method is the if statement, and therefore will be returned.

Insti:
The example solution here is certainly not an elegant solution, or I suspect particularly robust, but it passes all the existing tests.

PR #751, opened 2017-10-19
prime-factors: Prevent using built in Prime class

lugray:
## Why?
The exercise can be sidestepped entirely by using the built in Prime class, which defeats the point.

## What?
Ensure the built in prime class cannot be used by raising an error.

## How Has This Been Tested?
Adding `require "prime"` to a working solution causes it to fail with this change.

### DIFF
--- a/exercises/prime-factors/prime_factors_test.rb
+++ b/exercises/prime-factors/prime_factors_test.rb
@@ -1,5 +1,12 @@
 require 'minitest/autorun'
-require_relative 'prime_factors'
+Prime = nil
+class CheatError < StandardError; end
+begin
+  require_relative 'prime_factors'
+rescue TypeError => e
+  raise e unless e.message == "Prime is not a class"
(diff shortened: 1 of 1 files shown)


Insti:
Thanks for submitting this pull request :heart:

However I am not in favour of merging it, as I believe that using Ruby's built-in prime module is a valid solution to this problem.

There are two types of Exercism student, new programmers and experienced programmers.
Experienced programmers know how to implement a prime-factorisation algorithm and the discovery that Ruby has this built in is a benefit.
New programmers can also benefit from using the built-in module but are not prevented from implementing the full algorithm themselves.

The place to make this distinction and suggest alternate options is during code reviews of submitted solutions on the Exercism site.

lugray:
Fair enough!

Insti:
Thanks for taking the time to make the pull request. It was a good solution to the issue you identified. 👍

PR #758, opened 2017-10-24
raindrops: add test for 0 as % matches everything on 0

rever0f:
## Why?

So I would guess this exercise should return "0" on 0 input. Only 1 out of 12 responses I reviewed would return this result.

## What?
Added a test to verify a return of  "0" on 0 input.

## How Has This Been Tested?

I ran the test with an exercise that returned "0" and one that returned "PlingPlangPlong" and it errored as appropriate.

## Types of changes

- [ ] Enhancement (non-breaking change which adds functionality)
- [x] Breaking change (fix or enhancement that would cause existing functionality to change)

should only break test results.

## Checklist:

If this test performs as expected should I update book keeping to a new version?

### DIFF
--- a/exercises/raindrops/raindrops_test.rb
+++ b/exercises/raindrops/raindrops_test.rb
@@ -93,6 +93,11 @@ class RaindropsTest < Minitest::Test
     assert_equal "Plang", Raindrops.convert(3125)
   end
 
+  def test_the_sound_for_0_is_0
+    skip
+    assert_equal "0", Raindrops.convert(0)
+  end
+
   # Problems in exercism evolve over time, as we find better ways to ask
   # questions.
   # The version number refers to the version of the problem you solved,


Insti:
Thanks @rever0f.

The common test data for this problem comes from the https://github.com/exercism/problem-specifications repository,  could you open an issue there to discuss whether adding a test for zero is appropriate and what the expected value should be.

I believe zero has infinitely many factors so "PlingPlangPlong" would be the correct output.
But my math knowledge is a bit rusty. Why do you thing "0" is the desired result here?

Insti:
https://math.stackexchange.com/questions/776627/factors-of-zero

rever0f:
Sorry, thought I'd just looked this up recently and 0 had no factors, but must have been something else. Looking it up now as  far as I can tell the internet consensus is that 0 has infinite factors, which is kind of interesting. Not sure of a definitive source, but good enough for me.

Insti:
No problem. :)

Thanks for opening the Pull Request anyway. :heart:

Issue #761, opened 2017-10-27
meetup: re-implement exercise.


Insti:
[The README for meetup](https://github.com/exercism/ruby/blob/master/exercises/meetup/README.md) says that the input to the exercise should be a string describing when a meetup should occur, and the output should be a date.

> "First Monday of January 2017" -> 2017/1/2

However the tests for the Ruby track implementation expect an odd combination of initializer and method arguments.

`Meetup.new(8, 2013).day(:monday, :teenth)`

**Task:**

Re-write the tests and example solution so that the input is a `String` and the output is a `Date`

`assert_equal Date.new(2017, 1, 2), Meetup.new("First Monday of January 2017").to_date`

Note that the tests are generated from [common canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/meetup/canonical-data.json)

See also: 
* https://github.com/exercism/ruby/pull/759 (Pull Request for the test generator.)
* https://github.com/exercism/problem-specifications/issues/974 (Issue about test descriptions)

Insti:
I suspect this is still a problem. 
Please check and leave comment about what you find.

Insti:
#764 made good progress on this and probably just needs a few tweaks to complete the task.

PR #764, opened 2017-10-31
meetup: rewrite tests and example implementation to align with README

nwallace:


nwallace:
Here's an attempt to fix #761 . The README gives examples like "the Wednesteenth of January 2017", but the syntax I'm specifying in these tests is of the form "Teenth Wednesday of January, 2017". Is that right, or should it be "Wednesteenth"?

Insti:
Hi @nwallace thanks for taking this on :heart:

It should be "wednesteenth", those crazy words are one of the "features" of this this exercise.

There is currently a pull request open to clarify the description of this problem:
https://github.com/exercism/problem-specifications/pull/982
You probably should bare those in mind while working on this.

Insti:
This is looking good and is definitely on the right track 👍 

You will also need to update [the example solution](https://github.com/exercism/ruby/blob/master/exercises/meetup/.meta/solutions/meetup.rb) to pass the new test suite.

Insti:
The description changes have now been merged: https://github.com/exercism/problem-specifications/blob/master/exercises/meetup/description.md

PR #777, opened 2017-11-11
sublist: Implement exercise

njbbaer:
I've implemented the sublist exercise for Ruby. The tests and readme were generated from shared metadata, and I wrote a basic solution.

Previous attempts #427 #335 & #342 appear abandoned, so I decided to do it myself.

Please make comments and I will update if necessary. I'd like to see this through to be merged.

Edit: add link to [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/sublist/canonical-data.json)


njbbaer:
I added a basic entry for sublist to the bottom of `config.json` which appears determine its priority.

Are easier exercises placed earlier in the track? Where should sublist be positioned?

Insti:
New exercises are always placed last. 
Once they have been solved a few times they can be moved to a more appropriate place in the list.

Insti:
Firstly, this looks great 👍 
It looks like you've done everything that needs doing for this exercise.
I don't have time right now to look through it properly, I will do that later.

But I wonder if a more Rubyish interface with a specific List class and query methods would be better:
```ruby
a = List.new([1,2,3])
b = List.new([2,3])
a.sublist?(b) # false
a.superlist?(b) # true
a == b # false

```

config.json
```diff
@@ -912,6 +912,16 @@
...
       "difficulty": 3,
       "topics": [ "Mathematics" ]
+    },
+    {
+      "uuid": "1af13ab1-04ea-5380-0958-32c65f0dbe41932275d",
+      "slug": "sublist",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
```
Insti: Adding some topics from the [list of topics](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) would be good.

njbbaer:
I agree, the exercise should probably teach more, and I like those suggestions. I'll have to look into how I could make that work with the test generator, since the canonical data isn't structured that way.

config.json
njbbaer: Will do!

njbbaer:
The problem with your suggested interface @Insti is that the canonical data only gives affirmative examples. So for example, a test for sublist would also need to check that it is not a superlist or equal. This would mean testing multiple things per unit test, which is sorta bad practice and might confuse the learner. I could break them into separate tests, but I'm unsure how to use the generator to create multiple tests per data entry.

Insti:
It's true that the transformation from the `canonical-data.json` to the Ruby tests will be more complicated than usual, but it should still be possible.
We're also allowed to add our own tests, we don't have to only use those in the canonical data file.

I agree testing multiple things per test is undesirable.

> a test for sublist would also need to check that it is not a superlist or equal.

Why?

I suggest ignoring the canonical data and doing a TDD implementation of the new version from scratch and seeing what tests you need. Then comparing those tests to the tests given in the canonical data and seeing what the differences are.

njbbaer:
> > a test for sublist would also need to check that it is not a superlist or equal.
>
> Why?

Here is an excerpt from the canonical data:
```
    {
      "description": "empty list within non empty list",
      "property": "sublist",
      "listOne": [],
      "listTwo": [1, 2, 3],
      "expected": "sublist"
    },
    {
      "description": "non empty list contains empty list",
      "property": "sublist",
      "listOne": [1, 2, 3],
      "listTwo": [],
      "expected": "superlist"
    },
```
And sample tests for the new interface:
```
  def test_empty_list_within_non_empty_list
    assert [].sublist?([1, 2, 3])
  end

  def test_non_empty_list_contains_empty_list
    assert [1, 2, 3].superlist?([])
  end
```

Notice the sample data only gives affirmative examples. It shows you what is a sublist or superlist but doesn't show what isn't. Meaning a test for sublist or superlist that always returns true would always pass its tests.

The negative cases can be implicitly determined. If two lists are equal than they cannot also be unequal, or a sublist cannot also be a superlist. But that would require multiple function calls per test, or multiple tests per example. If you follow my thinking.

njbbaer:
@Insti Take a look at my updated implementation which uses your suggested interface and multiple assertions per generated test. Although this may not be good practice in general, I feel it is justified in this case, and clear enough for the learner to understand.

Insti:
Cool, thanks for this, 👍 sorry I've not looked at this earlier. I will review further when I get a chance.

njbbaer:
@Insti Please review this when you can. Like I said, I'd like to see this eventually merged.

Insti:
Sorry, thanks for the bump.

Insti:
I've not forgotten about this, I'm still trying to get to it, I've been busy.

njbbaer:
😞

PR #780, opened 2017-11-27
Explicitly disallow prime library

tenebrousedge:
Too many solutions avoid the point of this exercise by using the built-in prime-finding library.

### DIFF
--- a/exercises/sieve/README.md
+++ b/exercises/sieve/README.md
@@ -25,7 +25,9 @@ https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes
 
 Notice that this is a very specific algorithm, and the tests don't check
 that you've implemented the algorithm, only that you've come up with the
-correct list of primes.
+correct list of primes. Your solution must implement the algorithm
+described here, without using prime-finding libraries (if your language
+provides those).
 
 * * * *
 


Insti:
Thanks for submitting this pull request ❤️

However I am not in favour of merging it, as I believe that using Ruby's built-in prime module is a valid solution to this problem.

There are two types of Exercism student, new programmers and experienced programmers.
Experienced programmers know how to implement a prime-sieve algorithm and the discovery that Ruby has this built in is a benefit.
New programmers can also benefit from using the built-in module but are not prevented from implementing the full algorithm themselves.

The place to make this distinction and suggest alternate options is during code reviews of submitted solutions on the Exercism site.

Insti:
See also https://github.com/exercism/problem-specifications/issues/761 for some discussion about whether specific algorithm exercises are even a good idea. (They're probably not.)

tenebrousedge:
@Insti  Ah, so it seems that despite the problem specifically asking for that algorithm, solutions which do not use that algorithm, such as the `prime` library, are acceptable. Should then the problem description be revised to additionally describe other algorithms, such as prime division or the wheel sieve, or perhaps simply to insert some clarifying language with the opposite intent of the proposed changeset above?

Insti:
My opinion is that, it's OK that the problem exists in its current form.

Anyone who wants to can implement the specified algorithm.
Anyone who wants to use another method is free to do that too.
Anyone who wants to can skip the problem completely.

They can do this whether or not the README suggests it.

Reviews can point out if the correct algorithm was used or not.

The only case where it is helpful is if there is someone who wanted to implement the sieve algorithm but has not, and people tend to know when they submit whether they have or not.

In all cases it's up to the student to decide what they want to do and for the reviewers to help the student learn. We can't really "require" anybody to do anything.

I expect eventually the `sieve` exercise will disappear from Exercism and will be replaced by another less algorithm specific exercise, which is close to what you are suggesting.

PR #783, opened 2017-12-03
isogram: update generator to put input variable in failure message

bmkiefer:
I received feedback on my [PR](https://github.com/exercism/ruby/pull/782) saying that the failure message in the unit tests should leverage the same string variable. I changed the isogram generator to reflect that change. Not sure if this requires an update to the test suite version or not.


exercises/isogram/.meta/generator/isogram_case.rb
```diff
@@ -18,7 +18,7 @@ def failure_message
   end
 
   def reason
-    "'#{input}' #{is_or_not} an isogram"
+    "'\#{string}' #{is_or_not} an isogram"
```
kotp: I think "input" is reasonable here as a name, as perhaps the input may not be a string, and the statement would still be true, it is or not an isogram.  The generic is informative here, in my opinion.

exercises/isogram/.meta/solutions/isogram.rb
```diff
@@ -1,5 +1,5 @@
 module BookKeeping
-  VERSION = 4
+  VERSION = 5
```
kotp: I don't believe this changes the tests functionally, and would not break any solution that is in place, so the version will not need to be bumped.

exercises/isogram/isogram_test.rb
```diff
@@ -76,6 +76,6 @@ def test_duplicated_character_in_the_middle
 
   def test_bookkeeping
     skip
-    assert_equal 4, BookKeeping::VERSION
+    assert_equal 5, BookKeeping::VERSION
```
kotp: Version should be the same, non-breaking change for current solutions, no change in the tests themselves.

kotp:
Thank you again @bmkiefer!

PR #785, opened 2017-12-30
Order gems alphabetically in Gemfile

kytrinyx:
I don't necessarily think this is particularly important, especially since the Gemfile is quite short, but I have been ordering my Gemfiles alphabetically for a while, and like the consistency of knowing where to look for things.


kotp:
Than you @kytrinyx.

PR #793, opened 2018-01-16
clock: change "add" of negative minutes to "subtract" of positive minutes.

MattLewin:
I have read through and attempted to follow all your test modification, contribution, and PR guidelines. Please accept my apology if I booted it somehow.

This is the "sister" commit for [problem-specifications PR #1131](https://github.com/exercism/problem-specifications/pull/1131).

When attempting the clock exercise in Ruby, I found that the test cases added negative minute amounts rather than subtracting positive minute amounts. This created a situation where, despite what the README said, the exercise did not require the creation of a subtraction operation.

The PR above contains the changes related to the updates to `canonical-data.json`, and this PR contains the commit related to updating the Ruby test generator and the test file itself.


kotp:
> There is no such thing as subtraction, only the addition of negative amounts... arguably.  Is there a requirement for a subtraction method? -- Devils Advocate

I like an idea of an 'adjustment' rather than the idea of specifically adding.  It simplifies the model, and mirrors that little dial on my wind up alarm clock that has little +/- etchings in it.

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,30 @@ def name
...
-      "#{'(' if add_to_clock}Clock.at(#{hour}, ",
-      "#{minute})#{add_to_clock}#{')' if add_to_clock}.to_s"
-    ].join
+  def create_test
+    "#{ASSERTION_STR}, #{CLOCK_STR}.to_s" % [expected.inspect, hour, minute]
   end
 
-  def add_to_clock
-    " + #{add}" if respond_to?(:add)
+  def arithmetic_test(op, value)
```
kotp: Should this be `adjust_clock` instead of testing arithmetic?

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,30 @@ def name
...
   end
 
-  def simple_test
-    [
-      "assert_equal #{expected.inspect}, ",
-      "#{'(' if add_to_clock}Clock.at(#{hour}, ",
-      "#{minute})#{add_to_clock}#{')' if add_to_clock}.to_s"
-    ].join
+  def create_test
+    "#{ASSERTION_STR}, #{CLOCK_STR}.to_s" % [expected.inspect, hour, minute]
```
Insti: I like the idea here, simplifying test construction and removing the conditionals, but don't like this specific change.
It squashes the assertion, the expected value, and the subject of the test into one without any clarity about which is which. 

What about something like:
```ruby 
actual = "Clock.at( #{hour}, #{minute} )"
"assert_equal #{expected.inspect}, #{actual}"
```

Insti:
Will this still work with the [updated canonical data](https://github.com/exercism/problem-specifications/pull/1145)?

MattLewin:
@Insti, this will work with the updated canonical data. My suggested update of the canonical data was the reason I also submitted this PR.

Re your earlier comment re the generator squashing everything together. I was simply trying to produce a generator I thought was useful. (I thought it'd be kind of rude to propose new canonical data and not update at least one generator.) I am not at all married to my proposed implementation, as I am by no means a Ruby stud.

I'm happy to go with your implementation. Would you like me to make the changes?

kotp:
@mattkimmel I think making @Insti 's suggested changes (unless you can reason something that would be better) is a good direction. 

Specifically, the one change I am looking at was toward this result:

```ruby
actual = "Clock.at( #{hour}, #{minute} )"
"assert_equal #{expected.inspect}, #{actual}"
```

It has benefits, not the least which is, I believe, still the pattern being followed in other exercise tests as well, as they are generated.

PR #794, opened 2018-01-23
Singularize complex numbers.

### LINKED ISSUE #792: complex-number exercise file naming is inconsistent
The complex-number exercise ships with a test file named `complex_numbers_test.rb` and the minitest class is `ComplexNumbersTest`. But the working class referenced in the test is `ComplexNumber`. 

For consistency, I believe the test file should be named `complex_number_test.rb` and the minitest class should be `ComplexNumberTest`.

mftaff:
closes #792 

Note: I started with only the changes requested by @moveson but in order to get the tests to pass I needed to make complex numbers singular everywhere it appeared.

### DIFF
--- a/exercises/complex-numbers/complex_numbers_test.rb
+++ b/exercises/complex-numbers/complex_numbers_test.rb
@@ -2,7 +2,7 @@ require 'minitest/autorun'
 require_relative 'complex_numbers'
 
 # Common test data version: 1.0.0 '117d062'
-class ComplexNumbersTest < Minitest::Test
+class ComplexNumberTest < Minitest::Test
   def test_imaginary_unit
     # skip
     expected = ComplexNumber.new(-1, 0)


Insti:
Thanks for taking this on @mftaff.
What you've done here looks good from a technical perspective.

However, being based on the `complex-numbers` exercise in the [problem-specificaitons repository](https://github.com/exercism/problem-specifications/tree/master/exercises/complex-numbers) the system/tooling expects the files to be called `complex-numbers`.

So, if it were important, the consistency here would need to happen in the other direction, 

BUT, is `ComplexNumbers` and appropriate name for the class if it is only representing a single complex number?

mftaff:
@Insti 😸 I am not sure I fully understand your message. Could you elaborate on: 
`So, if it were important, the consistency here would need to happen in the other direction`
and 
`BUT, is ComplexNumbers and appropriate name for the class if it is only representing a single complex number?` 

Thanks!! (looking forward to making the desired changes)

Insti:
> if it were important, the consistency here would need to happen in the other direction

By 'in the other direction' I mean that the class implementing the solution would need to be renamed to `ComplexNumbers` rather than renaming the files as you have done here.

> is `ComplexNumbers` (and was a typo) **an** appropriate name for the class if it is only representing a single complex number?

I'm not sure changing the name of the class is the right thing to do, and maybe we should just live with the "inconsistency" that currently exists. I was wondering what you thought the class should be called.

mftaff:
I hear ya,
The reason I made the change was based on the issue (#792) raised by @moveson 
Maybe we can bring him into the conversation - I don't feel informed enough to meaningfully respond, as I only just starting delving into the source code.

mftaff:
@Insti  Okay the new commit reflects our conversation from #792

Insti:
Great, this looks right, thanks.
Sorry the whole process has been so confusing!

PR #799, opened 2018-03-28
Rename 'BookKeeping' to 'Acronym' in acronym test

mcdickenson:


kotp:
I believe this is OK, as it stands, no pull request required.  Compare to  [Gigasecond Test Suite](http://exercism.io/exercises/ruby/gigasecond/test-suite) and the original is as it should be.  It is testing the BookKeeping, which is helping the reviewer know which version of the test and exercise your solution is written against.

mcdickenson:
That makes sense, thanks @kotp 👍🏻

kotp:
Thank you very much @mcdickenson for contributing.

I checked your Ruby exercises, and noticed it has been a few years (for the ones I looked at).  So not surprised it took you by surprise.  Welcome back to Exercism!

PR #802, opened 2018-04-14
FAKE+NEWS=LIES found to be alphametic

n8chz:
One more test case added.


landongrindheim:
#803 Should resolve the error that caused the [initial CI failure](https://travis-ci.org/exercism/ruby/builds/366548840)

Insti:
Thanks for your contribution.

I'm going to close this PR for now.

While this is a potentially amusing example, does it exercise any part of the solution that was not previously exercised? 

Is there code that will pass all the other tests that will fail on this case? 

If this PR does test a previously untested aspect of the exercise we can re-open it.

Thanks.

Issue #24, opened 2014-07-13
submissions for grains exercise don't appear on the website


patbl:
It looks as though there's a bug preventing me from submitting revisions to the [grains exercise](http://exercism.io/submissions/b0143a6796e9026477140411) (but not others). When I try to submit my solution, I get this error:

```
There was an issue with your submission: Status: 400, Error: {     This attempt is a duplicate of the previous one.}
```

When I change my solution and resubmit, I don't get any error message, but my solution doesn't appear on the website. And if I try to resubmit the new solution, I get the previous error message.

PR #803, opened 2018-04-23
800 Remove dependency on require_all

landongrindheim:
From require_all's changelog:
  2.0.0:
  Merged PR #24 (https://github.com/jarmo/require_all/pull/24) Prior to
  version 2, RequireAll attempted to automatically resolve dependencies
  between files, thus allowing them to be required in any order. Whilst
  convenient, the approach used (of rescuing NameErrors and later retrying
  files that failed to load) was fundamentally unsafe and can result in
  incorrect behaviour (for example issue #8, plus more detail and
  discussion in #21).

  Thanks to Joe Horsnell (@joehorsnell)

This commit removes the dependency on the require_all gem by using
Ruby's `#require` method wherever it is needed.

### DIFF
--- a/lib/generator/implementation.rb
+++ b/lib/generator/implementation.rb
@@ -1,5 +1,6 @@
 require 'delegate'
 require 'forwardable'
+require_relative 'template_values'
 
 module Generator
   class Implementation
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,5 @@ source 'https://rubygems.org'
 
 gem 'minitest'
 gem 'rake'
-gem 'require_all'
 gem 'rubocop', '0.36.0'
 gem 'simplecov'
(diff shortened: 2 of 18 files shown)


lib/generator/implementation.rb
```diff
@@ -1,5 +1,6 @@
 require 'delegate'
 require 'forwardable'
+require_relative 'template_values'
```
Insti: Any reason you've used `require_relative` here?  and `relative` everywhere else?

Insti:
Thanks, this is good.
The fewer dependencies the better.
Explicitly requiring what is needed is better than magically including everything (which was a result of me being lazy while I was initially developing all this.)

Thanks.

Insti:
Hmm this passes the tests, but still fails when trying to run `generate --all`

PR #804, opened 2018-04-27
Add alphanumeric pattern generator exercise

rdavid1099:
Proposal (and implementation) to add new exercise.

Alphanumeric pattern generator and validator. Using `.` and `#`, the class has a few methods to generate patterns of given nth value or randomly. Furthermore, the class can validate if a given string matches the pattern its initialized with. The class can also return the total number of available combinations for a given pattern.


rdavid1099:
Due to the issue with `RequireAll`, the build is not passing the CI check. There appears to be a PR with a fix. Once that is in master, I can merge the changes into my branch and pass the CI check.

Insti:
Is it correct that this is a new exercise you have invented and it is not present in the https://github.com/exercism/problem-specifications repository?

Insti:
The CI fixes are now in master and I re-ran the CI check for this PR.

Insti:
There is a generator included here, but it does not work since there is no `canonical-data.json` to generate the data from.
I'd recommend removing the generator from this PR.

exercises/alphanumeric-pattern-generator/README.md
```diff
@@ -0,0 +1,33 @@
+# Alphanumeric Pattern Generator
+
+Create a program that can generate and manage [alphanumeric](https://en.wikipedia.org/wiki/Alphanumeric) strings based off a given pattern.
```
Insti: I'd use `based on` here, but this may be a regional thing.

exercises/alphanumeric-pattern-generator/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+An alphanumeric pattern is a string that contains alphabetic and numeric characters. In our example, `.` represents alphabetic characters (a-z) and `#` represents numeric characters (0-9), so the pattern `.#` could produce anything from `A0` to `Z9`.
+
+An alphanumeric patterns can also have constant values. For example, the pattern `#.JH1.` could generate any pattern including the values `JH1` as index 2 ,3 and 4 respectively such as `9FJH1P`.
+
```
Insti: I bit more information about what the task actually is would be helpful.

exercises/alphanumeric-pattern-generator/alphanumeric_pattern_generator_test.rb
```diff
@@ -0,0 +1,156 @@
+require 'minitest/autorun'
+require_relative 'alphanumeric_pattern_generator'
+
+class PatternGeneratorTest < Minitest::Test
```
Insti: class name should probably have 'Alphanumeric' at the start.

exercises/alphanumeric-pattern-generator/alphanumeric_pattern_generator_test.rb
```diff
@@ -0,0 +1,156 @@
...
+  def test_it_can_verify_a_single_alphabetic_character
+    pattern = '.'
+    pg = PatternGenerator.new(pattern)
+
+    assert_equal true, pg.verify('F')
+    assert_equal false, pg.verify('0')
```
Insti: try and only have one assertion per test, if you've got more than one then you're probably testing more than one thing and therefore can give each thing it's own descriptive test name.

exercises/alphanumeric-pattern-generator/alphanumeric_pattern_generator_test.rb
```diff
@@ -0,0 +1,156 @@
...
+    assert_equal true, pg.verify('r')
+    assert_equal true, pg.verify('R')
+  end
+
+  def test_it_generates_a_single_character_pattern
+    pattern = '.'
+    pg = PatternGenerator.new(pattern)
+
+    assert /[A-Z]/.match(pg.generate)
+    refute /[0-9]/.match(pg.generate)
```
Insti: Try and be consistent about usage of `assert/refute` and `assert_equal true/assert_equal false`
I'd prefer using `assert` and `refute` everywhere.

exercises/alphanumeric-pattern-generator/alphanumeric_pattern_generator_test.rb
```diff
@@ -0,0 +1,156 @@
...
+    pattern = '.#.'
+    pg = PatternGenerator.new(pattern)
+
+    assert_equal 'A0A', pg.generate(0)
+    assert_equal 'A1B', pg.generate(27)
+    assert_equal 'Z9Z', pg.generate(6759)
+    assert_equal 'Z9Y', pg.generate(6758)
+  end
+
+  def test_it_raises_exception_if_invalid_valure_is_passed_to_generate
```
Insti: typo in this line.

exercises/alphanumeric-pattern-generator/alphanumeric_pattern_generator_test.rb
```diff
@@ -0,0 +1,156 @@
...
+    pattern = '.#.'
+    pg = PatternGenerator.new(pattern)
+
+    assert_raises(ArgumentError) { pg.generate(10000) }
```
Insti: why is this an argument error?

Insti:
This exercise seems to be trying to test a lot of different things all at the same time.
What is the main point of this exercise? 
What aspect of programming are you trying to demonstrate?
Which of the things you are testing for here could be left unspecified because it doesn't REALLY matter if they are implemented or not?

Insti:
Thanks for your contribution :heart:, I've gone through it and added some feedback.

rdavid1099:
Been on a trip for business... Checking in on this PR now. I will implement changes and provide responses when I get a chance tomorrow. Thank you for all the comments, Geoff.

rdavid1099:
Been on a trip for business... Checking in on this PR now. I will implement changes and provide responses when I get a chance tomorrow. Thank you for all the comments, Geoff.

exercises/alphanumeric-pattern-generator/alphanumeric_pattern_generator_test.rb
rdavid1099: The consistency was broken due to the truthy/ falsey nature of `match`. I try to stay away from `assert` and `refute` if I am specifically looking for the values `true` and `false`, hence the reason I am mostly using `assert_equal` throughout the test suite. Testing regex with match returns either an integer or nil which is why I changed my patter to `assert` and `refute`.

exercises/alphanumeric-pattern-generator/alphanumeric_pattern_generator_test.rb
rdavid1099: `ArgumentError` may not be the most intuitive exception to throw. My idea behind it is that the number passed into `generate` is out of range of the total possible alphanumeric combinations for the given pattern.

exercises/alphanumeric-pattern-generator/alphanumeric_pattern_generator_test.rb
rdavid1099: Fair enough. I will change them to `assert_equal 0, /[A-Z]/.match(pg.generate)`. Thank you.

rdavid1099:
@Insti Thank you for all of the constructive feedback. I made the suggested changes and fixed the typos/ naming issues you pointed out.

To answer your question regarding the exercise's purpose. My goal was to build out a fun little generator class which utilizes the power of Ruby enumerables and basic math with a given pattern.  When I was first learning Ruby, I completed a similar exercise and learned a lot more of how to use enumerables in a practical manner.

With all that being said, I do think that we could remove the exception throwing. That does stand out as "out-of-scope" of the goals of the exercise.

What would you suggest?

PR #807, opened 2018-05-12
Fix incorrect documentation

Eventlessdrop:
A test suite will never have 0 assertions, and so I replaced line 39 to say that there were 16 assertions process opposed to 0.


kotp:
Thanks @Eventlessdrop!

PR #816, opened 2018-07-19
Add topics to config file

PatrickMcSweeny:
I noticed that the progress bars in the 'concepts' section weren't updating and I informed that this was because the exercises don't have topics assigned to them. This isn't a complete fix, but it's a start.

### DIFF
--- a/config.json
+++ b/config.json
@@ -34,6 +34,7 @@
       "difficulty": 1,
       "topics": [
         "time",
+        "mathematics"
       ]
     },
     {
@@ -215,7 +216,10 @@
       "core": false,
       "unlocked_by": null,
       "difficulty": 1,
-      "topics": null
+      "topics": [
+        "strings",
+        "parsing"
+      ]
     },
     {
       "slug": "prime-factors",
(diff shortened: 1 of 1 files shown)


Insti:
Awesome, thanks.

Unfortunately there seems to be a problem with the config.json see: https://travis-ci.org/exercism/ruby/builds/405892767#L1809

Do you think you could look through the file again and see if you can find the problem?

(It's also possible to run the `configlet` command on your own machine so you can check it there.)

config.json
```diff
@@ -11,87 +11,124 @@
...
     },
     {
       "slug": "gigasecond",
       "uuid": "0fb594a1-193b-4ccf-8de2-eb6a81708b29",
       "core": false,
       "unlocked_by": null,
       "difficulty": 1,
-      "topics": null
+      "topics": [
+        "time",
```
Insti: It might be this comma

PatrickMcSweeny:
I fixed the issue with the trailing comma and added a few more topics in the process.

Insti:
Great! Do you want to keep adding topics or shall we merge this as it is now?

PatrickMcSweeny:
This is all I want to do for right now. I will probably add more topics later as I complete more of the exercises myself.

Insti:
Thanks @PatrickMcSweeny :heart:

PR #818, opened 2018-07-27
Use progression from Go

koppen:
This is an attempt at setting up the Ruby track as a nextercism track.

As suggested in #815 I have cloned the Go tracks progression, removing exercises that are Go-specific, preserving Ruby-only exercises as bonus exercises.

After this changes the progression tree looks like this:

```
Ruby
====

core
----
├─ hello-world
│  └─ space-age
│
├─ two-fer
│  ├─ leap
│  ├─ triangle
│  └─ proverb
│
├─ hamming
│  ├─ rna-transcription
│  └─ nucleotide-count
│
├─ raindrops
│  ├─ accumulate
│  ├─ roman-numerals
│  └─ strain
│
├─ scrabble-score
│  ├─ etl
│  └─ protein-translation
│
├─ isogram
│  ├─ pangram
│  ├─ anagram
│  ├─ word-count
│  └─ run-length-encoding
│
├─ difference-of-squares
│  ├─ sum-of-multiples
│  ├─ pythagorean-triplet
│  ├─ largest-series-product
│  ├─ nth-prime
│  └─ isbn-verifier
│
├─ luhn
│  ├─ crypto-square
│  ├─ atbash-cipher
│  ├─ simple-cipher
│  └─ rotational-cipher
│
├─ grains
│  ├─ secret-handshake
│  ├─ sieve
│  ├─ perfect-numbers
│  ├─ allergies
│  └─ all-your-base
│
├─ clock
│  ├─ phone-number
│  ├─ pig-latin
│  ├─ prime-factors
│  └─ meetup
│
├─ robot-name
│  ├─ palindrome-products
│  ├─ say
│  └─ wordy
│
├─ tournament
│  ├─ pascals-triangle
│  ├─ diamond
│  ├─ ocr-numbers
│  ├─ kindergarten-garden
│  ├─ poker
│  └─ bowling
│
├─ twelve-days
│  ├─ house
│  ├─ food-chain
│  └─ beer-song
│
├─ matrix
│  ├─ queen-attack
│  ├─ transpose
│  ├─ saddle-points
│  ├─ minesweeper
│  └─ connect

bonus
-----
gigasecond
bob
acronym
flatten-array
series
scale-generator
bracket-push
rail-fence-cipher
circular-buffer
custom-set
grade-school
simple-linked-list
linked-list
binary-search
binary-search-tree
robot-simulator
change
collatz-conjecture
book-store
dominoes
alphametics
two-bucket
list-ops
affine-cipher
complex-numbers
zipper
```

[...]

PR #819, opened 2018-07-27
Run configlet fmt on `config.json`

Insti:
Ran configlet fmt on `config.json`

The result seems to be mostly re-ordering tags alphabetically.


Insti:
#818 is also correctly formatted.

PR #821, opened 2018-07-28
Unlock beer song

Insti:
The 'Beer song' exercise is the main focus of the book: 99 bottles
(https://www.sandimetz.com/99bottles)

It would be good if users could access the tests and specification
without having to unlock it.


Insti:
Hmm it still appears as locked on my dummy student account..

PR #518, opened 2016-12-28
Add phony target for test in Makefile

tommyschaefer:
Hi there!

I noticed while looking at the Travis CI output yesterday that the tests don't appear to be running.

From Travis CI log output:

```
$ make test
```

```
make: `test' is up to date.

The command "make test" exited with 0.
```

This is happening because we added a `test` directory, and the Makefile target is also named `test`.

If this looks good to you all and gets merged, I will rebase and push #514, #510, and #509 to run CI with the tests.

Thanks so much for your time!

PR #827, opened 2018-08-04
Change unlocking order step 1

F3PiX:
This:
- makes Hello World unlock a few basic exercises, (by changing `'unlocked_by: null` to `hello-world`)
- make all other exercises currently `unlocked_by: null` move up one or two levels ( unlocked either by TwoFer or by Hamming).

In #518 the request was to restrict the number of `unlocked_by: null` to 4 or 5 exercises.
Am I right that `unlocked_by: hello-world` has the same effect in the track? I used that, to make it more explicit. 

This PR is an intermediate step in the Ruby Overhaul™.   
As a temporary step, all the other exercises that used_to be `unlocked_by: null` are now moved up 1 or 2 levels, so either by TwoFer for the lowest difficulties, or Hamming, for most others, and some with the highest difficulties quite arbitrary by later core exercises.

I also lowered the difficulty for a few exercises, but they should still be in the same 'easy' category.

### DIFF
--- a/config.json
+++ b/config.json
@@ -425,7 +425,7 @@
       "slug": "acronym",
       "uuid": "74468206-68a2-4efb-8caa-782634674c7f",
       "core": false,
-      "unlocked_by": "flatten-array",
+      "unlocked_by": "two-fer",
       "difficulty": 2,
       "topics": [
         "regular_expressions",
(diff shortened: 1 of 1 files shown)


Insti:
Cool thanks. I'm going to have to leave it to someone else to be responsible for merging this as I don't think I should be involved in V2 track organisation. 

@hilary seems to have been following V2 closely and I think has the right permissions here. 

Or you could try and track down one of the other [maintainers](https://github.com/exercism/ruby/blob/master/config/maintainers.json)

kotp:
This looks good to me, @hilary how about you?

hilary:
Looks good to me.

PR #830, opened 2018-08-06
clock: refactored test; new generator; solution updated to pass

### LINKED ISSUE #828: Clock: Does the test force a non-idiomatic solution?
Hi all,

While doing the clock exercise I noticed:
- The test expects #at to be a class method.
- In order to deal with tests of the form `(Clock.at(12, 12) + 12).to_s` the solution requires making an instance of the class.
  - This means most solutions call `Clock.new` inside `def self.at`.
  - I am not an expert, but this test seems like it forces a hack rather than an idiomatic solution.
- With a single class method I was able to pass all tests except those tests that added minutes after the class method call, so I am wondering if they should be excluded.
- Or perhaps the test should be refactored to work with a Clock class object instead of using class methods.
- I'm thinking that refactoring the test not to use class methods would make the solutions more idiomatic while allowing all of the test to be present.

Thank you

guygastineau:
The amended solution passes all of the tests.

I had to break up the compare clock tests in order for a formatted string to be returned.

I am a bit stuck in the generator right now and would appreciate any help or guidance with that.
- Currently it fails saying that there is no variable named `clock1`.

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/clock/canonical-data.json)

closes #828


exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,31 @@ def name
   end
 
   def workload
-    property == 'equal' ? compare_clocks : simple_test
+    if property == 'create'
```
Insti: There are 4 known properties in the canonical-data you probably should list them all and raise an error if you find an unexpected one.

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,31 @@ def name
...
+      compare_clocks
+    end
   end
 
   private
 
   def compare_clocks
-    "clock1 = Clock.at(#{clock1['hour']}, #{clock1['minute']})
-    clock2 = Clock.at(#{clock2['hour']}, #{clock2['minute']})
+    "clock1 = Clock.new(hour: #{clock1['hour']}, minute: #{clock1['minute']})
```
Insti: where do you think `clock1` and `clock2` are coming from?

Insti: Oh canonical data.
I think you need to do `input['clock1']['hour']` etc

exercises/clock/clock_test.rb
```diff
@@ -5,281 +5,313 @@
 class ClockTest < Minitest::Test
```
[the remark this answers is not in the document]
guygastineau: There was one there already. I changed it.

exercises/clock/clock_test.rb
```diff
@@ -5,281 +5,313 @@
...
     skip
-    assert_equal "22:10", Clock.at(-121, -5810).to_s
+    assert_equal "22:10", Clock.new(hour: -121, minute: -5810).to_s
   end
 
   def test_add_minutes
     skip
-    assert_equal "10:03", (Clock.at(10, 0) + 3).to_s
+    clock = Clock.new(hour: 10, minute: 0)
+    clock + Clock.new(minute: 3)
```
Insti: `+` should still work if a clock is immutable. So I'd put the wrapping parentheses back.

 assert_equal "10:03", (clock + Clock.new(minute: 3)).to_s

exercises/clock/clock_test.rb
```diff
@@ -5,281 +5,313 @@
...
+    assert_equal "10:03", clock.to_s
   end
 
   def test_add_no_minutes
     skip
-    assert_equal "06:41", (Clock.at(6, 41) + 0).to_s
+    clock = Clock.new(hour: 6, minute: 41)
```
Insti: I'd call this `clock1` to match what happens latter with `clock1` and `clock2`

exercises/clock/.meta/generator/clock_case.rb
Insti: We also need tests for "sub"/`:-`

These should use the - operator rather than being adding negative minutes.

`assert_equal "10:03", (clock1 - Clock.new(minute: 3)).to_s`

Insti:
This all looks really good, you're on the right track. :+1:

guygastineau:
Thanks, hopefully I'll have a functional generator and these other tweaks up tonight

exercises/clock/clock_test.rb
```diff
@@ -5,281 +5,313 @@
...
   def test_sixty_minutes_is_next_hour
     skip
-    assert_equal "02:00", Clock.at(1, 60).to_s
+    assert_equal "02:00", Clock.new(hour: 1, minute: 60).to_s
   end
 
   def test_minutes_roll_over
     skip
-    assert_equal "02:40", Clock.at(0, 160).to_s
+    assert_equal "02:40", Clock.new(hour: 0, minute: 160).to_s
```
Insti: I think in this first section we should leave out the `hour:` or `minute:` argument when its value is zero.
Ideally we'd have separate tests for default keyword values, but sneaking the requirement in this way gets it done without having to do anything too special to the canonical tests. And it doesn't suddenly come as a huge surprise when we hit the `add` tests.

guygastineau: I like that idea.

guygastineau: Do you only want it left out for the tests where `property == 'create'` ?
Or should they be left out for all of them?

guygastineau:
Wanted to get this version of the generator and test up here.
I still need to add logic to the test in order to omit `hour:` and `minute:` when theirvalue is 0.

guygastineau:
It seems to me that the regular #to_s is being called instead of the `to_s` defined in the solution, and somehow it is grabbing the total minutes after `+` or `-`.

Like I said, I am pretty lost about how to fix that in the solution.

guygastineau:
I fixed that by adding `self` to the end of the `-` and `+` methods.  adding passes now, but the subract tests are getting the wrong answer.  I'm working on that now.

guygastineau:
Got the solution fixed.  Now to apply the logic in the generator.

guygastineau:
The test now excludes `hour: 0` and `minute: 0` for cases where `property == create`.
All of the tests are passing.

Let me know what you all think.  It might be ready for WIP to be removed.

Should I comment in the generator for future's sake?

guygastineau:
I need to run `bin/generate --update clock` in order to get the BookKeeping check in the test, right?

guygastineau:
With that last commit I assume this can shed the `WIP` from it's title.

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,49 @@ def name
   end
 
   def workload
-    property == 'equal' ? compare_clocks : simple_test
+    case property
+    when 'create'
+      simple_test
```
Insti: I find the when/then format works well in this sort of situation.
`when 'create' then simple_test`

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,49 @@ def name
...
+    when 'add'
+      add_to_clock
+    when 'subtract'
+      subtract_from_clock
+    when 'equal'
+      compare_clocks
+    else
+      raise "Encountered unknown property in canonical-data.json"
```
Insti: Nice error message :+1:

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,49 @@ def name
   end
 
   def workload
-    property == 'equal' ? compare_clocks : simple_test
+    case property
```
Insti: Good usage of a case statement :+1:

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,49 @@ def name
...
+      "Clock.new(#{"hour: #{input['hour']}" unless input['hour'] == 0}",
+      "#{", " unless input['hour'] == 0 || input['minute'] == 0}",
+      "#{"minute: #{input['minute']}" unless input['minute'] == 0}).to_s"
     ].join
   end
 
   def add_to_clock
-    " + #{add}" if respond_to?(:add)
+    "clock1 = Clock.new(hour: #{input['hour']}, minute: #{input['minute']})
+    assert_equal #{expected.inspect}, (clock1 + Clock.new(minute: #{input['value']})).to_s"
```
Insti: Here we're relying on the fact that the indentation of this file is the same as the indentation of the final test file, but this is just a coincidence.
You probably should use the [indent_lines helper method](https://github.com/exercism/ruby/blob/master/lib/generator/exercise_case/case_helpers.rb#L14)

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,49 @@ def name
...
   end
 
+  def subtract_from_clock
+    "clock1 = Clock.new(hour: #{input['hour']}, minute: #{input['minute']})
+    assert_equal #{expected.inspect}, (clock1 - Clock.new(minute: #{input['value']})).to_s"
```
Insti: indent_lines again.

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,49 @@ def name
...
 
   private
 
   def compare_clocks
-    "clock1 = Clock.at(#{clock1['hour']}, #{clock1['minute']})
-    clock2 = Clock.at(#{clock2['hour']}, #{clock2['minute']})
+    first_hour = input['clock1']['hour']
+    first_minute = input['clock1']['minute']
+    second_hour = input['clock2']['hour']
+    second_minute = input['clock2']['minute']
```
Insti: I like this usage of temp variables to tidy up and clarify :+1:

exercises/clock/.meta/generator/clock_case.rb
```diff
@@ -9,27 +9,49 @@ def name
...
   end
 
   def simple_test
     [
       "assert_equal #{expected.inspect}, ",
-      "#{'(' if add_to_clock}Clock.at(#{hour}, ",
-      "#{minute})#{add_to_clock}#{')' if add_to_clock}.to_s"
+      "Clock.new(#{"hour: #{input['hour']}" unless input['hour'] == 0}",
+      "#{", " unless input['hour'] == 0 || input['minute'] == 0}",
+      "#{"minute: #{input['minute']}" unless input['minute'] == 0}).to_s"
```
Insti: It's hard to see what is happening here, and those conditionals scare me. What about:
```ruby
    keyword_arguments = %w[hour minute].map { |key| [key, input[key]] }
    keyword_arguments.reject! { |_, value| value.zero? }
    keyword_arguments.map! { |key, value| "#{key}: #{value}" }

    clock_builder = "Clock.new(#{keyword_arguments.join(', ')}).to_s"
    "assert_equal #{expected.inspect}, #{clock_builder}"
  end
```

exercises/clock/.meta/solutions/clock.rb
```diff
@@ -1,24 +1,23 @@
...
 
 # Clock without dates exercise
 class Clock
-  def self.at(*args)
-    Clock.new(*args)
-  end
 
-  def initialize(hours=0, minutes=0)
-    @internal = hours * 60 + minutes
+  attr_reader :internal
```
Insti: Is it sensible to have a method called `internal` that can be accessed externally?
What does it represent? Why not call it that?

Insti:
This is looking good :+1:
I think the test file itself looks fine.
I have a few style related comments on the generator code, but what you have works. :heart:

exercises/clock/.meta/generator/clock_case.rb
guygastineau: You mean for when and the block on the same line?

exercises/clock/.meta/generator/clock_case.rb
guygastineau: Nice, I didn't really want to have to embed that logic the way it is right now.

I really like your suggestion here.  It took me a minute to get how it works, but it is awesome.

Thank you.

exercises/clock/.meta/solutions/clock.rb
guygastineau: I agree.  I just left it in from the last solution.

You think `total_minutes` would be a good would be good there.

exercises/clock/.meta/generator/clock_case.rb
Insti: yes.

exercises/clock/.meta/generator/clock_case.rb
Insti: Those that return multiple lines of output anyway.

exercises/clock/.meta/solutions/clock.rb
Insti: That would be fine. Definitely a better name than `internal`

Insti:
I think I responded to all your responses, let me know if I missed anything.

exercises/clock/.meta/solutions/clock.rb
guygastineau: Thanks

exercises/clock/.meta/generator/clock_case.rb
guygastineau: Got `indent_lines` on three of them, and used `indent_test` on the one line test for good measure.

exercises/clock/.meta/generator/clock_case.rb
guygastineau: I left the else and error message on two lines. Is that fine?

guygastineau:
@Insti 
Those were great reviews on the last commit. Thank you.

I think I got everything in accordance with those suggestions plus a little extra tidying along the same lines.

Let me know what you think :smile:

guygastineau:
@F3PiX and @kotp 

If you have time please let me know what you all think too.
I appreciate your opinions :smiley:

guygastineau:
@Insti 

Thanks. And Thanks for all your guidance.  After this gets merged I will make a generator for affine-cipher.

Do you know how to get rid of the WIP status? I thought that removing it from the title would do the trick.

Insti:
> Do you know how to get rid of the WIP status? I thought that removing it from the title would do the trick.

No :( I thought it should too.
I tried changing the title in case it was a permissions thing, but apparently not..

petertseng:
From https://github.com/apps/wip

> One of the pull request commit messages

Insti:
I can merge it even with the WIP bot check failing so I wouldn't bother amending commit messages.

guygastineau:
@petertseng @Insti 

Thanks guys!

guygastineau:
Are we waiting on more reviews?

Insti:
> Are we waiting on more reviews?

Yes, sometimes I wait a couple of days for one of the other maintainers to have a quick look over it, (or myself with fresh eyes) but it's been a couple of days so we can probably merge this :)

Insti:
Thanks for all your work on this @guygastineau :heart:

guygastineau:
Thank your for all your help with this.

guygastineau:
I'm excited to go do this exercise now!

PR #832, opened 2018-08-10
affine-cipher: built generator

guygastineau:
Hi all. I figured I would tackle the generator for affine-cipher now that I have some generator experience.

I did the following:
1. built the generator
2. generated the test
3. ran `bin/generate --update affine-cipher`
4. added the BookKeeping module to the solution for `VERSION = 1`

----
The original test I made when I ported the exercise didn't store the input['phrase'] and expected in variables for every test, but I did this in the generator.  I thought it would:
1. make the generator clearer and easier to implement/update.
2. make things clearer to the student when doing the exercise.


Insti:
[affine-cipher description](https://github.com/exercism/problem-specifications/blob/master/exercises/affine-cipher/description.md)
[affine-cipher canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/affine-cipher/canonical-data.json)

Insti:
(This is an issue for problem-specifications repository.)
There is a problem with the canonical_data.json file:

From the [problem-specifications README](https://github.com/exercism/problem-specifications#test-data-format-canonical-datajson)
> If an error is expected (because the input is invalid, or any other reason), the value at "expected" should be an object containing exactly one property, "error", whose value is a string. 

` "expected": "Error: a and m must be coprime."`
should be:
` "expected": { error: "Error: a and m must be coprime." }`

exercises/affine-cipher/affine_cipher_test.rb
```diff
@@ -1,54 +1,69 @@
...
 require_relative 'affine_cipher'
 
-class AffineTest < Minitest::Test
+# Common test data version: 1.0.0 377d917
+class AffineCipherTest < Minitest::Test
   def test_encode_yes
-    #skip
+    # skip
     cipher = Affine.new
     cipher.addkey(5, 7)
```
Insti: Do we need a separate addkey method?
Why can we not just do: `Affine.new(5,7)`

exercises/affine-cipher/.meta/generator/affine_cipher_case.rb
```diff
@@ -0,0 +1,52 @@
...
+        "cipher.addkey(#{a}, #{b})",
+        "plaintext = '#{input['phrase']}'",
+        "ciphertext = '#{expected}'",
+        "assert_equal ciphertext, cipher.encode(plaintext)"
+      ], 4
+    )
+  end
+
+  def decode_workload
+    a, b = input['key']['a'], input['key']['b']
```
Insti: I'd use `key_a` and `key_b` as variable names.

Insti:
Hi, Thanks, this looks great :heart:

I think this is well done and is certainly good enough to merge as is :+1: 

The `canonical_data.json` needs fixing, so it might be good to do that first and then make the minor updates to the generator that will require.

I've added some comments about some things I noticed in the code which I would do differently. These are all very much my own opinion. You don't need to change anything if you don't want to.

exercises/affine-cipher/affine_cipher_test.rb
guygastineau: That's a good point. I think my reasons were twofold:
1. While developing this exercise I first made programs in bash and ruby with no tests.  I have since developed the Affine class from that ruby program into one that can be instanciated, load text, and have it's keys change. Essentially I set it up so the object could be used to brute force against a ciphertext with a lost key easily.
  - I realize that this implementation necessitated by the tests.
2. I thought it would be clearer that the keys were being added if they had their own function.
  - Although initially I wanted the keys to be added with `cipher.key = [5, 7]` or something.
#### I realize it will probably be more idiomatic and fitting for exercism if the keys are just added with new.
I'll make that change in the generator if you think it is best.

exercises/affine-cipher/.meta/generator/affine_cipher_case.rb
guygastineau: sounds good

exercises/affine-cipher/.meta/generator/affine_cipher_case.rb
```diff
@@ -0,0 +1,52 @@
...
+class AffineCipherCase < Generator::ExerciseCase
+
+  def workload
+    if expected.start_with?("Error:")
+      error_workload
+    elsif property == "encode"
+      encode_workload
+    elsif property == "decode"
+      decode_workload
+    end
```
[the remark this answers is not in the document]
guygastineau: Nice, I really like case statements, and that is a cool guard statement.

guygastineau:
@Insti 
I also meant to ask:
Is there a way to check the value of the string returned by an ArgumentError, or is that against track policy?

guygastineau:
I got the PR to `problem-specifications`.  I'll change the status to WIP until that goes through.

guygastineau:
I ran `bin/generate --update affine-cipher` again and it now made the test version 2.

I assume this was the right course of action in this case even though version 1 of the test was never merged to master?

Please let me know if it should be version 1.

exercises/affine-cipher/.meta/.version
```diff
@@ -0,0 +1 @@
+2
```
Insti: The first published version should always be 1.

Insti:
For reference, the PR that fixed the canonical-data: https://github.com/exercism/problem-specifications/pull/1295

exercises/affine-cipher/affine_cipher_test.rb
```diff
@@ -1,122 +1,149 @@
...
-class AffineTest < Minitest::Test
+# Common test data version: 2.0.0 8026923
+class AffineCipherTest < Minitest::Test
   def test_encode_yes
-    #skip
-    cipher = Affine.new
-    cipher.addkey(5, 7)
-    assert_equal 'xbt', cipher.encode('yes')
+    # skip
+    cipher = Affine.new(5, 7)
```
Insti: Passing the parameters in via `new` is much better :+1:

exercises/affine-cipher/affine_cipher_test.rb
```diff
@@ -1,122 +1,149 @@
...
     plaintext = 'The quick brown fox jumps over the lazy dog.'
     ciphertext = 'swxtj npvyk lruol iejdc blaxk swxmh qzglf'
     assert_equal ciphertext, cipher.encode(plaintext)
   end
 
   def test_encode_with_a_not_coprime_to_m
     skip
-    cipher = Affine.new
-    assert_raises(ArgumentError) { cipher.addkey(6, 17) }
+    assert_raises(ArgumentError) { Affine.new(6, 17) }
```
Insti: What happens if you pass in valid numbers in the wrong order?
eg: `Affine.new(17, 5)`

(This is probably not anything we need to care about here, I'm just wondering.)

exercises/affine-cipher/affine_cipher_test.rb
```diff
@@ -1,122 +1,149 @@
...
   end
 
   def test_decode_numbers
     skip
-    cipher = Affine.new
-    cipher.addkey(25, 7)
-    assert_equal 'testing123testing', cipher.decode('odpoz ub123 odpoz ub')
+    cipher = Affine.new(25, 7)
+    plaintext = 'testing123testing'
+    ciphertext = 'odpoz ub123 odpoz ub'
```
Insti: I wonder if these should be the other way around in the decode tests.
`ciphertext` first and `plaintext` second?

exercises/affine-cipher/.meta/generator/affine_cipher_case.rb
```diff
@@ -0,0 +1,51 @@
...
+        "cipher = Affine.new(#{key_a}, #{key_b})",
+        "plaintext = '#{expected}'",
+        "ciphertext = '#{input['phrase']}'",
+        "assert_equal plaintext, cipher.decode(ciphertext)"
+      ], 4
+    )
+  end
+
+  def error_workload
+    key_a, key_b = input['key']['a'], input['key']['b']
```
Insti: Pro tip: 
Each of these methods contains a `key_a, key_b = input['key']['a'], input['key']['b']` line.

I'd extract 2 methods that look something like:
```
def key_a
  input['key']['a']
end
```

This way it's clear 
a) what key_a is.
b) it's the same for all tests.
c) you're not repeating yourself all over the place.
d) if the key used in the canonical data ever changes (low probability, but possible) you only need to fix it in one place.

Insti:
Thanks @guygastineau, just the version change and I think this is ready to go.
I left some other comments but they are getting really pedantic at this point :)
I'd be happy to merge what we've already got. :heart:

exercises/affine-cipher/affine_cipher_test.rb
guygastineau: Well, key_b can essentially be any number whereas key_a must be coprime to 26.

Since it is technically fine for key_b to be coprime to 26 as well (just not required), I think there could be too many pitfalls to making the object determine which value should be key_a.

exercises/affine-cipher/affine_cipher_test.rb
guygastineau: I thought about doing this before. I might try it out.

exercises/affine-cipher/.meta/.version
guygastineau: Okay, so do I just change that by hand then, since the --update flag on `bin/generate` wants to go to 2 now?

Or is there some way to generate it to one in this case?

exercises/affine-cipher/affine_cipher_test.rb
guygastineau: In short 17, 5 is a valid key.  5, 17 is also a valid key, but they are not the same key and they will encoded to different values.

Insti: OK, thanks for the explanation.

guygastineau:
Thanks for that feedback @Insti 

Just sent up a new commit.

Insti:
Perfect! :heart: 
Thanks for all your work on this @guygastineau

guygastineau:
@Insti
Once again, thank you for helping me make this PR better.

I will continue building generators for exercises without one in my free time.

PR #842, opened 2018-08-21
generator: Report on broken generators

Insti:
When running `bin/generate --all` when a generator failed by raising an error it stopped the process. This change rescues the error, remembers the failure and continues on to the next generator.

At the end of the run a list of the broken generators is reported:
```
$ bin/generate --all
Generated affine-cipher tests version 1
...
Generated zipper tests version 71

Broken generators: (Run these individually to see the error raised.)

acronym
all-your-base
...
```

If a single generator is run the error will be raised and reported as normal:
```
$ bin/generate acronym
.../lib/generator/exercise_case.rb:27:in `method_missing': undefined local variable or method `phrase' for #<AcronymCase:0x00000000020edec0> (NameError)
...
```

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -9,4 +9,19 @@ paths = Generator::Paths.new(track: EXERCISM_RUBY_ROOT, metadata: METADATA_REPOS
 generators = Generator::CommandLine.new(paths).parse(ARGV)
 exit 1 unless generators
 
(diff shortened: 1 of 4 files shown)


Insti:
No objections, merged.

Insti:
It would be nice if `bin/generate` was tested.

PR #848, opened 2018-08-23
rna-transcription: Update tests to expect ArgumentError if invalid input is supplied

tylerferraro:
The Ruby rna transcription challenge has several tests that expect an empty string to be returned if invalid input is supplied. I feel it's better to expect an error/exception to be thrown as is done in other language implementations such as Python. It will help students build good habits around class and method design.

### DIFF
--- a/exercises/rna-transcription/.meta/generator/rna_transcription_case.rb
+++ b/exercises/rna-transcription/.meta/generator/rna_transcription_case.rb
@@ -2,20 +2,6 @@ require 'generator/exercise_case'
 
 class RnaTranscriptionCase < Generator::ExerciseCase
   def workload
-    if expects_error?
-      assert_raises(ArgumentError) { test_case }
-    else
-      assert_equal { test_case }
-    end
-  end
-
-  private
-
-  def test_case
     "assert_equal '#{expected}', Complement.of_dna('#{dna}')"
-  end
-
-  def expects_error?
-    expected.is_a? Hash
   end
 end


Insti:
Thanks for doing this @tylerferraro :heart:

The error case tests have been removed from the [canonical data](https://github.com/exercism/problem-specifications/commit/cb35c090d7d655e92bc5d86c1c16f1450f07c8bc#diff-3910f220e27365c92ddd7517cae5c9ad)

So this won't be an issue any more once the tests are regenerated.

tylerferraro:
@Insti Just to make sure I understand correctly. By removing the tests from the canonical data, when the tests are generated next time, these error test won't exist at all, correct? Wouldn't it benefit the student to think about edge cases such as a string with invalid characters?

Insti:
> By removing the tests from the canonical data, when the tests are generated next time, these error test won't exist at all. correct? 

This is correct.

> Wouldn't it benefit the student to think about edge cases such as a string with invalid characters?

Probably. But trying to do/test too much at the same time is also an issue.

There should be a dedicated string validation exercise for thinking about string validations.

If you see string validation issues in solutions you should mention it as part of the mentoring process but I don't think it should be part of the fixed test cases.

Insti:
See also: https://github.com/exercism/problem-specifications/issues/902

tylerferraro:
Now that this is approved, what's the process for merging?

Insti:
I usually wait a while to see if any of the other maintainers want to comment and/or agree and hit the merge button. If nothing happens, I merge it myself.

PR #850, opened 2018-08-28
Add Rubocop rake task

Insti:
Adds a Rubocop rake task.
It will run by default after the tests.

Also fix the things that Rubocop complained about.


Insti:
Yay thanks :heart:

PR #855, opened 2018-09-07
palindrome-products: Simplifying the exercise and removing need for a struct

zwarburg:
This exercise seemed overly complicated to me so I've simplified it to focus on the essentials.


Insti:
Thanks. I'm a bit busy right now and there is a lot going on here, but I'll look at this in a day or two. :heart:

(You could help me out by adding a comment here linking to the `README.md` for this problem and the `canonical-data.json` file which lives in the `exercism/problem-specifications` repository.)

Insti:
I agree that the problem is too complicated.

The project Euler problem that this is based on is much more sensible:
> A palindromic number reads the same both ways. The largest palindrome made from the product of two 2-digit numbers is 9009 = 91 × 99.
> Find the largest palindrome made from the product of two 3-digit numbers.

But project Euler only needs 1 number, and we need to have multiple input based tests.

Maybe we should just test for the largest number.
What about testing for a list of all the palindrome products? Or will there be too many?

It would probably be worthwhile creating an issue about this in the common https://github.com/exercism/problem-specifications repository, since this exercise is based on the common exercise that all Exercism tracks can implement.

Insti:
I'll close this, since no progress has been made for a while. Thanks for making it though :heart:
Feel free to re-open it if you want to continue working on it.

PR #871, opened 2018-10-05
change: Regenerate Tests

pgaspar:
Update `change` tests to: `1.3.0 258c807`

Update generator to use the standard error indicator.
Update solution to pass updated tests.

### DIFF
--- a/exercises/change/.meta/solutions/change.rb
+++ b/exercises/change/.meta/solutions/change.rb
@@ -1,6 +1,9 @@
 class Change
   attr_reader :coins, :target
 
+  class NegativeTargetError < ArgumentError; end
+  class ImpossibleCombinationError < StandardError; end
+
   def initialize(coins, target)
     @coins  = coins.sort.reverse
     @target = target
@@ -8,10 +11,11 @@ class Change
   end
 
   def generate
+    raise NegativeTargetError if target < 0
     return [] if target.zero?
 
     calculate_change(coins, [], target)
-    raise ArgumentError if total_change.none?
+    raise ImpossibleCombinationError if total_change.none?
 
     total_change.sort
   end
--- a/exercises/change/.meta/solutions/change.rb
+++ b/exercises/change/.meta/solutions/change.rb
@@ -11,8 +11,9 @@ class Change
     return [] if target.zero?
 
     calculate_change(coins, [], target)
+    raise ArgumentError if total_change.none?
 
-    total_change.any? ? total_change.sort : -1
+    total_change.sort
   end
 
   def self.generate(coins, target)
--- a/exercises/change/change_test.rb
+++ b/exercises/change/change_test.rb
@@ -45,21 +45,21 @@ class ChangeTest < Minitest::Test
 
   def test_error_testing_for_change_smaller_than_the_smallest_of_coins
     skip
-    assert_raises(ArgumentError) do
+    assert_raises(Change::ImpossibleCombinationError) do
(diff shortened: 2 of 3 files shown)


exercises/change/change_test.rb
```diff
@@ -45,16 +45,22 @@ def test_no_coins_make_0_change
 
   def test_error_testing_for_change_smaller_than_the_smallest_of_coins
     skip
-    assert_equal -1, Change.generate([5, 10], 3)
+    assert_raises(ArgumentError) do
+      Change.generate([5, 10], 3)
```
Insti: `ArgumentError` is probably OK here, but what about returning a subclass which is more descriptive?

exercises/change/change_test.rb
```diff
@@ -45,16 +45,22 @@ def test_no_coins_make_0_change
...
+    end
   end
 
   def test_error_if_no_combination_can_add_up_to_target
     skip
-    assert_equal -1, Change.generate([5, 10], 94)
+    assert_raises(ArgumentError) do
+      Change.generate([5, 10], 94)
```
Insti: `ArgumentError` is probably not appropriate here.
Maybe `nil` or a custom error?

exercises/change/change_test.rb
```diff
@@ -45,16 +45,22 @@ def test_no_coins_make_0_change
...
+    end
   end
 
   def test_cannot_find_negative_change_values
     skip
-    assert_equal -1, Change.generate([1, 2, 5], -5)
+    assert_raises(ArgumentError) do
+      Change.generate([1, 2, 5], -5)
```
Insti: `ArgumentError` is reasonable here.

(I don't think invalid input like this is something we should be testing for, but that's an unrelated issue. (and very much a matter of opinion.))

pgaspar:
Hello @iHiD! Thank you for your review. Wanted us to deal with #870 before this one, that's why I took so long to answer.

I understand and agree with your points, but I have a question:

Given how all 3 of these "error cases" are specified on the canonical data as `"expected": {"error": ...}`, how do you typically differentiate between them in the `change_case.rb` file? Should I check for the test name using something [like this](https://github.com/exercism/ruby/blob/master/exercises/alphametics/.meta/generator/alphametics_case.rb#L9)?

Insti:
I assume you meant @Insti there :)

Yeah you'll need to do something like that. Either check the test name or error text I guess.

iHiD:
Ah! I was confused :)

pgaspar:
Oh no! Sorry @iHiD 😅 Wrote that in a hurry and didn't notice it!

@Insti Thanks :+1: I'll work on something soon.

pgaspar:
@Insti I just pushed some tentative changes. I didn't follow your suggestions to the letter, but wanted to get your opinion regardless.

Please let me know what you think. I also appreciate suggestions for better error names :+1:

exercises/change/.meta/solutions/change.rb
```diff
@@ -1,18 +1,23 @@
 class Change
   attr_reader :coins, :target
 
+  class NegativeTargetError < ArgumentError; end
+  class ImpossibleCombinationError < StandardError; end
```
Insti: I really like these two :+1:

Insti:
Do you think this is ready to merge now @pgaspar ?

pgaspar:
I think so @Insti! Thank you 👍

Insti:
Thanks for all your work on these @pgaspar :heart:

PR #874, opened 2018-10-08
Ruby/Acronym : promote to core exercise

F3PiX:
This is one step in a process where I'm working on analyzing the exercises, in order to provide the best experience for both students and mentors. In the current order of the exercises, there's a gap between the very simple TwoFer exercise, and the next exercise Hamming, which is not so simple at all. So, I'm looking for exercises that bridge that gap, and also reordering the existing exercises. 

One question in the re-ordering is if adding extra core exercises would affect the queue and mentoring happiness. We're doing an experiment with another easy-to-mentor, early core exercise.

Acronym is our first choice, because:
- A _lot_ of the first submissions are very similar
- One snippet / precooked comment can cover 80% of new submissions, so it's supposed to be easy/fast to mentor
- It's a real step up from TwoFer, but not a huge jump in required Ruby skills.
- As a bonus: its Minimal Solution uses a method (`String#scan`) that's a real asset to be used in following exercises.

Updating Mentor Notes right now.

### DIFF
--- a/config.json
+++ b/config.json
@@ -424,9 +424,9 @@
     {
       "slug": "acronym",
       "uuid": "74468206-68a2-4efb-8caa-782634674c7f",
-      "core": false,
+      "core": true,
       "unlocked_by": "two-fer",
-      "difficulty": 2,
+      "difficulty": 1,
       "topics": [
         "regular_expressions",
         "strings",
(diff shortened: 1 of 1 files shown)


F3PiX:
Thanks @kotp!! It's live in the track.

Insti:
Awesome, thanks @F3PiX keep up the good work :+1:

PR #875, opened 2018-10-09
Make Acronym unlock WordCount

F3PiX:
... to check if Acronym's introduction to String#scan will change how many students will use `scan` in their first iteration in Word Count, and if that makes mentoring Word Count more interesting. 
  
Side effect: we're short one exercise being unlocked by TwoFer now that we take Word Count out.


Insti:
Thanks. I'm really loving your informative pull request descriptions :+1:

F3PiX:
Thanks @Insti !

PR #796, opened 2018-01-30
Update accumulate, close #795

matthew-campbell:
Update the advanced section of the accumulate readme to ensure the
correct enumerator returns when no block is given.

PR #876, opened 2018-10-10
Accumulate: Fix additional enum test in readme

stmllr:
The ["Advanced" section of the accumulate README](https://github.com/exercism/ruby/blob/5f9efd80026a70bb881657c3f1d8281a19158db4/exercises/accumulate/README.md#advanced) suggests to return enumerator if no block was given. It provides an optional test for that.

However the test seems broken, since it expects `number * number` calculation being returned by `each`.
It should rather use `map` to return the result:

Reference: #796 #395


Insti:
Awesome! thanks @stmllr :heart:

PR #879, opened 2018-10-16
Ruby/config Update core positions in track

F3PiX:
I totally misunderstood how the arrangement of the core exercises worked, so I never put Acronym early in the track as intended in #874 .
With this PR, Acronym goes to the right place on the track, just after TwoFer. 

And while I was at it, I placed all the core exercises at the top of the file, for easy reference.
As I finally realized that there never have been a rearrangement in Exercism V2, I changed the sequence of the other core exercise where their position was really weird. 
For instance: Isogram is now an early exercise instead of one second to last.
Matrix has moved to an earlier position et cetera. 

To keep the changes readable:
No core exercises changed to side or the other way around; no changes in the unlocking of the exercises.  (The latter will follow in a new PR soon, because currently Hamming is unlocking the majority of the sides. )

Note that this is rather a preparation for upcoming changes than the actual rearrangement. However, if you feel the sequence is off, it would be helpful for me to know.

### DIFF
--- a/config.json
+++ b/config.json
@@ -271,7 +271,6 @@
         "logic"
       ]
     },
-    
     {
       "slug": "word-count",
       "uuid": "e9d29769-8d4d-4159-8d6f-762db5339707",
--- a/config.json
+++ b/config.json
@@ -28,6 +28,44 @@
       ]
     },
     {
+      "slug": "acronym",
+      "uuid": "74468206-68a2-4efb-8caa-782634674c7f",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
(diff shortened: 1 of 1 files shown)


F3PiX:
@Insti  :-)
That's a dangerous assumption, seeing that I missed the point last time 😜 

Did you actually check this with configlet? Because I haven't work with that yet.

Insti:
`configlet lint` is run as part of the Travis CI checks: https://travis-ci.org/exercism/ruby/builds/442333055#L1826 So if that's green then the `config.json` file is at least well formatted.

F3PiX:
Aha! That's good to know. Thanks for enlightening me :-)

PR #883, opened 2018-10-20
A compact solution to alphametics.

n8chz:
@Insti [suggested](https://github.com/exercism/ruby/issues/849#issuecomment-431613956) that a pull request might be apropos.

### DIFF
--- a/exercises/alphametics/.meta/solutions/alphametics_compact.rb
+++ b/exercises/alphametics/.meta/solutions/alphametics_compact.rb
@@ -1,0 +1,30 @@
+class Alphametics
+  def self.solve(input)
+    input.upcase!
+    keys = input.gsub(/[^A-Z]/,"").chars.to_a.uniq
+    words = input.scan(/[A-Z]+/)
+    final_letters = words.map {|x| x.slice(-1)}
+    first_letters = words.map {|x| x.slice(0)}
+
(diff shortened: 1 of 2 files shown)


n8chz:
The FAKE NEWS test was proposed by me in an earlier PR. That was probably a bad move as it turns out to be "non-uniquely alphametic" (?) and the first solution my refined program finds is not the same one found by my original Exercism iteration.

When you say "To get the original version of the tests back" do you mean the original version of the tests before the unfortunate addition of the non-unique alphametism, or before the addition of the one with the leather roleplay with trolls and horses?

Insti:
The one that is the same as the current master in this repository.

https://github.com/exercism/ruby/blob/467ba57afac6653d67dc0a3ca4b23c999c147ce2/exercises/alphametics/alphametics_test.rb

Insti:
I'm not sure why the Github diff was showing that there were changes to the test file. But it all merged cleanly and now we have a faster alphametics solution. 

Thanks a lot @n8chz :heart:

n8chz:
Seems I'm also faster than the canonical solution to `kindergarten-garden` but only by a little, 0.011437s vs. 0.009183s. Most of these I suck at, though. :)

Insti:
How many times would the CI checks need to run to make it worth your time to make a PR for that? :smile:

PR #886, opened 2018-10-24
New exercise: High Scores

F3PiX:
This is part of the rearrangement of the track. 
One of the things missing is an easy exercise with an array as a list and some list manipulations.  
The Problem Specification PR is here:  https://github.com/exercism/problem-specifications/pull/1378

This High Scores exercise is meant to be a core exercise 
TwoFer -> Acronym ->  Isogram -> **High Scores**  -> Matrix -> ...

As we want to test drive, it's added here as a side exercise first. 
This exercise is also going to be the first core in the track where the students have to instantiate a class and work with multiple methods. We paid extra attention to the sequence of the tests, so the code is building up nicely, and at the end there should be some nice mentoring opportunities left.   

Thanks to @pgaspar, who took care of the implementation.  🙇

### DIFF
--- a/exercises/high-scores/.meta/generator/high_scores_case.rb
+++ b/exercises/high-scores/.meta/generator/high_scores_case.rb
@@ -1,0 +1,11 @@
+require 'generator/exercise_case'
+
+class HighScoresCase < Generator::ExerciseCase
+  def workload
+    [
+      "scores = #{scores}",
+      "expected = #{expected.inspect}",
+      "assert_equal expected, HighScores.new(scores).#{property}"
+    ]
+  end
+end
--- a/exercises/high-scores/high_scores_test.rb
+++ b/exercises/high-scores/high_scores_test.rb
@@ -1,0 +1,89 @@
+require 'minitest/autorun'
+require_relative 'high_scores'
+
+# Common test data version: 0.0.0
+class HighScoresTest < Minitest::Test
+  def test_list_of_scores
+    # skip
+    scores = [30, 50, 20, 70]
+    expected = [30, 50, 20, 70]
+    assert_equal expected, HighScores.new(scores).scores
+  end
+
+  def test_latest_score
+    skip
+    scores = [100, 0, 90, 30]
+    expected = 30
+    assert_equal expected, HighScores.new(scores).latest
+  end
+
+  def test_highest_score
+    skip
+    scores = [40, 100, 70]
+    expected = 100
+    assert_equal expected, HighScores.new(scores).highest
+  end
+
+  def test_personal_bests
+    skip
+    scores = [50, 30, 10]
+    expected = [50, 30, 10]
+    assert_equal expected, HighScores.new(scores).top
+  end
+
+  def test_personal_bests_highest_to_lowest
+    skip
+    scores = [20, 10, 30]
+    expected = [30, 20, 10]
+    assert_equal expected, HighScores.new(scores).top
+  end
+
+  def test_personal_bests_when_there_is_a_tie
+    skip
+    scores = [40, 20, 40, 30]
+    expected = [40, 40, 30]
+    assert_equal expected, HighScores.new(scores).top
+  end
+
+  def test_personal_bests_when_there_are_less_than_3
+    skip
+    scores = [30, 70]
+    expected = [70, 30]
+    assert_equal expected, HighScores.new(scores).top
+  end
+
+  def test_personal_bests_when_there_is_only_one
+    skip
+    scores = [40]
+    expected = [40]
+    assert_equal expected, HighScores.new(scores).top
+  end
+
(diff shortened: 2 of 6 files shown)


exercises/high-scores/.meta/generator/high_scores_case.rb
```diff
@@ -0,0 +1,11 @@
+require 'generator/exercise_case'
+
+class HighScoresCase < Generator::ExerciseCase
+  def workload
+    [
+      "scores = #{scores}",
+      "output = #{expected.inspect}",
```
Insti: I'd prefer that the variable here was `expected`

exercises/high-scores/high_scores_test.rb
```diff
@@ -0,0 +1,75 @@
+require 'minitest/autorun'
+require_relative 'high_scores'
+
+# Common test data version: 1.0.0 dd88548
```
Insti: Maybe the Common test data version should be `0.0.0` until this gets merged into problem-specifications.

exercises/high-scores/high_scores_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    scores = [4, 10, 7]
+    output = 10
+    assert_equal output, HighScores.new(scores).highest
+  end
+
+  def test_returns_personal_bests
+    skip
+    scores = [5, 3, 1]
+    output = [5, 3, 1]
+    assert_equal output, HighScores.new(scores).top
```
Insti: I like `top` with no arguments :+1:

Insti:
OK, I've left some feedback, most of it is fairly minor and this is certainly mergeable as is :star: 

Also see my comments in the https://github.com/exercism/problem-specifications/pull/1378 PR.

Do you want to get this PR merged ASAP or are you happy to wait a week or so for the problem-specifications PR to go through?

exercises/high-scores/.meta/generator/high_scores_case.rb
pgaspar: The reason why I went with `output` (also thought of `result`) was to make it easier to parse visually, since we have them in every spec:

```ruby
scores = [1, 2, 3, 4]
output = [1, 2, 3, 4]
result = [1, 2, 3, 4]
expected = [1, 2, 3, 4]
```

That being said, if you think the we're losing meaning by using those variable names, I'll change it. What do you think?

Insti: I'd rather choose variable names because they are meaningful rather than because they make the code look prettier.

You can always add a space after "scores" if you really care about lining things up.

"output" can be ambiguious, what is it the "output" of? In this case I understand it to be the the `expected_output` which I'd rather shorten to `expected`
 as it makes the `assert_equal` line make more sense.

The general format being:
`assert_equal expected, actual`

Then we have:

`assert_equal output, HighScores.new(scores).#{property}`
vs
`assert_equal expected, HighScores.new(scores).#{property}`

"expected" is also consistent with what most of the other exercises use in their tests.

(Although, this might not be a valid argument since I'm probably responsible for all of them being that way :) )

Insti: Just to be clear: I think you should change it.

pgaspar: Thank you for the detailed comments - I don't feel strongly about this and I'm typically not a proponent of making sure everything lines up because in "real life" it tends to be hard to maintain. 

This alone is enough reason for me:

> "expected" is also consistent with what most of the other exercises use in their tests.

Consistency++ I'll change it!

pgaspar:
@Insti heads up that I'm planning on squashing these commits before we merge.

F3PiX:
YAY!

F3PiX:
And _after_ merging I found this in the docs:
>Reference the PR in problem-specifications if it hasn't been merged yet, this *must not be merged* until the exercism/problem-specifications PR is merged.

Should I reverse the merge?  @Insti, do you know?

cc @pgaspar

pgaspar:
@F3PiX I saw that before too but forgot to ask @Insti about it.

From my limited testing we should be good. It just means that anyone trying to generate the tests for the whole track will see this at the end of the output:

```
Broken generators: (Run these individually to see the error raised.)

high-scores
```

It doesn't block the other generators from running and it doesn't change our existing tests. @Insti will know better though.

Insti:
@pgaspar that sounds about right.

I'm trying to remember if there was some other reason but I think It's mostly so the generator works correctly and will have the correct version information.

I think it's not going to be a problem in this case.

Insti:
Thanks for your work on this @F3PiX and @pgaspar :heart:

F3PiX:
Thanks for your support @Insti :-)) Much appreciated.

PR #888, opened 2018-10-27
Replacing refuse to assert on Isogram exercise test

OtavioHenrique:
Hi all, this is my first PR here so... be patient

I started at exercism today and I'm trying to do all ruby exercises to after this try became a mentor and help.

## Overview

When I was at isogram exercise I got stuck because my tests didn't pass and I found this `refute` at the test, and my question is: should be `refute` there?

https://github.com/exercism/ruby/blob/4aab415d1f12a6b1ef23d75de14151e3a2a4ee9e/exercises/isogram/isogram_test.rb#L24-L28

 I think the correct is `assert`, for example, with the given example, `zzyzx` the correct output should be `"Expected false, 'zzyzx' is not an isogram"`, but with the current test, It will not pass... Am I wrong?

### DIFF
--- a/exercises/isogram/isogram_test.rb
+++ b/exercises/isogram/isogram_test.rb
@@ -18,13 +18,13 @@ class IsogramTest < Minitest::Test
   def test_word_with_one_duplicated_character
     skip
     input = "eleven"
-    refute Isogram.isogram?(input), "Expected false, '#{input}' is not an isogram"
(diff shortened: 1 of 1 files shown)


Insti:
I guess you worked out you were mistaken. That's OK :heart:

Thanks for making a PR with an improvement and thanks for writing a good description of the issue and your proposed solution :star:

PR #889, opened 2018-10-31
refactor: pull romans up into constant

Oneill38:
Small refactor to pull the roman numerals into a constant since they will never change and don't need to live inside a method. If this were a running program it would also save a bit of memory since we wouldn't be instantiating a new hash every time the `ROMAN_MAPPINGS` are called


Insti:
Thanks :heart:

PR #893, opened 2018-10-31
Add Armstrong Numbers exercise

Tuxified:
[canonical-data](https://github.com/exercism/problem-specifications/tree/master/exercises/armstrong-numbers)

As this is my first stab at a Ruby exercise I didn't use the test generator, hope that's ok.

The solution is not using Integer.digits as that is available on Ruby 2.4, not previous ones.
Let me know what you think :)

TODO:
- [x] update config.json

As for the config .json metadata, difficulty 1 or 2, categories: integers, loops ?

### DIFF
--- a/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
+++ b/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
@@ -1,10 +1,7 @@
 class ArmstrongNumbers
   def self.is_valid?(number)
-    digits = number.to_s.chars.map(&:to_i)
-    exponent = digits.size
+    exponent = number.digits.size
 
-    digits.reduce(0) do |acc, digit|
-      digit**exponent + acc
-    end == number
+    number.digits.sum { |digit| digit**exponent } == number
   end
 end
--- a/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
+++ b/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
@@ -1,6 +1,6 @@
 class ArmstrongNumbers
   def self.is_valid?(number)
-    digits = number.to_s.split('').map(&:to_i)
+    digits = number.to_s.chars.map(&:to_i)
     exponent = digits.size
 
     digits.reduce(0) do |acc, digit|
--- a/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
+++ b/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
@@ -1,4 +1,4 @@
-class ArmstrongNumber
+class ArmstrongNumbers
   def self.is_valid?(number)
     digits = number.to_s.split('').map(&:to_i)
     exponent = digits.size
--- a/exercises/armstrong-numbers/armstrong_numbers_test.rb
+++ b/exercises/armstrong-numbers/armstrong_numbers_test.rb
@@ -1,0 +1,50 @@
+require 'minitest/autorun'
+require_relative 'armstrong_number'
+
+class ArmstrongNumbersTest < Minitest::Test
+  def test_one_digit_numbers_are_armstrong_numbers
+    assert ArmstrongNumber.is_valid?(1)
+    assert ArmstrongNumber.is_valid?(6)
+    assert ArmstrongNumber.is_valid?(9)
+  end
+
+  def test_two_digit_numbers_aren_t_armstrong_numbers
+    refute ArmstrongNumber.is_valid?(11)
+    refute ArmstrongNumber.is_valid?(89)
+    refute ArmstrongNumber.is_valid?(44)
+  end
+
+  def test_three_digit_numbers_that_are_the_armstrong_numbers
+    assert ArmstrongNumber.is_valid?(153)
+    assert ArmstrongNumber.is_valid?(370)
+  end
+
+  def test_three_digit_numbers_that_aren_t_the_armstrong_numbers
+    refute ArmstrongNumber.is_valid?(555)
+    refute ArmstrongNumber.is_valid?(662)
+  end
+
+  def test_four_digit_number_that_is_the_armstrong_numbers
+    assert ArmstrongNumber.is_valid?(1634)
+  end
+
+  def test_four_digit_number_that_is_not_the_armstrong_numbers
+    refute ArmstrongNumber.is_valid?(9989)
+  end
+
+  def test_seven_digit_number_that_is_the_armstrong_numbers
+    assert ArmstrongNumber.is_valid?(9926315)
+  end
+
+  def test_seven_digit_number_that_is_not_the_armstrong_numbers
+    refute ArmstrongNumber.is_valid?(2369989)
+  end
+
+  def test_ten_digit_number_that_is_the_armstrong_numbers
+    assert ArmstrongNumber.is_valid?(4679307774)
+  end
+
+  def test_fourteen_digit_number_that_is_the_armstrong_numbers
+    assert ArmstrongNumber.is_valid?(28116440335967)
+  end
+end
--- a/exercises/armstrong-numbers/armstrong_numbers_test.rb
+++ b/exercises/armstrong-numbers/armstrong_numbers_test.rb
@@ -1,50 +1,50 @@
 require 'minitest/autorun'
-require_relative 'armstrong_number'
+require_relative 'armstrong_numbers'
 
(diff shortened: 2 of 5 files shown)


Insti:
Great. thanks! 
@F3PiX can you give your estimation of difficulty and where this should fit in the tree?

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
```diff
@@ -0,0 +1,10 @@
+class ArmstrongNumbers
+  def self.is_valid?(number)
+    digits = number.to_s.split('').map(&:to_i)
```
Insti: Have you thought about using `chars` here?

exercises/armstrong-numbers/armstrong_numbers_test.rb
```diff
@@ -0,0 +1,50 @@
+require 'minitest/autorun'
+require_relative 'armstrong_numbers'
+
+class ArmstrongNumbersTest < Minitest::Test
+  def test_one_digit_numbers_are_armstrong_numbers
+    assert ArmstrongNumbers.is_valid?(1)
```
Insti: You'll need to add `skip`s to all the test methods (apart from the first) to be consistent with all the other tests.

F3PiX:
Hi @Tuxified , @insti! 

What I like most is that it's an math exercise that _doesn't_ ask to reimplement an existing Ruby method 👍  

I feel using `digits` and `sum` would add value regarding the Ruby idiom, and make it a bit more about Ruby than about maths.  
`number.digits.sum { |digit| digit**exponent } == number 
I don't think we should avoid Ruby 2.4 at all.  
(Ruby 2.5 is adding fun: `number.digits.sum { |digit| digit.pow(exponent) } == number` , but let's not go there.)
 
It's an easy exercise regarding the math, but the Enumerable makes it more of a level unlocked by Isogram. 
So:
Difficulty 3 seems fit, due to the Enumerable,  topic: math, unlocked by Isogram

Note
- I think I liked the singular class name better. Because it's _a_ check, for a single number, not a collection of sorts. ??

Thoughts?

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
```diff
@@ -0,0 +1,10 @@
+class ArmstrongNumbers
+  def self.is_valid?(number)
```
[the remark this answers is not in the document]
F3PiX: I like `self.valid?` better than `self.armstrong?`, although it's not really about validity, is it? I can't think of a better name though. 🤔

exercises/armstrong-numbers/armstrong_numbers_test.rb
```diff
@@ -0,0 +1,50 @@
...
+    assert ArmstrongNumbers.is_valid?(9)
+  end
+
+  def test_two_digit_numbers_aren_t_armstrong_numbers
+    refute ArmstrongNumbers.is_valid?(11)
+    refute ArmstrongNumbers.is_valid?(89)
+    refute ArmstrongNumbers.is_valid?(44)
+  end
+
+  def test_three_digit_numbers_that_are_the_armstrong_numbers
```
F3PiX: `test_three_digit_numbers_that_are_the_armstrong_numbers` Should this be `test_three_digit_numbers_that_are_armstrong_numbers` ?

Same below.

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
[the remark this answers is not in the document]
petertseng: In response to this comment I heavily considered suggesting `is?` because I thought it would be slightly amusing... but `ArmstrongNumbers.is?(10)` actually works in the opposite direction compared to `is_a?` so it would be unconventional.

If using `include?`, then `ArmstrongNumbers.include?(10)` *does* read naturally, but I think it would be surprising because it hints that you can use other Enumerable methods on ArmstrongNumbers (because `include?` is on Enumerable) but you cannot.

Insti:
> I think I liked the singular class name better. Because it's a check, for a single number, not a collection of sorts. ??

I really agree with your logic here.

However the convention is that the class name always matches the exercise slug,  so the `s` needs to be there.

See also: https://github.com/exercism/ruby/issues/854

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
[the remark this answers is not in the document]
Insti: I really like `include?` :+1:

F3PiX:
> However the convention is that the class name always matches the exercise slug, so the s needs to be there.

It's a new exercise, so we can change the slug? 🚀

F3PiX:
Hey @Tuxified What's the status?

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
Tuxified: Well spotted :)

exercises/armstrong-numbers/armstrong_numbers_test.rb
Tuxified: Done!

Tuxified:
Apologies it took too long to respond, wasn't my intention to neglect your efforts. I've updated to include your comments, however I'm still unsure about the name of method and/or implementing it as an Enumerable. 

Are you adamant on the Enumerable implementation or are you open for a different method name?

F3PiX:
Hey Toni, thanks for the next round!

Looks to me that there's some confusion in the Enumerable discussion.
Let me try to recap and see if we're all talking about the same:
- Candidates for name of the method:
  - `is_valid?` : let's follow the styleguide and don't prefix predicate methods with `is_...`
  - `include?` : does cover the meaning, but if I understand @petertseng 's comment correctly: as `include?` is also a method in Enumerable, using this name _suggests_ a) that `ArmstrongNumbers` responds to other Enumerable methods as well. It's not about using Enumerable methods in the implementation of the `method_to_be_named`. And b) that ArmstrongNumbers override the existing implementation of `Enumerable#include?`.
- `valid?` is also a bit misleading, because it's mostly use for validations, and that's not exactly what we're doing; a non-armstrong-number is not per se invalid, it's just not an armstrong number.
- I feel the `slug = class name` convention is making naming things even harder, but as we're stuck with it (for now), I got giggly and thought of: `really?` ?

Concluding:
- Let's avoid names that start with `is_`, and names that have a meaning in and/or are part of Ruby. 
- I was thinking: maybe use a synonym for the english 'include', like `contain?` , or more like `valid`: `confirm?` ,`affirm?`, or more like checking: `check?`,  `proven?`.

I'm okay of any of those, and I'd say let's just 'vote' for our preferences and let Tony decide. 
(- Unless, of course, my recap of the discussion is off. -)

@insti will be better at assessing the tests thing. All I know: there are exercises that use Ruby 2.4 and they work. I didn't look into that.

Insti:
The tests and solution need to work with Ruby 2.1.2 so you'll need to either "downgrade" the code or write a "polyfill" to support the new syntax. Or create a new issue/PR to upgrade the minimum version of ruby we support.

Insti:
@Tuxified 
For the method name, instead of `is_valid?` use `include?`.
(Don't worry about the `Enumerable` stuff.)

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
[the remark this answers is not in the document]
Insti: Including `Enumerable` is not really relevant to this PR but is something that could be interesting to explore as part of a solution.

guygastineau:
Is this abandoned? I am willing to take over implementation of this exercise.
It has been a long time since I contributed here, but I am coming back to `Ruby`, so I figured helping here will help me ramp back up :smile:

@Tuxified I don't want to step on your toes... If you aren't interested in finishing this I could build off of your work, so that you are still listed in the git history.

Tuxified:
Hi Guy,

No problem, I lost track of this PR, so if you're willing to pick it up, plz do. No need to keep this PR BTW, I don't care that much about keeping git history :P

guygastineau:
Cool. I'm on it.

guygastineau:
@Tuxified I went ahead and rebased your work onto the current master, so I could keep your commits in #988 .  I figured there was no reason to redo all the work from scratch, but I didn't want to take your work without credit :smiling_imp:

Tuxified:
Closing this, as #988 is a better choice :P

PR #897, opened 2018-12-12
ruby: make High Scores core, move Isogram up

F3PiX:
In the process of the Great Ruby Overhaul, we're promoting High Scores to a core exercise. And moving Isogram a bit further in the track, because there are too many ways to solve it for it to be at the beginning of the track.

### DIFF
--- a/config.json
+++ b/config.json
@@ -40,15 +40,13 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


F3PiX:
Thanks Victor!

PR #899, opened 2018-12-14
Redivide side exercises

F3PiX:
This is a next step in improving the unlocking of the side exercises.

Three goals for this step
- Make High Scores, as a new core exercise, unlock some sides
- Reduce the heap of sides unlocked by Hamming, and divide them more evenly
- Make the easiest exercises have really easy side exercises only
- I made it so that the last core exercises Tournament is the end of the track, and does not unlock any side exercises

Notes:
- [ ] The easiest exercises don't have enough side exercises, but I feel that's better than having too hard ones.
- [ ] Should the last core indeed be the closure of the track? 
- [ ] Luhn has too many side exercises unlock; this will be solved in a next iteration, where we'll add a few more advanced core exercises, and then we can redivide again.
- [ ] After Hamming, I moved some exercises up and down, but that's still quite arbitrary.

### DIFF
--- a/config.json
+++ b/config.json
@@ -213,7 +213,7 @@
       "slug": "rna-transcription",
       "uuid": "41f66d2a-1883-4a2c-875f-663c46fa88aa",
       "core": false,
-      "unlocked_by": "hamming",
+      "unlocked_by": "two-fer",
       "difficulty": 2,
       "topics": [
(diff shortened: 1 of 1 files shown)


F3PiX:
Thank you @insti 😍

PR #916, opened 2019-01-19
series: Regenerate README.md with new hints

### LINKED ISSUE #913: Generate things
@pgaspar Thanks for helping 💙 

This things are ready to be generated:
- [x] Series Readme (to include the hints.md) but NOT the canonical data tests. #911 
- [x] OCR: Readme (to include the hints.md) #901
- [ ] ~High Scores adjustments?~

pgaspar:
Regenerates the Readme file to include the new hints introduced on #911. Cleaned up some whitespace and moved the file to the correct directory first.

Resolves #913.


F3PiX:
@pgaspar A question for the sake of future updates of  `.md` files. 
This is how the hints.md look on the site:
  
schermafbeelding 2019-01-21 om 10 14 22

Usually I use double spaces at the end of the lines, which creates a line break, but maintainers hate those :-) What's the best way to format?

pgaspar:
Oh, right. To add a blank line between paragraphs you need to do 2 line breaks. Like this:

```markdown
First paragraph spanning
multiple lines.

Second paragraph!
```

Sorry I missed this. I'll change that.

F3PiX:
> To add a blank line between paragraphs you need to do 2 line breaks.

Yeah, I know, but as the double spaces do the same, is it worth the extra effort to change them (afterwards) into double line breaks? 🤔 Or are the double spaces not rendered correctly on the site?

pgaspar:
My experience is that a lot of people have configured their editors to automatically remove trailing whitespace, which I have done as well.

I think this is for multiple reasons - the first being that it's hard to track where extra spaces are and wether they were added intentionally or not. For example, I just assumed your whitespaces were a byproduct of some copy pasting and/or editing.

Other reasons are that extra spaces can create some git noise, or that they may hide some hard to find bugs.

Specifically on Exercism, only `affine-cipher`, `list-ops`, `ocr-numbers` and `two-fer` have trailing whitespace in their description files on [exercism/problem-specifications](https://github.com/exercism/problem-specifications). (`grep exercises/**/description.md -e " $"`)

For all these reasons I would appreciate it if we followed the no trailing whitespace guideline to keep everything consistent :+1: At least in places where we can easily replace them with alternatives. (PS: it's also on the [rubocop styleguide](https://github.com/rubocop-hq/ruby-style-guide#no-trailing-whitespace) although without any explanation).

F3PiX:
> For all these reasons I would appreciate it if we followed the no trailing whitespace guideline to keep everything consistent 👍 

I agree.

PR #918, opened 2019-01-20
Promote Word Count to core, demote math exercises

F3PiX:
Another step in the Ruby Track Overhaul! 💎

* Most important updates:

   - Promote Word Count to core: between Series and Hamming
  - Demote Grains and Difference of Squares to side (due to a decision to [make all math exercises side exercises](https://github.com/exercism/exercism/issues/4110) in all tracks. This wasn't implemented in Ruby yet, because we had no substitutes ready.)
 
  To make it easier to redivide side exercises, I combine promoting and demoting exercises at the same time. 

* Side effects: I had to redivide the side exercises.
  - the ones unlocked by Grains and Difference of Squares are now divided over other core exercises
  - Word Count needed a few side exercises, I stole them from other core exercises on roughly the same level
  - And then I had to redivide some others, to balance the number of sides for each core.

The changes for the side exercises are quite arbitrary, and will change all the time during upcoming changes.
The changes to the core exercises are for real! 🎉

---
Don't merge before https://github.com/exercism/website-copy/pull/709

### DIFF
--- a/config.json
+++ b/config.json
@@ -213,7 +213,7 @@
     {
       "slug": "grains",
       "uuid": "22519f53-4516-43bc-915e-07d58e48f617",
-      "core": true,
+      "core": false,
       "unlocked_by": "series",
       "difficulty": 4,
       "topics": [
--- a/config.json
+++ b/config.json
@@ -76,6 +76,18 @@
       ]
     },
     {
+      "slug": "word-count",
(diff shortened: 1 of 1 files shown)


F3PiX:
Thanks both of you! I'll merge this as soon as the [Mentor Notes](exercism/website-copy#709) are live!

PR #929, opened 2019-01-26
Use squiqqly heredoc in tests

ccouzens:
Previously it used regular heredocs and removed the indentation using
regular expressions.
I think it is now easier to read and understand.

~A side-effect of this change is to upgrade the project to Ruby 2.3.
This is the version that introduced squiqqly heredocs.
I think it's reasonable to upgrade, as Ruby 2.3 first came out over 3
years ago (2015-12-25).~

Update: Ruby has been updated to 2.5 in #933

### DIFF
--- a/exercises/beer-song/.meta/generator/beer_song_case.rb
+++ b/exercises/beer-song/.meta/generator/beer_song_case.rb
@@ -3,7 +3,7 @@ require 'generator/exercise_case'
 class BeerSongCase < Generator::ExerciseCase
   def workload
     [
-      "expected = #{indent_heredoc(expected, 'TEXT', 0, ".gsub(/^ */, '')" )}\n",
+      "expected = #{indent_heredoc(expected, 'TEXT' )}\n",
       "assert_equal expected, BeerSong.recite(#{start_bottles}, #{take_down})\n"
     ].join
   end
--- a/exercises/beer-song/beer_song_test.rb
+++ b/exercises/beer-song/beer_song_test.rb
@@ -5,7 +5,7 @@ require_relative 'beer_song'
 class BeerSongTest < Minitest::Test
   def test_first_generic_verse
     # skip
-    expected = <<-TEXT.gsub(/^ */, '')
+    expected = <<~TEXT
     99 bottles of beer on the wall, 99 bottles of beer.
     Take one down and pass it around, 98 bottles of beer on the wall.
     TEXT
@@ -14,7 +14,7 @@ class BeerSongTest < Minitest::Test
 
   def test_last_generic_verse
     skip
-    expected = <<-TEXT.gsub(/^ */, '')
+    expected = <<~TEXT
     3 bottles of beer on the wall, 3 bottles of beer.
     Take one down and pass it around, 2 bottles of beer on the wall.
     TEXT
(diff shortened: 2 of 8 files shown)


ccouzens:
I think if it was me reviewing this pull request, I might ask for it to be split into 2 PRs:
1. Upgrade to Ruby 2.3
2. Switch to squiggly heredocs

I'd be happy to split if it gets suggested

ccouzens:
Hi @F3PiX ,

I've opened an issue about updating Ruby. https://github.com/exercism/ruby/issues/930

I don't think upgrading should affect too much. The tests pass on this branch, `rubocop` passes. There are some differences when I regenerate the tests, but I assume that's because some of the tests are stale.

Kind regards,

Chris

ccouzens:
By more verbose tests do you mean something like:

```ruby
      input = "ABC\n123"
```

instead of the current 

```ruby
    input = <<-INPUT.gsub(/^ {6}/, '').strip
      ABC
      123
    INPUT
```
?

F3PiX:
@ccouzens 
>By more verbose tests do you mean something like:
   input = "ABC\n123"
Exactly :-) 

And I'd be doubly happy because that means we can handle the Ruby update separately :-)

pgaspar:
I like the idea behind these changes a lot! I noticed we were using non-squiggly heredocs before but had not taken the time to look into it.

I agree these should be 2 separate PRs: one for the Ruby 2.3 update and another for the heredoc changes (which depends on the first PR).

> I was just thinking the other day that the heredocs in the tests for this exercise are not a good choice, because we're introducing heredocs later in the track than this exercise appears.
Because of this, I'd prefer much more verbose tests, so that (beginner) students can read the tests fairly easy.

@F3PiX I assume you're talking specifically about Transpose, right?

I think this is a separate issue - these PRs improve the status-quo and make tests simpler to understand. As for Transpose, if we want to simplify the tests, that's something that should to be dealt with on its own - for example by not using `indent_heredoc` on transpose's generator, which is what generates the heredocs in the test files.

kotp:
Thank you for this, I had started to work on the same problem (The heredocs not having squiggly) last week, did not have time to finish it up.)

I don't mind having Heredocs in the tests before they are formally introduced in the exercises.  The tests are there to confirm, not necessarily to teach (though they should be idiomatic, and "perfect" examples of tests because they are delivered to the students).  If they are not understood (the heredocs) that is OK, they are exposure.  By the time the Heredocs are introduced, the students have at least seen them, even if they have not be directed to analyze them.

Ruby 2.3 is now under security maintenance phase, until the end of the March of 2019, and so I would be OK if this is moved up even one more minor version...

F3PiX:
> @F3PiX I assume you're talking specifically about Transpose, right?

For removing the heredocs from the tests, yes. I made Transpose to be unlocked by Matrix, and I don't want heredocs at that point in the track. In Beer Song and Tournament it seems appropriate. 

I like the squiqqly's a lot in heredocs in general, and in Tournament

And I'm worried about auto-generating tests. See my remarks in #930

ccouzens:
Now that the ruby upgrade has been separately merged, I've rebased this branch to not include it.

@F3PiX can you tell me more about how the tests are autogenerated?
Is it a different process to running `./bin/generate`?

F3PiX:
@ccouzens I'm so sorry, but we decided to not merge PR's that change exercises at the moment. We're currently going through a huge restructurering of the Ruby track, and it's very hard to keep that project on track when exercises change in the meantime. 
In a few weeks, there will be more clearity on things. And we will be in a better place to discuss what can and can not be changed, and we'll do our best to update the communications for contributors in the meantime. 
I sincerely hope to see you back in a while.

F3PiX:
@ccouzens The future is now! Sorry for the delay, I'm going to merge this finally. 
Thank you very much! 💙 

cc @pgaspar

ccouzens:
Thank you

PR #933, opened 2019-02-02
Upgrade to ruby 2.5 and rubocop 0.63

ccouzens:
https://github.com/exercism/ruby/issues/930

I'd prefer to only upgrade ruby, but doing so breaks rubocop.
Upgrading rubocop on its own also isn't an option as it requires newer versions of ruby.

`./bin/compile --all`'s output doesn't change and the tests are passing.

Rubocop needed quite substantial changes. I'd be interested how the original .rubocop.yml was generated so that I can make my rubocop change less drastic.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,487 @@
 AllCops:
   TargetRubyVersion: 2.3
 
+# Supports --auto-correct
+Layout/AccessModifierIndentation:
+  Description: Check indentation of private/protected visibility modifiers.
+  StyleGuide: https://github.com/bbatsov/ruby-style-guide#indent-public-private-protected
+  Enabled: false
+  EnforcedStyle: indent
+  SupportedStyles:
+  - outdent
+  - indent
+  IndentationWidth: 
+
+# Supports --auto-correct
+Layout/AlignArray:
+  Description: Align the elements of an array literal if they span more than one line.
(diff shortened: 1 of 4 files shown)


kotp:
> I'd be interested how the original .rubocop.yml was generated 

`rubocop --auto-gen-config` can help you in this regard, I think.

.rubocop_todo.yml
```diff
@@ -0,0 +1,235 @@
+# This configuration was generated by
```
kotp: This file should probably not be here... as in not submitted to version control.

.rubocop_todo.yml
```diff
@@ -0,0 +1,235 @@
...
+# `rubocop --auto-gen-config --no-auto-gen-timestamp`
```
kotp: Though, this is the command that I likely used to get some idea of what styles were being used...

kotp:
I have no idea what the `mry` command is.

ccouzens:
@kotp thanks for the approval.

I didn't know what `mry` was either until yesterday.

https://github.com/rubocop-hq/rubocop/blob/master/manual/configuration.md#updating-the-configuration-file

https://github.com/pocke/mry

The names of some of the cops and the names of their configuration changes between versions of rubocop. `mry` will automatically update the `.rubocop.yml` to replace the old names with the new names. It's not perfect, but it did save a lot of time.

PR #936, opened 2019-02-10
high-scores: Regenerate tests

pgaspar:
Update tests to: ~`3.0.0 999ec47`~ `4.0.0 ad1f9c4`

These are the changes introduced by:
* https://github.com/exercism/problem-specifications/pull/1447
* https://github.com/exercism/problem-specifications/pull/1459

### v3.0.0 notes

Right now I kept the new method name `personal_top_three`, but a case could be made for naming it `personal_top_3` or rather `personal_top3`, as the [rubocop styleguide suggests](https://github.com/rubocop-hq/ruby-style-guide#snake-case-symbols-methods-vars-with-numbers). Personally, I'm not a fan of `personal_top3` 🤔 Thoughts are welcome.

@F3PiX I believe we can make this change without affecting the overhauling efforts, but let me know if that's not the case.

### v4.0.0 notes

Also updated README.md :+1:

### TODO

- [x] Update changelog on the mentor notes as soon as this is merged (https://github.com/exercism/website-copy/pull/783)
- [x] Upload mentor notes to remove references to the `report` method.

### DIFF
--- a/exercises/high-scores/.meta/solutions/high_scores.rb
+++ b/exercises/high-scores/.meta/solutions/high_scores.rb
@@ -15,8 +15,8 @@ class HighScores
     scores.last
   end
 
-  def personal_top
-    scores.sort.reverse.take(3)
+  def personal_top_three
+    scores.max(3)
   end
 
   def report
--- a/exercises/high-scores/.meta/solutions/high_scores.rb
+++ b/exercises/high-scores/.meta/solutions/high_scores.rb
@@ -18,10 +18,4 @@ class HighScores
   def personal_top_three
     scores.max(3)
   end
-
-  def report
-    difference = "#{personal_best - latest} short of" if personal_best != latest
-
-    "Your latest score was #{latest}. That's #{difference} your personal best!".squeeze
-  end
 end
--- a/exercises/high-scores/README.md
+++ b/exercises/high-scores/README.md
@@ -2,7 +2,7 @@
 
 Manage a game player's High Score list.
 
-Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write methods that return the highest score from the list, the last added score, the three highest scores, and a report on the difference between the last and the highest scores.
+Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write methods that return the highest score from the list, the last added score and the three highest scores.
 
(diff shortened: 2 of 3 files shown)


exercises/high-scores/.meta/solutions/high_scores.rb
```diff
@@ -15,13 +15,7 @@ def latest
...
-  def personal_top
-    scores.sort.reverse.take(3)
-  end
-
-  def report
-    difference = "#{personal_best - latest} short of" if personal_best != latest
-
-    "Your latest score was #{latest}. That's #{difference} your personal best!".squeeze
+  def personal_top_three
+    scores.max(3)
```
F3PiX: I _think_ I'd prefer to have the 'minimal solution for approval', not the 'best'.
(Because it seems that some mentors refer to the example, or at least look at it.)

pgaspar: This shouldn't be too relevant 😕 These solutions are not for mentor consumption - and even if a mentor does see this I don't see it as a big issue at all. I'd rather keep things fairly separate - mentor notes are one thing, these solutions are another.

What do you think? Do you feel strongly about this? I can revert the change, but I don't want us to start changing these solutions every time we change the mentor notes - unless that's something we decide to do across all tracks.

pgaspar:
> I wonder: should we wait with merging until the Mentor Note update is merged?

Let's try to merge them at the same time.

exercises/high-scores/.meta/solutions/high_scores.rb
F3PiX: Not _very_ strongly, but let's say: preferrable. Strangely exactly because I agree with you: that we shouldn't start changing these solutions. Unless it's a bad example/ bad practice / outdated practice.  
AFAIK the current line is none of those,  or is it?

pgaspar: It's not. I'll change it back.

pgaspar: But really, the only reason I didn't commit `max(3)` the first time around is because I didn't know about it 😅

pgaspar: > It's not. I'll change it back.

Actually, `scores.sort.reverse.take(3)` does do a bit of extra work, when compared to `max(3)`. 

Also, my point about not starting to change these solutions was specifically about changing them every time we change the mentor notes. I think we should be free to change these as we please, without the mentor notes having much influence. My main point is that (at least for now) they are different things and should change independently, if needed.

F3PiX: Ooooh, this was absolutely not meant to be blocking merging this PR! I'm sorry.

PR #939, opened 2019-02-24
resistor-colors: New exercise implementation

### LINKED ISSUE #938: Implement New Resistor exercise: ResistorSignificants
In the Resistor Series exercises, [#2](https://github.com/exercism/problem-specifications/pull/1466) is being launched. 

This is a nice simple exercise in Ruby. It can be solved with a simple `map`, and it's just a level above what tutorials offer re: simple array operations. (The first in the series is too simple in Ruby.)

In the Track Anatomy Project, we discussed creating more diversity in the kind of exercises tracks offer. In Ruby, we can add some scripting exercises. Resistor #2 is a great candidate. 

Can we implement this as a scripting exercise? And who would be up to that?

pgaspar:
Closes #938. The goal was to have something that passes the tests from https://github.com/exercism/problem-specifications/pull/1466 so we can start the conversation.

TODO:
- [x] Add to config.json

### DIFF
--- a/exercises/resistor-colors/README.md
+++ b/exercises/resistor-colors/README.md
@@ -1,0 +1,52 @@
+# Resistor Colors
+
+If you want to build something using a Raspberry Pi, you'll probably use _resistors_. For this exercise, you need to know two things about them:
+
+* Each resistor has a resistance value.
+* Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
+To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
+
+In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take two colors as input, and output the correct number.
+
+The band colors are encoded as follows:
+
+- Black: 0
+- Brown: 1
+- Red: 2
+- Orange: 3
+- Yellow: 4
+- Green: 5
+- Blue: 6
+- Violet: 7
+- Grey: 8
+- White: 9
+
+* * * *
+
+For installation and learning resources, refer to the
+[Ruby resources page](http://exercism.io/languages/ruby/resources).
+
+For running the tests provided, you will need the Minitest gem. Open a
+terminal window and run the following command to install minitest:
+
+    gem install minitest
+
+If you would like color output, you can `require 'minitest/pride'` in
+the test file, or note the alternative instruction, below, for running
+the test file.
+
+Run the tests from the exercise directory using the following command:
+
+    ruby resistor_colors_test.rb
+
+To include color from the command line:
+
+    ruby -r minitest/pride resistor_colors_test.rb
+
+
+## Source
+
+Maud de Vries, Erik Schierboom [https://github.com/exercism/problem-specifications/issues/1464](https://github.com/exercism/problem-specifications/issues/1464)
+
+## Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/config.json
+++ b/config.json
@@ -221,6 +221,16 @@
         "math"
       ]
     },
+    { "slug": "resistor-colors",
+      "uuid": "57b0c57e-26a5-4ccf-aaf2-2fefddd918c1",
+      "core": false,
+      "unlocked_by": "acronym",
+      "difficulty": 2,
+      "topics": [
+        "array",
+        "loops"
+      ]
+    },
     {
       "slug": "rna-transcription",
       "uuid": "41f66d2a-1883-4a2c-875f-663c46fa88aa",
--- a/config.json
+++ b/config.json
@@ -224,7 +224,7 @@
     { "slug": "resistor-colors",
       "uuid": "57b0c57e-26a5-4ccf-aaf2-2fefddd918c1",
       "core": false,
-      "unlocked_by": "acronym",
+      "unlocked_by": null,
       "difficulty": 2,
       "topics": [
         "array",
--- a/exercises/resistor-colors/.meta/generator/resistor_colors_case.rb
+++ b/exercises/resistor-colors/.meta/generator/resistor_colors_case.rb
@@ -1,0 +1,7 @@
+require 'generator/exercise_case'
+
+class ResistorColorsCase < Generator::ExerciseCase
+  def workload
+    assert_equal(expected, "ResistorColors.value(#{colors})")
+  end
+end
--- a/exercises/resistor-colors/.meta/solutions/resistor_colors.rb
+++ b/exercises/resistor-colors/.meta/solutions/resistor_colors.rb
@@ -1,0 +1,21 @@
+module ResistorColors
+  COLORS = %w(
+    black
+    brown
+    red
+    orange
+    yellow
+    green
+    blue
+    violet
+    grey
+    white
+  ).freeze
+
+  def self.value(color_bands)
+    color_bands
+      .map { |color| COLORS.index(color) }
+      .join
+      .to_i
+  end
+end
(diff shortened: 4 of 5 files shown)


F3PiX:
Aha cool! That helps!

This was supposed  to be an array exercise, making use of the fun fact that the colours are coded 0..9, so we can use the index in the array to find the code. 
Ideally, the exercise would steer into the direction of something like:

```
color_bands.map {|color| COLORS.index(color) }.join.to_i
```
Can we make that happen?

pgaspar:
I assume `color_bands` is the input? If so you should suggest changing the problem specification so it defines the method input as an array and not two separate attributes.

> Can we make that happen?

Other than that, I can change the solution to be more in line with your snippet. If the input doesn't change to an array, though, I expect many students to prefer something along the lines of:

```ruby
"#{COLORS.index(color1)}#{COLORS.index(color2))".to_i
```

vs.

```ruby
[color1, color2].map {|color| COLORS.index(color) }.join.to_i
```

This last solution does a lot more work, specially for something that will only ever receive 2 colors (right?). It makes an array out of the two color variables, maps through it, joins the result creating a string, converts the string back to integer. The first approach makes a string using interpolation and converts it to integer.

F3PiX:
The input _is_ going to change to an array. To be continued.

exercises/resistor-significants/.meta/solutions/resistor_significants.rb
```diff
@@ -0,0 +1,21 @@
...
+    brown
+    red
+    orange
+    yellow
+    green
+    blue
+    violet
+    grey
+    white
+  )
```
F3PiX: Let's freeze this.

F3PiX:
> This [map] solution does a lot more work, specially for something that will only ever receive 2 colors (right?). It makes an array out of the two color variables, maps through it, joins the result creating a string, converts the string back to integer. The first approach makes a string using interpolation and converts it to integer.

- A resistor can have more than 2 bands (1 upto 6), and students can be well aware of that fact through the description. But, in this exercise it is always about 2 bands, yes.
- I'd think that the `map` solution is better: more idiomatic, more open/close, better readability, better maintainability. And given that we don't care about optimization at this point in the track. 
Is it true then that the map solution is better? 
- If ^ is  true, we can go two ways: when students create a solution with string interpolation, the mentoring can point to `map`. Or, we just accept the string interpolation as 'minimal solution for approval', and use the next exercise in the series, that adds the third band, to illustrate the usefulness of een open/close solution. 
- If ^^ is not true, we just go with the string interpolation.

pgaspar:
> * I'd think that the `map` solution is better: more idiomatic, more open/close, better readability, better maintainability. And given that we don't care about optimization at this point in the track.
  Is it true then that the map solution is better?

I think this is all a bit arguable. If the input is going to be an array from the get go I'd do something like this, just because I prefer mapping over accessing array elements directly using `[0]` or `first`. RE open/closed, what scenario are you thinking about specifically?

So, to clarify, I agree with changing the solution when the input changes on the problem specification now that the input is an array. But I still have some doubts regarding making it about open/closed - a good example of how extension and modification would look like here would help!

pgaspar:
The exercise has been renamed from `resistor-significants` to `resistor-colors`. I'm updating the PR to reflect that.

pgaspar:
The build is failing because `config.json` was not updated. @F3PiX if you can add a commit with that change I'd appreciate it!

pgaspar:
Squashed everything into a single commit. Squashed commits:

- resistor-significants: Initial approach
- resistor-significants: map approach
- resistor-significants: Array input
- resistor-colors: New exercise

pgaspar:
Looks good to me @F3PiX. Do you want someone else to review this?

F3PiX:
Thanks Pedro :bow: I added the config.json.

As stated in the OP, we should discuss a bit more how we want to use the exercise in the track. 

- The most important thing: avoid that the student has to decide on the data structure. I'd prefer to give them the COLORS array, by:
  - Adding a hint to the Readme, where they can copy/paste the COLORS array 
  OR  
  - Adding a stub file with the COLORS array in it. Discussed [here](https://github.com/exercism/ruby/issues/541). I'm not a fan of adding stub files everywhere, but in this case, giving them the data structure for the Colors is a big pro to me.   

The other thing on my wish list: add some script exercises. This could be a candidate for the first,  simple one.  Using `gets.chomp` is part of most beginner tutorials I know of, but `ARGV` would be more interesting, but also levels it up, and in that case the logic may be a bit too simple. 

Thoughts?

pgaspar:
> The most important thing: avoid that the student has to decide on the data structure. I'd prefer to give them the COLORS array

I'm not entirely sure why we want to force people to follow this specific approach so much. I thought Exercism was not about prescribing the single correct solution but allowing for experimentation and leveraging the fact that we have real mentors to explore the pros and cons of other approaches.

> The other thing on my wish list: add some script exercises. This could be a candidate for the first, simple one. Using gets.chomp is part of most beginner tutorials I know of, but ARGV would be more interesting, but also levels it up, and in that case the logic may be a bit too simple.

I shared some of my thought regarding this on Slack, but I'll share them here too for others to read: This seems to deviate a bit from the test-first approach. Throughout the track we try to get people to run their code through their tests, not by calling it via terminal. In `TwoFer` it's common to see people use `puts` and `p` and to have `TwoFer.two_fer("Some String")` at the end of the file. That is usually indicative that they’re not running the tests but rather manually debugging through their console / IDE.

Another point is that `gets.chomp` is hard to test. Here are [some examples](https://stackoverflow.com/questions/16948645/how-do-i-test-a-function-with-gets-chomp-in-it) of what people do. This by itself forces the code to have a certain structure which may be confusing for beginners.

I'm skeptical about the need to have script exercises in Exercism. I'd rather if we continued focusing on running exercises via the tests, which is a good habit to interiorize from early on IMO.

Curious about what others think.

F3PiX:
@pgaspar My conclusion for now is: let's merge it as is, and I'll monitor what happens.  
 
re:  the scripting exercises: they shouldn't be non-TDD. And I totally agree that an exercise at this level should have easy to read tests! Good point.

exercises/resistor-colors/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+* Each resistor has a resistance value.
+* Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
+To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
+
+In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take two colors as input, and output the correct number.
+
+The band colors are encoded as follows:
+
+- Black: 0
```
kotp: Traditionally, these bands are expressed as starting with value 1.  It fits the mnemonic that we use to remember the values.  Since this describes the problem domain, there is no reason to order them as we order things in the programming domain (often using 0 as the starting point).

F3PiX: That's not what I found.

For quick reference:
https://en.wikipedia.org/wiki/List_of_electronic_color_code_mnemonics
All these mnemonics starts with BB, for Black and Brown.

In Dutch it is `**Z**ij Bracht Rozen op Gerrits Graf bij view grauw weer`
Z for Zwart = Black. `weer` = Wit = White

The fun is: it's one of the rare real life cases where the index 0 _does_ have meaning, and that was the inspiration for the exercise.  :joy:

kotp: Well, it has been quite a few decades. :)  And I am going off of my rusty memory.  Good to go then.

F3PiX:
Thanks all, and especially @pgaspar for delivering _yet another_ nice addition to the track! 

🍾

pgaspar:
@kotp I believe merging is blocked until you ✅ the PR.

kotp:
I noted that it is currently not core and unlocked by null.

F3PiX:
Thanks Victor!

PR #947, opened 2019-03-16
Added note for hexadecimal exercise readme

EvgeniyRRU:
When I viewed Community Solutions for hexadecimal exercise, I detected that, according to Readme, one can simply using Ruby standard library method `to_i(16)`. I think we should to fix it

### DIFF
--- a/exercises/hexadecimal/README.md
+++ b/exercises/hexadecimal/README.md
@@ -7,6 +7,11 @@ teal: 008080, navy: 000080).
 
 The program should handle invalid hexadecimal strings.
 
+## Note
+
+- Implement the conversion yourself.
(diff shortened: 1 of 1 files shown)


EvgeniyRRU:
@F3PiX I don't understand you. There is not `hits.md` file in this exercise. I've just copied this block from 'binary' exercise

petertseng:
EBWODP: I will make clear that I mean to say the existing readme **already** says what I claimed it says, by changing the following:

> this very same file says 

this very same file *already* says

EvgeniyRRU:
@petertseng Oh, sorry, I read it carefully, so I decline this PR.

PR #948, opened 2019-03-25
grep: New exercise

### LINKED ISSUE #946: New advanced exercise: Add Grep
The start of the Ruby Track got a lot of attention in the previous months. It's time to give the end of the track a bit of Tender Loving Care as well. After Luhn and Clock, there's a lot more advanced features to explore. 
One thing that is missing, is Importing and reading a file. 

Problem Specifications has Grep: https://github.com/exercism/problem-specifications/blob/5326cf2b32fd5df77cda3ade20d088d9a928b8b3/exercises/grep/canonical-data.json

Looks like a good candidate to add as a side, and maybe to be promoted to core later on.

Anybody feel like implementing it for Ruby?

pgaspar:
Closes #946.

### Implementation

The implementation was heavily influenced by [python's](https://github.com/exercism/python/blob/a83f53697a1a213967838a78b86f2b358727d76f/exercises/grep/example.py#L45). It can certainly be improved, but we may want to leave that for the mentor notes themselves, since the main goal here is to get CI to pass.

### Test Generation

The tests are generated using the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/grep/canonical-data.json), but the file contents are hardcoded in a custom test template. I ended up doing this because I couldn't easily pass custom variables from the test case file to the ERB test template. Without that I can't parse the canonical data comments and make the file contents dynamic, like some tracks do.

I couldn't find other exercises that did anything close to this and looking at the generator code I didn't find a clear way to pass a `file_data` hash to the template file.

~Note: I really want to merge #929... Having the generator write `<<-EXPECTED.gsub(/^ */, '').strip` instead of `<<~EXPECTED` on the test file adds a layer of extra complexity we could easily avoid.~ We've merged this and I updated the code 👍.

@exercism/ruby 

**TODO**

- [x] Add to `config.json`
- [ ] Add issue to create mentor notes

### DIFF
--- a/exercises/grep/README.md
+++ b/exercises/grep/README.md
@@ -1,0 +1,95 @@
+# Grep
+
+Search a file for lines matching a regular expression pattern. Return the line
+number and contents of each matching line.
+
+The Unix [`grep`](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) command can be used to search for lines in one or more files
+that match a user-provided search query (known as the *pattern*).
+
+The `grep` command takes three arguments:
+
+1. The pattern used to match lines in a file.
+2. Zero or more flags to customize the matching behavior.
+3. One or more files in which to search for matching lines.
+
+Your task is to implement the `grep` function, which should read the contents
+of the specified files, find the lines that match the specified pattern
+and then output those lines as a single string. Note that the lines should
+be output in the order in which they were found, with the first matching line
+in the first file being output first.
+
+As an example, suppose there is a file named "input.txt" with the following contents:
+
+```text
+hello
+world
+hello again
+```
+
+If we were to call `grep "hello" input.txt`, the returned string should be:
+
+```text
+hello
+hello again
+```
+
+### Flags
+
+As said earlier, the `grep` command should also support the following flags:
+
+- `-n` Print the line numbers of each matching line.
+- `-l` Print only the names of files that contain at least one matching line.
+- `-i` Match line using a case-insensitive comparison.
+- `-v` Invert the program -- collect all lines that fail to match the pattern.
+- `-x` Only match entire lines, instead of lines that contain a match.
+
+If we run `grep -n "hello" input.txt`, the `-n` flag will require the matching
+lines to be prefixed with its line number:
+
+```text
+1:hello
+3:hello again
+```
+
+And if we run `grep -i "HELLO" input.txt`, we'll do a case-insensitive match,
+and the output will be:
+
+```text
+hello
+hello again
+```
+
+The `grep` command should support multiple flags at once.
+
+For example, running `grep -l -v "hello" file1.txt file2.txt` should
+print the names of files that do not contain the string "hello".
+
+* * * *
+
+For installation and learning resources, refer to the
+[Ruby resources page](http://exercism.io/languages/ruby/resources).
+
+For running the tests provided, you will need the Minitest gem. Open a
+terminal window and run the following command to install minitest:
+
+    gem install minitest
+
+If you would like color output, you can `require 'minitest/pride'` in
+the test file, or note the alternative instruction, below, for running
+the test file.
+
+Run the tests from the exercise directory using the following command:
+
+    ruby grep_test.rb
+
+To include color from the command line:
+
+    ruby -r minitest/pride grep_test.rb
+
+
+## Source
+
+Conversation with Nate Foster. [http://www.cs.cornell.edu/Courses/cs3110/2014sp/hw/0/ps0.pdf](http://www.cs.cornell.edu/Courses/cs3110/2014sp/hw/0/ps0.pdf)
+
+## Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/grep/grep_test.rb
+++ b/exercises/grep/grep_test.rb
@@ -1,0 +1,402 @@
+require 'minitest/autorun'
+require_relative 'grep'
+
+# Common test data version: 1.2.0 4f2efaa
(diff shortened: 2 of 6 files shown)


exercises/grep/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+The `grep` command takes three arguments:
+
+1. The pattern used to match lines in a file.
+2. Zero or more flags to customize the matching behavior.
+3. One or more files in which to search for matching lines.
+
+Your task is to implement the `grep` function, which should read the contents
+of the specified files, find the lines that match the specified pattern
+and then output those lines as a single string. Note that the lines should
```
kotp: Do we need them to output or return the lines as a single string?  (We want to be careful here, as output such as to the screen or to a file, as opposed to the return of a method...) The tests as they are written right now does not support testing for output.

exercises/grep/README.md
```diff
@@ -0,0 +1,95 @@
...
+As an example, suppose there is a file named "input.txt" with the following contents:
+
+```text
+hello
+world
+hello again
+```
+
+If we were to call `grep "hello" input.txt`, the returned string should be:
+
```
kotp: The language here is `returned` and so doesn't match the language commented on above.

kotp:
I noticed the conflglet failure as well as reported by Travis CI.  I may have some time to look at the file name problem indicated as well as the configlet issue, if someone doesn't beat me to it, or already have some ideas.

exercises/grep/README.md
pgaspar: That's a great question. The readme (originally from [problem-specifications](https://github.com/exercism/problem-specifications/blob/master/exercises/grep/description.md)) occasionally mention "output", but all the implementations I saw tested solutions using return values. I'm not sure how we would reliably test this otherwise.

There's some discussion around this [here](https://github.com/exercism/ruby/issues/946#issuecomment-471363505).

exercises/grep/README.md
pgaspar: (:+1: please see my response above)

pgaspar:
> I noticed the conflglet failure as well as reported by Travis CI.

Yes, I noted above that before merging we need to add this exercise to `config.json`. I didn't do that because I wasn't sure where @F3PiX wanted it.

> I may have some time to look at the file name problem indicated as well as the configlet issue, if someone doesn't beat me to it, or already have some ideas.

Not sure what file name problem you're referring to 🤔

kotp:
>> I may have some time to look at the file name problem indicated as well as the configlet issue, if someone doesn't beat me to it, or already have some ideas.

> Not sure what file name problem you're referring to 🤔 

I think it was corrected... so that part is good to go.

exercises/grep/README.md
kotp: > Not sure how we could reliably test this otherwise

Minitest has an assertion for output... but it wouldn't be hard even without it, as we have access to `STDOUT` and friends regardless.  Perhaps unsurprisingly, the assertion is `assert_output` and `must_output`.

kotp: I think setting some running code with a "library guard" would be a great inclusion to the exercise.  It is a simple thing to teach (and I already mention it as a mentor when people do include running code, or I need to show a bug, I will explain why I wrote it like I did in the code, so they can copy and paste it directly in the library file).

F3PiX:
> we need to add this exercise to `config.json`. I didn't do that because I wasn't sure where @F3PiX wanted it.

It should be somewhere at the end of the track. The file reading is a feature above 'level 3'.  And moreover, it's also not the simplest form of reading a file. And it's also totally new.   

I'd say: let's make it `"core": false, `"unlocked_by": "12-days"` and difficulty: 8. (Difficulty and topics are not yet part of the Track Anatomy set up, so I don't have opinions on those.  For the difficulty I tend to use 2 difficulties for each level but that's not official.)

So, it starts as a side, so the mentors can play with it first, and we can come up with mentor notes. 
When we know more know how difficult it really is for students, we can add it to the cores.

exercises/grep/grep_test.rb
```diff
@@ -0,0 +1,402 @@
...
+    skip
+    pattern = "may"
+    flags = []
+    files = ["midsummer-night.txt"]
+    expected = <<~EXPECTED.rstrip
+      Nor how it may concern my modesty,
+      But I beseech your grace that I may know
+      The worst that may befall me in this case,
+    EXPECTED
+
```
F3PiX: Thinking TDD, would it make sense to put all the no-flags tests at the beginning?

exercises/grep/README.md
pgaspar: @kotp didn't know about `assert_output` and `must_output`, thanks!

> I think setting some running code with a "library guard" would be a great inclusion to the exercise.

Could you help me understand this better, please? Maybe share an instance of one of these mentoring sessions so I can see what you're referring to?

exercises/grep/grep_test.rb
pgaspar: The order is defined by the canonical data. Do we want to go against that? (It may be worth considering opening a PR on the problem-specifications repo if you feel strongly about changing the order everywhere).

F3PiX: Aha, no, we don't want to go against that. And as the TDD sequence may differ from language to language, I don't think it's worth to change it in the prob-specs.

exercises/grep/README.md
kotp:
```diff
- and then output those lines as a single string
+ and then output those lines.

pgaspar: The readme comes from problem-specifications - maybe it's worth opening an issue there so the language is clarified?

pgaspar:
Thank you both!

PR #950, opened 2019-03-25
acronym: Update and exclude new tests

pgaspar:
We've [recently](https://github.com/exercism/ruby/pull/931#discussion_r251869850) stopped updating the Acronym exercise because of two new test cases:

* `test_apostrophes` (introduced by https://github.com/exercism/problem-specifications/pull/1377)
* `test_underscore_emphasis` (introduced by https://github.com/exercism/problem-specifications/pull/1436)

This PR changes the generator to automatically exclude these two test cases. With this change we can still get updates from the canonical version without adding the tests we don't want at the same time.

If the idea of excluding certain tests proves useful we may want to consider supporting it in the base `ExerciseCase` so it can be easily be used by other generators.

### DIFF
--- a/lib/generator/test_template.erb
+++ b/lib/generator/test_template.erb
@@ -6,5 +6,5 @@ class <%= exercise_test_classname %> < Minitest::Test
 <%=
   test_cases.map.with_index do |test_case, index|
     test_case.to_s(index.zero?)
-  end.join("\n")
+  end.compact.join("\n")
 %>end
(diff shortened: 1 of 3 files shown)


lib/generator/test_template.erb
```diff
@@ -6,5 +6,5 @@ class <%= exercise_test_classname %> < Minitest::Test
 <%=
   test_cases.map.with_index do |test_case, index|
     test_case.to_s(index.zero?)
-  end.join("\n")
+  end.compact.join("\n")
```
pgaspar: Without this change the resulting `acronym_test.rb` would end with:

```ruby
  end


end
```

Instead of:

```ruby
  end
end
```

F3PiX: 😍

kotp:
Thank you @pgaspar

Issue #935, opened 2019-02-08
Temporarily not accepting changes to some exercises


emcoding:
A massive overhaul of the Ruby track is currently underway.
One of the side effects is that changes to current exercises can have a huge impact on the new structure, unknown to the contributor.

Therefore, we don't accept changes, at least not without discussing them in an issue first, for the exercises listed below.
Other exercises, including adding new, have less impact to the Track Anatomy in general. (We still prefer an issue first, before a PR.)
In case your looking to contribute, check this issue: https://github.com/exercism/ruby/issues/900 for our wishlist. 

Thanks for your understanding! 

---

Hello World
Two Fer
Resistor Color Duo
Acronym
High Scores
Matrix
Series 
Word Count
Hamming
Raindrops
Isogram
Scrabble Score
Luhn 
Clock
Twelve Days
Tournament

ETL
Grep
Phone Number
Proverb
Resistor Color Trio

kenliu:
any update on this?

kotp:
Are we due a status update on this... @F3PiX

F3PiX:
Yes, we are. 

My thinking is to add a list here of exercises we don't accept changes for, at least not without discussing it in an issue first. I'm thinking of the core exercises, and some exercises that are currently researched as core candidates. 

How does that sound?
@exercism/ruby

F3PiX:
I updated the description with the list of exercises.

kotp:
How about new exercises coming in...

F3PiX:
I don't think everyone of those needs to be under Track Anatomy scrutiny. If needed, I'll explain. For new exercises it's mostly the start of the track that I'd worry about. In general, more exercises is more for us and for students to choose from :-)  

But! I'll add a link to exercises we _are_ looking for help for.

kotp:
Well, 'reverse' is probably at the beginning... Can you give your thumbs up over there... it is the reason I realized we needed an update over here.

PR #952, opened 2019-03-27
[WIP] Handle test cases with to_s evaluating to nil

pgaspar:
This change makes it possible for `ExerciseCase`s to override the `to_s` method, returning `nil` for tests we don't want for the Ruby track. You can see an example of this on #953.

Without this change these excluded tests would still not appear on the generated test file, but an extra new-line would be present where they would be, given how:

```ruby
# Without the change:
[1, 2, nil, 4].join("\n")
# => "1\n2\n\n4"

# With the change:
[1, 2, nil, 4].compact.join("\n")
# => "1\n2\n4"
```

In practice, what this means is that, for example, on the Acronym update from #953 the generated test file `acronym_test.rb` will end with:

```ruby
  end
end
```

Instead of:

```ruby
  end


end
```

Note: #953 relies on this change.

### DIFF
--- a/lib/generator/test_template.erb
+++ b/lib/generator/test_template.erb
@@ -6,5 +6,5 @@ class <%= exercise_test_classname %> < Minitest::Test
 <%=
   test_cases.map.with_index do |test_case, index|
     test_case.to_s(index.zero?)
-  end.join("\n")
+  end.compact.join("\n")
 %>end


kotp:
Note #935 though.

kotp:
Thank you @pgaspar

PR #953, opened 2019-03-27
[WIP] acronym: Update and exclude new tests

pgaspar:
Note: this PR was initially merged and reverted (#950). This new version separates the changes to the test template into its own PR (#952). This PR thus depends on PR #952.

We've [recently](https://github.com/exercism/ruby/pull/931#discussion_r251869850) stopped updating the Acronym exercise because of two new test cases:

* `test_apostrophes` (introduced by https://github.com/exercism/problem-specifications/pull/1377)
* `test_underscore_emphasis` (introduced by https://github.com/exercism/problem-specifications/pull/1436)

This PR changes the generator to automatically exclude these two test cases. With this change we can still get updates from the canonical version without adding the tests we don't want at the same time.

If the idea of excluding certain tests proves useful we may want to consider supporting it in the base `ExerciseCase` so it can be easily be used by other generators.

### DIFF
--- a/exercises/acronym/.meta/generator/acronym_case.rb
+++ b/exercises/acronym/.meta/generator/acronym_case.rb
@@ -4,4 +4,19 @@ class AcronymCase < Generator::ExerciseCase
   def workload
     assert_equal(expected,  "Acronym.abbreviate('#{phrase}')")
   end
+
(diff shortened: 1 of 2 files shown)


kotp:
Note #935 though.

PR #961, opened 2019-05-30
Move ETL close to WordCount

F3PiX:
- This is preparing for promoting ETL to a core exercise. As a core, ETL should go just after Word Count. Let's first check if, as a side,  it isn't too hard on this point in the track by making it unlocked by WordCount.   
- The difficulty score is arbitrary; the Track Anatomy Project didn't touch those yet. I just made it one step more difficult than WordCount.

@exercism/ruby

### DIFF
--- a/config.json
+++ b/config.json
@@ -420,8 +420,8 @@
       "slug": "etl",
       "uuid": "0d66f3db-69a4-44b9-80be-9366f8b189ec",
       "core": false,
-      "unlocked_by": "scrabble-score",
-      "difficulty": 1,
+      "unlocked_by": "word-count",
+      "difficulty": 4,
       "topics": [
         "loops",
         "maps",


kotp:
The changes look good.  However the Readme suggests that it should be after the Scrabble game is done... otherwise it isn't really indicating the historical truth.

F3PiX:
That's an interesting observation @kotp. 
For reference, here's the relevant part of the description:

```
We're going to extract some scrabble scores from a legacy system.

The old system stored a list of letters per score:

1 point: "A", "E", "I", "O", "U", "L", "N", "R", "S", "T",
2 points: "D", "G",
...
The shiny new scrabble system instead stores the score per letter, which makes it much faster and easier to calculate the score for a word. It also stores the letters in lower-case regardless of the case of the input letters:

"a" is worth 1 point.
"b" is worth 3 points.
....

Your mission, should you choose to accept it, is to transform the legacy data format to the shiny new format.
```

I did not interpret this as if the legacy was the system from a preceding exercise. On the contrary, I read it as inspiration to use the second format for the Scrabble Score exercise :-) 

Do we need to take action as follow up? 
(I'm merging this for now, but please let me know if we need to follow up.)

kotp:
I think if the only reference to "a legacy system" is indicating "the old system" on the following line, then the old system would be our prior Scrabble solution?  If I had seen "An old system" it would have indicated to me some system, rather than the one that we may be familiar with.

It may be due to having done "the old system" in a specific way that we are reminded of that old system specifically.

kotp:
There may be nothing to do here.  If it is before any other Scrabble exercises, then it will automatically not be referring to that, as it hasn't happened yet.

PR #964, opened 2019-06-04
PrimeFactors: Replace `for` with `of` (for is a Ruby keyword)

yyyc514:
`for` is a Ruby keyword, so this is just bad naming if you ask me.

### DIFF
--- a/exercises/prime-factors/.meta/solutions/prime_factors.rb
+++ b/exercises/prime-factors/.meta/solutions/prime_factors.rb
@@ -1,5 +1,5 @@
 class PrimeFactors
-  def self.for(number)
+  def self.of(number)
     factors = []
     divisor = 2
     while number > 1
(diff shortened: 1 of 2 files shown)


kotp:
I agree, but there is going to need to be changes to the generator to put this into play.

yyyc514:
Fixed solution also.

yyyc514:
> I agree, but there is going to need to be changes to the generator to put this into play. 

Think I would need a clearer hint than that?

kotp:
"Need" was too strong a word, but it is good to put the generator into play as we make changes.  Look to [the generator section](https://github.com/exercism/ruby#implementing-a-generator) of the Readme for more details.

kotp:
:camel:

yyyc514:
> but not because for is a keyword.

Maybe partly because. :-)  Anyways, thanks for approval.

PR #967, opened 2019-06-13
Discuss: FlattenArray: Is this actually a valid solution?

yyyc514:
I actually presumed that was NOT the point of this exercise and always have be students implement an actual solution from primitives... I think the README should be updated to disallow flatten/compact.  Am I misunderstanding the goal of this exercise?

If we have some agreement I can add a working solution.


F3PiX:
There's already a discussion https://github.com/exercism/ruby/issues/966
We're nog going to implement this before we finished this discussion.

PR #971, opened 2019-07-12
Promote Resistor Color Duo to core

F3PiX:
Resistor Color Duo was introduced in March, and added as a side exercise on [Level 1](https://github.com/exercism/ruby/issues/969) first.
As a core exercise, it brings in the `integer` datatype to level 1, and a simple array iterator (for beginners) or reading values from a hash. 
Looking at the [community solutions](https://exercism.io/tracks/ruby/exercises/resistor-color-duo/solutions), there's lots of opportunities to guide to more idiomatic Ruby.

(cc ~@exercism/track-maintainers~ Ruby maintainers)

### DIFF
--- a/config.json
+++ b/config.json
@@ -304,7 +304,7 @@
       "slug": "leap",
       "uuid": "06eaa2dd-dc80-4d38-b10d-11174183b0b6",
       "core": false,
-      "unlocked_by": "two-fer",
+      "unlocked_by": "resistor-color-duo",
       "difficulty": 1,
       "topics": [
         "booleans",
--- a/config.json
+++ b/config.json
@@ -27,6 +27,16 @@
         "strings"
       ]
     },
+    { "slug": "resistor-color-duo",
+      "uuid": "57b0c57e-26a5-4ccf-aaf2-2fefddd918c1",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "array",
+        "loops"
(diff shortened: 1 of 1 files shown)


cmccandless:
For the record, the correct team to tag on this is @exercism/ruby. @exercism/track-maintainers subscribes maintainers of *all* tracks to this issue.

SleeplessByte:
That might have been the point @cmccandless , but unsure :)

kotp:
I think it was intentional due to exercism/website-copy#1160

F3PiX:
It was not intented for _all_ maintainers, only ruby. Sorry.

config.json
```diff
@@ -27,6 +27,16 @@
         "strings"
       ]
     },
+    { "slug": "resistor-color-duo",
+      "uuid": "57b0c57e-26a5-4ccf-aaf2-2fefddd918c1",
+      "core": true,
+      "unlocked_by": null,
```
pgaspar: Shouldn't this be `"unlocked_by": "two-fer"`?

F3PiX: No, cores are unlocked by null. It's the place in the file that defines the position in the track, for cores.  
(Why do even add this key for cores 🤔...? )

F3PiX:
Thanks @pgaspar @kotp ! Here we go.

config.json
pgaspar: Sorry, my bad!

PR #973, opened 2019-07-20
resistor-color: New exercise

### LINKED ISSUE #970: Add Resistor Color (#1) to replace Flatten Array as a side exercise
Currently, we have only 2 exercises that are unlocked by Hello World: Gigasecond and Flatten Array. 
Flatten Array [is problematic](https://github.com/exercism/ruby/issues/966), because it's basically a built-in Ruby method. And to implement it yourself, it's not a nice warmer-upper. 
There's currently no other candidate for this position on the track. 
Although Resistor Color is basically 'tutorial level', and therefore too easy for the track, it's a better warmer-upper than Flatten Array. So let's add it as a side exercise.

pgaspar:
Closes #970. Have a look at the canonical data [here](https://github.com/exercism/problem-specifications/tree/master/exercises/resistor-color). 🕺

### DIFF
--- a/exercises/resistor-color/.meta/generator/resistor_color_case.rb
+++ b/exercises/resistor-color/.meta/generator/resistor_color_case.rb
@@ -1,0 +1,20 @@
+require 'generator/exercise_case'
+
+class ResistorColorCase < Generator::ExerciseCase
+  def workload
+    send("#{snake_case(property)}_workload")
+  end
+
+  private
+
+  def color_code_workload
+    assert_equal(expected, "ResistorColor.color_code(#{color.inspect})")
+  end
+
+  def colors_workload
+    [
+      "expected = #{expected}",
+      "assert_equal expected, ResistorColor.colors"
+    ]
+  end
+end
--- a/exercises/resistor-color/.meta/generator/resistor_color_case.rb
+++ b/exercises/resistor-color/.meta/generator/resistor_color_case.rb
@@ -11,10 +11,12 @@ class ResistorColorCase < Generator::ExerciseCase
     assert_equal(expected, "ResistorColor.color_code(#{color.inspect})")
   end
 
+  # Instead of specifying a `colors` method like the canonical data suggests
+  # we expect the student to define a COLORS constant.
   def colors_workload
     [
       "expected = #{expected}",
-      "assert_equal expected, ResistorColor.colors"
+      "assert_equal expected, ResistorColor::COLORS"
     ]
   end
 end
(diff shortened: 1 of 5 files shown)


F3PiX:
Yay!!! Thanks Pedro :-)
One question:
I don't know nottin' about the generator. What are the consequences / problems of using `ResistorColor::COLORS` in the tests to return the array of colors?

I'm asking, because from a track anatomy point of view, and to make this suitable to be unlocked by HelloWorld, this should be a single method with a one-liner. And, it feels a bit weird that we have one method that's only there to satisfy the tests. 

Is that possible? Or would it mean that we have to manually regen the tests everytime something changes? 

Note: I'm working on a new description for Resistor Color (the series). I'd vote for leaving the description as is for now. Is that okay by everyone?

kotp:
> Note: I'm working on a new description for Resistor Color (the series). I'd vote for leaving the description as is for now. Is that okay by everyone?

Thanks for mentioning it, I almost mentioned it in the approval.  Wasn't quite sure how to express it.

pgaspar:
@F3PiX:

> I don't know nottin' about the generator. What are the consequences / problems of using ResistorColor::COLORS in the tests to return the array of colors?

Hmm, I think I can make that work - I think I understand what you want here.

> I'm asking, because from a track anatomy point of view, and to make this suitable to be unlocked by HelloWorld, this should be a single method with a one-liner. And, it feels a bit weird that we have one method that's only there to satisfy the tests.

Heh, this is the way to interpret the canonical specification, which specifies a second method to return the colors array ([here](https://github.com/exercism/problem-specifications/blob/master/exercises/resistor-color/canonical-data.json#L34-L39)). But I understand your points above and will change the generator to make that work :+1:

F3PiX:
Briliant! Thanks. 

> Heh, this is the way to interpret the canonical specification, which specifies a second method to return the colors array

Yeah, it's not a bug, it's a feature :-) I'm still happy with asking to return the array of colors, and this is exactly how I envisioned it. 🙏 Supergreat.

exercises/resistor-color/.meta/generator/resistor_color_case.rb
```diff
@@ -0,0 +1,22 @@
...
+  def color_code_workload
+    assert_equal(expected, "ResistorColor.color_code(#{color.inspect})")
+  end
+
+  # Instead of specifying a `colors` method like the canonical data suggests
+  # we expect the student to define a COLORS constant.
+  def colors_workload
+    [
+      "expected = #{expected}",
+      "assert_equal expected, ResistorColor::COLORS"
```
F3PiX: 😍

kotp:
Thank you @pgaspar ...

PR #980, opened 2019-08-01
Add reverse-string to ruby exercises

bergbergs:
Add reverse-string exercise to ruby/exercises.

### DIFF
--- a/config.json
+++ b/config.json
@@ -17,6 +17,16 @@
       ]
     },
     {
+      "slug": "reverse-string",
+      "uuid": "64c64613-9dd9-461e-90ba-90f84e17e197",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "strings"
+      ]
+    },
+    {
       "slug": "two-fer",
       "uuid": "1304b188-6d08-4361-be40-c6b1b88e5e54",
       "core": true,
--- a/config.json
+++ b/config.json
@@ -17,16 +17,6 @@
       ]
     },
     {
-      "slug": "reverse-string",
-      "uuid": "64c64613-9dd9-461e-90ba-90f84e17e197",
-      "core": false,
-      "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-        "strings"
-      ]
-    },
-    {
       "slug": "two-fer",
       "uuid": "1304b188-6d08-4361-be40-c6b1b88e5e54",
       "core": true,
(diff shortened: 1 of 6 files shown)


config.json
```diff
@@ -16,6 +16,16 @@
         "strings"
       ]
     },
+    {
```
F3PiX: In this file, the core exercises come first. The side exercises, like this one, should go below the core exercises (Tournament is the last core.) Easiest is to move it to the end of the file. There's no real system other than having the cores first, but lately we started adding new ones to the end.  

You can of course wait to make the change until we decided on adding the exercise or not.

exercises/reverse-string/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Reverse a string
+
+For example:
+input: "cool"
+output: "looc"
+
+# Hints
+
+Ruby documentation is pretty extensive! Look through Ruby's built-in `String` methods and see if you can find something that will help you out with this exercise.
```
F3PiX:
```suggestion
The Ruby library is pretty extensive! Look through Ruby's built-in `String` methods and see if you can find something that will help you out with this exercise.
```
It's more the language than the documentation that's extensive. Can you live with this?

bergbergs:
Thanks for directing me to the documentation on restructuring. Hadn't seen that doc previously. I'll review it and keep my eye on the discussion in the meantime.

config.json
bergbergs: I have updated the config.json file. Thanks!

F3PiX:
@bergbergs I'm sorry, but as nobody voted to add this, I'm not going to merge. I sincerely hope that you will keep contributing, and I'm glad you took on Transpose. :-)  

@pgaspar, @kotp Shouldn't we start to document which changes/exercises have been considered or dicussed previously? So that (new) contributors can see if exercises are rejected as possible candidates, and not put in the work. If only I could think of a good way to do that....

kotp:
> Shouldn't we start to document which changes/exercises have been considered or dicussed previously? So that (new) contributors can see if exercises are rejected as possible candidates, and not put in the work. If only I could think of a good way to do that...

We could have a ".rejected_exercises" file, that we can list.

kotp:
Though, before implementing exercises as a pull request, there should probably be an issue created.  Where we can verify the points that it teaches, etc., then we can approve the addition, or reject the addition.  It may still get to the PR stage, but at least we can tag the issue as "Won't Implement" by tag, and the PR as well.

F3PiX:
I created issue #983 to discuss how to document the rejected exercises.

Issue #949, opened 2019-03-25
Transpose: change tests to use strings instead of heredocs


emcoding:
This came up in #929 
 
I prefer to not have heredocs in tests before we introduced them in an exercise. 

```ruby
input = "ABC\n123"

# instead of the current

input = <<~INPUT.strip
  ABC
  123
INPUT
```

For reference: 
>  not using indent_heredoc on transpose's generator, which is what generates the heredocs in the test files.

goalaleo:
I agree. For some reason (likely some editor setting) the lines consisting white space + empty line, are automatically stripped of the whitespace causing those tests to fail for me.

<img width="339" alt="Screenshot 2019-04-23 10 48 14" src="https://user-images.githubusercontent.com/3275302/56563941-a4360380-65b5-11e9-96f6-603feaf63739.png">

bergbergs:
If this is still a concern, I'd be happy to take a stab at a pull request on this one.

F3PiX:
Thanks @bergbergs Sounds great. Please do.

kytrinyx:
Closing, as this was fixed in #981

PR #981, opened 2019-08-01
Update transpose test case with new case_helper method.

bergbergs:
Issue #949 indicates that it would be preferable to display the `input` and `expected` values as a single-line string rather than as a heredoc. 

From the issue:

    input = "ABC\n123"
    
    # instead of the current
    
    input = <<~INPUT.strip
        ABC
        123
    INPUT

In this pull request, I have made the following changes:

- Create a new case_helper method that can take an array of strings and display with a separator defaulting to `"\n"`.
- Update `transpose_case.rb` so that it uses this `on_single_line` case helper where it previously used `indent_heredoc`.
- Generate updated test cases for `transpose_test.rb`.

All test cases still complete successfully after changes.

Please let me know if I can make any additional changes. Thanks!

### DIFF
--- a/config.json
+++ b/config.json
@@ -559,8 +559,8 @@
       "slug": "flatten-array",
       "uuid": "2df8ed82-2a04-4112-a17b-7813bcdc0e84",
       "core": false,
-      "unlocked_by": "hello-world",
-      "difficulty": 1,
+      "unlocked_by": "series",
+      "difficulty": 3,
       "topics": [
         "arrays",
         "recursion"
--- a/exercises/transpose/.meta/generator/transpose_case.rb
+++ b/exercises/transpose/.meta/generator/transpose_case.rb
@@ -4,18 +4,12 @@ class TransposeCase < Generator::ExerciseCase
 
   def workload
     [
-      "input = #{indent_heredoc(input_lines, 'INPUT', 2, delimiter_mod)}",
+      "input = #{to_string(lines)}",
       "",
-      "expected = #{indent_heredoc(expected, 'EXPECTED', 2, delimiter_mod)}",
+      "expected = #{to_string(expected)}",
       "",
-      "assert_equal expected, Transpose.transpose(input)",
+      "assert_equal expected, Transpose.transpose(input)"
     ]
-  end
-
-  private
-
-  def delimiter_mod
-    ".strip"
   end
 
 end
--- a/lib/generator/exercise_case/case_helpers.rb
+++ b/lib/generator/exercise_case/case_helpers.rb
@@ -26,6 +26,15 @@ module Generator
         ].join("\n")
       end
 
+      # combine array of string elements into a single string
+      # as part of workload with optional separator
+      #
+      #  example usage:  to_string(["foo", "bar"])
+      #  example output: "foo\nbar"
+      def to_string(phrases, separator="\n")
+        phrases.join(separator).inspect
+      end
+
       def underscore(number)
         fail ArgumentError, "#{number.inspect} is not an Integer" unless number.is_a? Integer
         number.to_s.reverse.gsub(/...(?=.)/, '\&_').reverse
--- a/test/generator/exercise_case/case_helpers_test.rb
+++ b/test/generator/exercise_case/case_helpers_test.rb
@@ -53,9 +53,36 @@ module Generator
           indent_heredoc(["foo", "bar"], 'TEXT', 1)
         end
       end
+
       def test_heredoc
         expected = "<<~TEXT\n foo\n bar\nTEXT"
         assert_equal expected, HeredocCase.new.workload
+      end
+
+      class ToStringCase
+        include CaseHelpers
+
+        def workload(phrases)
+          to_string(phrases)
+        end
+      end
+
+      def test_to_string_with_two_strings
+        phrases = ["foo", "bar"]
+        expected = "foo\nbar".inspect
+        assert_equal expected, ToStringCase.new.workload(phrases)
+      end
+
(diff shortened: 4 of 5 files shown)


exercises/transpose/.meta/generator/transpose_case.rb
```diff
@@ -4,18 +4,12 @@ class TransposeCase < Generator::ExerciseCase
 
   def workload
     [
-      "input = #{indent_heredoc(input_lines, 'INPUT', 2, delimiter_mod)}",
+      "input = #{on_single_line(lines)}",
```
kotp: `not_the_heredoc(lines)`  Because the "lines" itself indicates that it could not possibly be on a single line.  We are addressing heredoc vs string with lines.  Not that the name is necessarily good.  But the "single line" name seems to indicate we want to remove new lines, and that isn't the case.

lib/generator/exercise_case/case_helpers.rb
```diff
@@ -26,6 +26,15 @@ def indent_heredoc(lines, delimiter, depth = 0, delimiter_method = nil)
...
       end
 
+      # combine array of string elements into a single string
+      # as part of workload with optional separator
+      #
+      #  example usage:  on_single_line(["foo", "bar"])
+      #  example output: "foo\nbar"
+      def on_single_line(lines, separator="\n")
+        lines.join(separator).inspect
+      end
```
kotp: Needs a test to catch if this behavior changes in a way that is not wanted.

exercises/transpose/.meta/generator/transpose_case.rb
bergbergs: I've updated the name in the latest version.

lib/generator/exercise_case/case_helpers.rb
bergbergs: I've added a few tests to catch behavior with strings, and with mixed classes (string, integer, nil) in the latest version.

test/generator/exercise_case/case_helpers_test.rb
```diff
@@ -57,6 +57,29 @@ def test_heredoc
...
       end
+
+      class ToStringCase
+        include CaseHelpers
+
+        def workload(phrases)
+          to_string(phrases)
+        end
+      end
+      def test_to_string_with_two_strings
```
F3PiX: NIce! And could you please add whitespace between the tests?
(Unless there's a secret convention that I'm not aware of for following the spacing as it is in line 56. )

test/generator/exercise_case/case_helpers_test.rb
```diff
@@ -57,6 +57,29 @@ def test_heredoc
...
         assert_equal expected, HeredocCase.new.workload
       end
+
+      class ToStringCase
+        include CaseHelpers
+
+        def workload(phrases)
+          to_string(phrases)
+        end
+      end
```
kotp:
```suggestion
      end

```

test/generator/exercise_case/case_helpers_test.rb
```diff
@@ -57,6 +57,29 @@ def test_heredoc
...
+      def test_to_string_with_two_strings
+        phrases = ["foo", "bar"]
+        expected = "foo\nbar".inspect
+        assert_equal expected, ToStringCase.new.workload(phrases)
+      end
+      def test_to_string_with_string_and_integer
```
kotp:
```suggestion

      def test_to_string_with_string_and_integer
```

test/generator/exercise_case/case_helpers_test.rb
```diff
@@ -57,6 +57,29 @@ def test_heredoc
...
+        phrases = ["foo", 123]
+        expected = "foo\n123".inspect
+        assert_equal expected, ToStringCase.new.workload(phrases)
+      end
+      def test_to_string_with_string_with_nil
```
kotp:
```suggestion

      def test_to_string_with_string_with_nil
```

bergbergs:
Alrighty, should be all set now with all of the requested changes as far as I can tell.

pgaspar:
@bergbergs would you mind squashing the 11 commits into one, please? You can see guidelines on how to do this [here](https://github.com/exercism/docs/blob/master/contributing/git-basics.md#squashing).

bergbergs:
> @bergbergs would you mind squashing the 11 commits into one, please? You can see guidelines on how to do this [here](https://github.com/exercism/docs/blob/master/contributing/git-basics.md#squashing).

I think this is all taken care of now. Thanks all for the patience on this PR; I'll get the hang of it one day! 😅

config.json
```diff
@@ -559,8 +559,8 @@
       "slug": "flatten-array",
       "uuid": "2df8ed82-2a04-4112-a17b-7813bcdc0e84",
       "core": false,
-      "unlocked_by": "hello-world",
-      "difficulty": 1,
+      "unlocked_by": "series",
+      "difficulty": 3,
```
kotp: Looks like this was merged in, unintended.

pgaspar: I'm not sure what happened here but the merge commit doesn't include this change (thankfully). I merged in mobile so I might have missed something, but I was sure I saw this change had been reverted. It's weird Github still shows it here.

kotp: Oh, good.

kotp: I am seeing: 

https://github.com/exercism/ruby/blob/master/config.json

So it appears to me that it is there changed...

pgaspar: Nope, it's still in `master` 😓 I'm going to revert this change in another PR.

PR #985, opened 2019-08-05
Revert "Revert unintended config.json change from #981"

kotp:
Reverts exercism/ruby#984

As intended via another PR.


pgaspar:
What a mess 😅 Glad it's sorted now.

PR #988, opened 2019-08-08
armstrong-numbers:  port

guygastineau:
I have rebased the last ten commits of @Tuxified's work from #893 onto the current master.
[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/armstrong-numbers/canonical-data.json)

### DIFF
--- a/config.json
+++ b/config.json
@@ -57,17 +57,14 @@
       "unlocked_by": null,
       "difficulty": 2,
       "topics": [
-        "arrays",
-        "sequences",
-        "strings",
-        "regular_expressions"
+        "arrays"
       ]
     },
     {
       "slug": "armstrong-numbers",
       "uuid": "77c5c7e6-265a-4f1a-9bc4-851f8f825420",
       "core": false,
-      "unlocked_by": "isogram",
+      "unlocked_by": "series",
       "difficulty": 3,
       "topics": [
         "math"
--- a/config.json
+++ b/config.json
@@ -57,7 +57,20 @@
       "unlocked_by": null,
       "difficulty": 2,
       "topics": [
-        "arrays"
+        "arrays",
+        "sequences",
+        "strings",
+        "regular_expressions"
+      ]
+    },
+    {
+      "slug": "armstrong-numbers",
+      "uuid": "77c5c7e6-265a-4f1a-9bc4-851f8f825420",
+      "core": false,
+      "unlocked_by": "isogram",
+      "difficulty": 3,
+      "topics": [
+        "math"
       ]
     },
     {
--- a/exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
+++ b/exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
@@ -1,0 +1,12 @@
+require 'generator/exercise_case'
+
+class ArmstrongNumbersCase < Generator::ExerciseCase
+
+  def workload
+    assert_or_refute(expected, call_armstrong)
+  end
+
+  def call_armstrong
+    "ArmstrongNumbers.include?(#{underscore(input['number'])})"
+  end
+end
--- a/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
+++ b/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
@@ -1,0 +1,7 @@
+class ArmstrongNumbers
+  def self.is_valid?(number)
+    exponent = number.digits.size
+
+    number.digits.sum { |digit| digit**exponent } == number
+  end
+end
--- a/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
+++ b/exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
@@ -1,5 +1,5 @@
 class ArmstrongNumbers
-  def self.is_valid?(number)
+  def self.include?(number)
     exponent = number.digits.size
 
     number.digits.sum { |digit| digit**exponent } == number
--- a/exercises/armstrong-numbers/README.md
+++ b/exercises/armstrong-numbers/README.md
@@ -33,6 +33,7 @@ To include color from the command line:
 
     ruby -r minitest/pride armstrong_numbers_test.rb
 
+
 ## Source
 
 Wikipedia [https://en.wikipedia.org/wiki/Narcissistic_number](https://en.wikipedia.org/wiki/Narcissistic_number)
--- a/exercises/armstrong-numbers/README.md
+++ b/exercises/armstrong-numbers/README.md
@@ -1,0 +1,41 @@
+# Armstrong Numbers
+
+An [Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) is a number that is the sum of its own digits each raised to the power of the number of digits.
+
+For example:
+
+- 9 is an Armstrong number, because `9 = 9^1 = 9`
+- 10 is *not* an Armstrong number, because `10 != 1^2 + 0^2 = 1`
+- 153 is an Armstrong number, because: `153 = 1^3 + 5^3 + 3^3 = 1 + 125 + 27 = 153`
+- 154 is *not* an Armstrong number, because: `154 != 1^3 + 5^3 + 4^3 = 1 + 125 + 64 = 190`
+
+Write some code to determine whether a number is an Armstrong number.
+
+* * * *
+
+For installation and learning resources, refer to the
+[Ruby resources page](http://exercism.io/languages/ruby/resources).
+
+For running the tests provided, you will need the Minitest gem. Open a
+terminal window and run the following command to install minitest:
+
+    gem install minitest
+
+If you would like color output, you can `require 'minitest/pride'` in
+the test file, or note the alternative instruction, below, for running
+the test file.
+
+Run the tests from the exercise directory using the following command:
+
+    ruby armstrong_numbers_test.rb
+
+To include color from the command line:
+
+    ruby -r minitest/pride armstrong_numbers_test.rb
+
+## Source
+
+Wikipedia [https://en.wikipedia.org/wiki/Narcissistic_number](https://en.wikipedia.org/wiki/Narcissistic_number)
+
+## Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/armstrong-numbers/armstrong_numbers_test.rb
+++ b/exercises/armstrong-numbers/armstrong_numbers_test.rb
@@ -1,59 +1,50 @@
 require 'minitest/autorun'
 require_relative 'armstrong_numbers'
 
+# Common test data version: 1.1.0 b3c2522
 class ArmstrongNumbersTest < Minitest::Test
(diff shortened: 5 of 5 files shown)


guygastineau:
I intend to make a test generator for this.  It has been awhile since I have done so.  I should have some work up for that later tonight or tomorrow.

guygastineau:
At first I thought I would add integer delimiting into the generator. Then I was sad when I realized that converting them back into integers destroys the formatting.

Then I felt silly when I realized that it is written into the test as a string :rofl:

exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  private
+
+  def delimit_integers(integer)
+    integer
+      .digits
+      .each_slice(3)
+      .map(&:join)
+      .join('_')
+      .reverse
```
guygastineau: I guess it would actually be more efficient to `reverse` before `join`, because this way fewer things are reversed.  I think it would be `O(n / 3 + 1)`.

Does that matter you all?

guygastineau: Nevermind, I just tried, but I forgot the order of the digits was reversed by digits (which is necessary), so it just messed things up :smile: Leaving this as it is.

guygastineau:
Okay, I think I am finished messing with this. :smile:

config.json
```diff
@@ -60,6 +60,16 @@
         "arrays"
       ]
     },
+    {
+      "slug": "armstrong-numbers",
+      "uuid": "77c5c7e6-265a-4f1a-9bc4-851f8f825420",
+      "core": false,
+      "unlocked_by": "isogram",
```
F3PiX: @guygastineau I didn't realize before that Isogram has moved since the last time we discussed adding this exercise. Let's make it    
```suggestion
      "unlocked_by": "series",
```
Sorry that I missed this.

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
```diff
@@ -0,0 +1,6 @@
+class ArmstrongNumbers
+  def self.include?(number)
+    raise_to_length = ->(n) { n ** number.digits.size }
+    number.digits.sum(&raise_to_length) == number
+  end
```
F3PiX: LOL Lambda calculus, right? 
Since the solution file is sometimes consulted by (new) mentors and students, and to avoid miscommunications in general, I'd prefer to have a solution that is in sync with the solution that we strive for. 
That's mostly a matter of opinion, I don't think we have a policy for this.

And I'm slightly confused too. ;-) Does this mean that you think the original solution is not good enough? Or?

config.json
F3PiX: Note to self: without the lambda, it could be a level 1 side exercise. Then, it may need a more step by step help in the readme. 

(This doesn't need to be addressed in this PR, unless you'd prefer to do so, Guy. )

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
guygastineau: Oh, I just really wanted to use stabby lambda for the hell of it, and the guide for porting exercises on this track says the solution doesn't need to be ideal (although I think this one is fun 😃)

I have no problem reverting it.  That is a simple fix 😃

(I think that porting guide was written before the profound track anatomy project happened, so wanting example solutions that fit the level of the exercise is a reasonable desire given all that you all have done to polish this track 😃)

config.json
guygastineau: Great, I figured there might need to be some changes to it in the `config.json`.
I'm on it 😃

guygastineau: I mean, I will change it to unlock from `series` I don't especially feel like writing step-by-step instructions for closures in a lvl 1 exercise readme extension.

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
guygastineau: Reverted.

config.json
guygastineau: changed

guygastineau:
Also, I noticed that you all have a very explicit merge history here. I am very impressed :heart_eyes: 

This PR has a load of commits, so I would be more than happy to rebase it into a couple commit.
I would probably squash @Tuxified's commits into 1 (might use `fixup` on one or two that aren't relevant to the final product here), and I can squash mine together, but I'll `fixup` or `drop` any that don't need to be there (like the one introducing the lambda).

Anyway, this is just an idea to keep both contributors in the history without having a 20 commit merge for this simple exercise.  Let me know what you all would prefer, and if I should do anything about it :smile:

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
```diff
@@ -0,0 +1,7 @@
+class ArmstrongNumbers
+  def self.include?(number)
+    exponent = number.digits.size
+
+    number.digits.sum { |digit| digit ** exponent } == number
```
kotp: Don't use spaces around `**` as it emulates how we expect to see it (similar to x<sup>y</sup>, very close together), and makes it clear it isn't a typographical error due to style showing intent.

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
kotp: The example solution is a proof of concept, and if it is not "normal" this means that we will likely spot it quickly, if for some reason it comes to be a solution submitted by a student.  So there are some advantages to having a non ideal solution, as long as it "proves the concept".

kotp: On the other hand, since this isn't `example.rb` it may be being presented in the note as an acceptable solution... one to approve.  In that situation, the story changes.

exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
```diff
@@ -0,0 +1,36 @@
...
```
kotp: Do we do this often enough that this should be in our available tools for generators?

guygastineau: Are you asking if this should be added to the generator class, or are you saying it already exists?

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
guygastineau: I'll change it, but I strongly disagree.  I think operators should always have spaces in languages where it is possible, but I would not consider myself a real rubyist yet (, so I defer to your opinion).

guygastineau:
So, which solution do you all want?

exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class ArmstrongNumbersCase < Generator::ExerciseCase
+
+  def workload
+    case expected
+    when true  then true_workload
+    when false then false_workload
+    else raise "unexpected 'expected' value encountered"
+    end
+  end
```
pgaspar: @guygastineau You can simplify this a bit if you use the `assert_or_refute` helper defined [here](https://github.com/exercism/ruby/blob/master/lib/generator/exercise_case/assertion.rb#L16).

exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
pgaspar: I believe @kotp was asking the former, but it seems it already exists, yes. It's called `underscore` and it's defined [here](https://github.com/exercism/ruby/blob/master/lib/generator/exercise_case/case_helpers.rb#L38-L41).

This, together with `assert_or_refute` from above should make this a 1-line `workload` method, I think.

guygastineau: Okay, thank you.

exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
```diff
@@ -0,0 +1,12 @@
+require 'generator/exercise_case'
+
+class ArmstrongNumbersCase < Generator::ExerciseCase
+
+  def workload
+    assert_or_refute(expected, call_armstrong)
+  end
+
+  def call_armstrong
+    "ArmstrongNumbers.include?(#{underscore(input['number'])})"
```
guygastineau: I kept this function, because otherwise the single line in workload is way over 80 characters long.

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
guygastineau: done

exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
guygastineau: done

guygastineau:
Thanks for all the pointers everybody :smile:

I I will wait to squash into two commits until the work is approved (so I don't have to do it twice)

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
F3PiX: > Oh, I just really wanted to use stabby lambda for the hell of it 
Yeah, you mentioned the lambda fun somewhere else, so that's what I thought it was 🤓 .

FYI: 
My thinking is: The beauty of this exercise to me is that it does 'light math', that can be solved with 'early' Ruby methods without being scary mathy. And then it can be refactored to use `digits.sum`. 
For other purposes, this exercise would need more 'progressions' in other aspects, to make the learning curve between exercises steep enough on the higher levels.
  
Thanks for reverting :-)

exercises/armstrong-numbers/.meta/solutions/armstrong_numbers.rb
F3PiX: In case it's a comfort: no spaces around exponents is recommended by one of the Ruby Styleguides. Not saying I like it, but I settle for consistency :-)
https://rubystyle.guide/#spaces-operators

guygastineau:
Thanks @kotp

I'll rebase this after dinner.

exercises/armstrong-numbers/.meta/generator/armstrong_numbers_case.rb
pgaspar: Looks great :+1:

guygastineau:
I suppose I could have done some squashes mixed in with the fixups, but at least both contributors are represented in the history (which was my main goal here).

guygastineau:
It's ready for you all now.  Thanks for working with me on this ;)

Are there any exercises that need generators? I think there used to be a file or an issue somewhere showing which ones needed a generator implemented.  I'm happy to hunt them down and make them ;)

PR #990, opened 2019-08-09
atbash-cipher: implemented generator

guygastineau:
[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/atbash-cipher/canonical-data.json)

This exercise was only half implemented before, so I had to change the solution.

I regenerated the `README.md`, but there was no change.

### DIFF
--- a/exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
+++ b/exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
@@ -1,0 +1,27 @@
+require 'generator/exercise_case'
+
+class AtbashCipherCase < Generator::ExerciseCase
+  def workload
+    case property
+    when 'encode' then encode_workload
+    when 'decode' then decode_workload
+    else raise 'unexpected property encountered'
+    end
+  end
+
+  def encode_workload
+    [
+      "plaintext = '#{phrase}'",
+      "ciphertext = '#{expected}'",
+      "assert_equal ciphertext, AtbashCipher.encode(plaintext)"
+    ].join("\n")
+  end
+
+  def decode_workload
+    [
+      "ciphertext = '#{phrase}'",
+      "plaintext = '#{expected}'",
+      "assert_equal plaintext, AtbashCipher.decode(ciphertext)"
+    ].join("\n")
+  end
+end
--- a/exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
+++ b/exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
@@ -14,7 +14,7 @@ class AtbashCipherCase < Generator::ExerciseCase
       "plaintext = '#{phrase}'",
       "ciphertext = '#{expected}'",
       "assert_equal ciphertext, Atbash.encode(plaintext)"
-    ].join("\n")
+    ]
   end
 
   def decode_workload
@@ -22,6 +22,6 @@ class AtbashCipherCase < Generator::ExerciseCase
       "ciphertext = '#{phrase}'",
       "plaintext = '#{expected}'",
       "assert_equal plaintext, Atbash.decode(ciphertext)"
-    ].join("\n")
+    ]
   end
 end
--- a/exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
+++ b/exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
@@ -13,7 +13,7 @@ class AtbashCipherCase < Generator::ExerciseCase
     [
       "plaintext = '#{phrase}'",
       "ciphertext = '#{expected}'",
-      "assert_equal ciphertext, AtbashCipher.encode(plaintext)"
+      "assert_equal ciphertext, Atbash.encode(plaintext)"
     ].join("\n")
   end
 
@@ -21,7 +21,7 @@ class AtbashCipherCase < Generator::ExerciseCase
     [
       "ciphertext = '#{phrase}'",
       "plaintext = '#{expected}'",
-      "assert_equal plaintext, AtbashCipher.decode(ciphertext)"
+      "assert_equal plaintext, Atbash.decode(ciphertext)"
     ].join("\n")
   end
 end
--- a/exercises/atbash-cipher/.meta/solutions/atbash_cipher.rb
+++ b/exercises/atbash-cipher/.meta/solutions/atbash_cipher.rb
@@ -1,4 +1,4 @@
-class AtbashCipher
+class Atbash
   def self.encode(plaintext)
     new(plaintext).encode
   end
@@ -14,25 +14,25 @@ class AtbashCipher
   end
 
   def encode
-    chunk convert(normalize(text))
+    chunk(convert)
   end
 
   def decode
-    convert(normalize(text))
+    convert
   end
 
   private
 
-  def convert(s)
-    s.tr(alphabet, key)
+  def convert
+    normalize.tr(alphabet, key)
   end
 
   def chunk(s)
     s.scan(/.{1,5}/).join(' ')
   end
 
-  def normalize(s)
-    s.downcase.gsub(/[^a-z0-9]/, '')
+  def normalize
+    text.downcase.gsub(/[^a-z0-9]/, '')
   end
 
   def alphabet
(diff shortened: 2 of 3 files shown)


exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
```diff
@@ -0,0 +1,27 @@
...
+      "plaintext = '#{phrase}'",
+      "ciphertext = '#{expected}'",
+      "assert_equal ciphertext, AtbashCipher.encode(plaintext)"
+    ].join("\n")
+  end
+
+  def decode_workload
+    [
+      "ciphertext = '#{phrase}'",
+      "plaintext = '#{expected}'",
```
guygastineau: Changing the order on these variables for the two sections keeps it in line with the structure from the `affine-cipher` tests.  I think it also is nice for the student to see the variables appear in an order consistent with the variables' relationships to the `assert_equal` call.

exercises/atbash-cipher/.meta/solutions/atbash_cipher.rb
```diff
@@ -1,16 +1,24 @@
-class Atbash
+class AtbashCipher
```
guygastineau: Didn't follow the proper convention before.

exercises/atbash-cipher/.meta/solutions/atbash_cipher.rb
```diff
@@ -1,16 +1,24 @@
...
   def self.encode(plaintext)
     new(plaintext).encode
   end
 
-  attr_reader :plaintext
+  def self.decode(ciphertext)
+    new(ciphertext).decode
+  end
+
+  attr_reader :text
```
guygastineau: Given that `decode` is certainly not passed un-enciphered text I thought it would be better to reduce this instance variable's name down to the common base of `plaintext` and `ciphertext`.

exercises/atbash-cipher/.meta/solutions/atbash_cipher.rb
```diff
@@ -1,16 +1,24 @@
...
+    @text = text
   end
 
   def encode
-    chunk convert(normalize(plaintext))
+    chunk convert(normalize(text))
+  end
+
+  def decode
+    convert(normalize(text))
```
guygastineau: It smells like `convert(normalize(text))` would be better off stored in a method, or `convert` could be `normalize(text).tr(alphabet, key)`, so that `encode` could be `chunk convert` and `decode` could be `convert`.

Thoughts?

I just didn't want to make drastic changes unless you all thought they were necessary :smile:

exercises/atbash-cipher/atbash_cipher_test.rb
```diff
@@ -1,46 +1,103 @@
...
 
-class AtbashTest < Minitest::Test
-  def test_encode_no
-    assert_equal 'ml', Atbash.encode('no')
+# Common test data version: 1.2.0 d5238bd
+class AtbashCipherTest < Minitest::Test
+  def test_encode_yes
+    # skip
+    plaintext = 'yes'
+    ciphertext = 'bvh'
```
guygastineau: For these short values using variables seems like overkill, but I assume a cleaner `generator` file is more important.

exercises/atbash-cipher/.meta/solutions/atbash_cipher.rb
kotp: I thought `Atbash` is fine, as it is a cipher, Like saying `GuygastineauPerson` when we already know you are a person or `MiamiCity` when we know it is a city.

guygastineau: I was under the impression that it is a hard requirement of this track for class names to be the CamelCase equivalent of the exercise slugs.

At least that is what was said in the `armstrong-numbers` exercise when it was brought up that `ArmstrongNumber` would fit better...

kotp: It is a style requirement that it is CameCase, but I wasn't positive that it had to exactly match the slug... though I was aware that it usually does.

kotp: Confirmed that it doesn't have to, it will be delivered, etc.  But if we want to, that is fine as well.  No strong feeling about it being a perfect match, and I think it is named well as it is.

Anyone else have strong feelings about this change either way?

An example of where this may change in the future may be the "Grains" method, as well as potentially "Series".  The slug may not end up matching the class name, or even the file name(s).  The slug identifies the exercise...

guygastineau: So, it should stay as is in the PR or revert to Atbash?

kotp: I would leave it as it was... (there wasn't a request to change it, and it won't be confusing when I update the exercise and my program doesn't fail all the tests due to the update).

guygastineau: Cool. That is what I thought you meant.

What do you think about my suggestions for changing `#convert`?

kotp: It is fine.  No strong opinion on the change, except that it doesn't belong in this PR.  The changes should be only to let the solution pass the tests as they were generated.  Other changes aren't really relevant to the PR.

They can happen here, but I also don't have strong feelings about the example solutions being anything other than a proof that the exercise can be written in the track.  It doesn't have to be exemplary, as that isn't its purpose.

guygastineau: Right, well I'm sorry I did it in my last commit.  I opened the solution to revert the class name, and my OCD just got the better of me.

guygastineau: Emacs EnhRuby mode was freaking out about the 1 letter argument names too :rofl:

kotp: I reread the first message in the PR, and the intention is to complete the solution as well, as it was "half implemented" so the solution does belong here, though the scope of the PR has grown, since I thought you were only doing the generator, rather than "also the half implementation".

Since the title was only implemented generator, that is what is persisted at the top (title) of the page viewer, and that is what my review was based on.

exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
Insti: this is fine.

exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
```diff
@@ -0,0 +1,27 @@
...
+    else raise 'unexpected property encountered'
+    end
+  end
+
+  def encode_workload
+    [
+      "plaintext = '#{phrase}'",
+      "ciphertext = '#{expected}'",
+      "assert_equal ciphertext, Atbash.encode(plaintext)"
+    ].join("\n")
```
Insti: I think `workload` can return an array of lines, so you don't need to join these yourself.

guygastineau: Really? Cool. I'll get to it.

guygastineau:
I hope I didn't eat up all the Travis build time with the frequent commits I've had to my various PR's.

exercises/atbash-cipher/.meta/generator/atbash_cipher_case.rb
guygastineau: Got it.

kotp:
@kytrinyx Any ideas on how to get the CI to reset?  

travis-ci_expected

kytrinyx:
Sorry, my bad! I enabled third party restrictions for Exercism, and thought that any previously installed and configured apps would keep access. In retrospect this was a silly assumption. I've enabled Travis now. I think you're going to need to kick the tires by pushing to the PR again, unfortunately, because Travis wasn't actually told about this build.

guygastineau:
I just realized that all solutions are now supposed to have stubs.  That might be outside the scope of this PR, but if it is acceptable to add one, then that is a way to add a meaningful commit for `licking the tires` on Travis.

guygastineau:
I could also amend my last commit message and force push to do this.

kotp:
> I think you're going to need to kick the tires by pushing to the PR again

I might be able to just tell Travis to run the job again... doing that now.

kotp:
And whatever you do, don't be licking the tires.

It worked... now that it can see the status.  I will do the same for some others.

PR #991, opened 2019-08-09
needs-a-genny: maintenance script

guygastineau:
I made this script to report which exercises still need a gen(ny)erator.

Let me know if you all want it as a tool.

PS. I just wanted to include minimal functionality.  If folks would prefer that it print the number, or do other things too I am more than willing change this to incorporate features.  All suggestions welcome :smile:


bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+genny_path = ->(slug) {
+  ['exercises', slug, '.meta', 'generator'].join('/')
+}
+
+genny_path_exists = ->(slug) { !Dir.exist?(genny_path[slug]) }
+
+puts Dir.children('exercises')
+       .filter(&genny_path_exists).sort.join("\n")
```
kotp: Indentation here, but otherwise good to go.

bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+genny_path = ->(slug) {
+  ['exercises', slug, '.meta', 'generator'].join('/')
+}
+
+genny_path_exists = ->(slug) { !Dir.exist?(genny_path[slug]) }
```
Insti: This method name is misleading, since you seem to be checking if the path does not exist.

bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+genny_path = ->(slug) {
+  ['exercises', slug, '.meta', 'generator'].join('/')
```
Insti: I'd probably use [File#join](https://ruby-doc.org/core-2.5.0/File.html#method-c-join) here

bin/needs-a-genny
kotp: I looked at that 3 times and missed this, definitely correct.

bin/needs-a-genny
guygastineau: I wasn't sure if it should indent relative to `Dir` or `puts`. Thanks for the clarification.

bin/needs-a-genny
guygastineau: Will do :smile:

guygastineau:
@Insti  actually no (about JS)

I know JS, but I've been spending a lot of time recently on Rust and Haskell.  Ramping up in Ruby now, so I figured I should help out here.

I don't actually mind defining these with `def` if you all would prefer (they were just so short).

bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+genny_path = ->(slug) {
```
petertseng: since it seems that this file has chosen to elide parentheses where possible (according to lines 4 and 7), it seems that `(slug)` should become `slug` instead.

bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
...
+  File.join 'exercises', slug, '.meta', 'generator'
+}
+
+no_genny_path = ->(slug) { !Dir.exist? genny_path[slug] }
```
petertseng: since it seems that this file has chosen to elide parentheses where possible (according to lines 4 and 7), it seems that `(slug)` should become `slug` instead.

bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
...
+
+puts Dir.children('exercises')
+  .filter(&no_genny_path).sort.join("\n")
```
petertseng: since `puts` with an array will already print each element on a separate line, the `join("\n")` may be elided if that is desired.

bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+genny_path = ->(slug) {
+  File.join 'exercises', slug, '.meta', 'generator'
+}
+
+no_genny_path = ->(slug) { !Dir.exist? genny_path[slug] }
+
+puts Dir.children('exercises')
```
petertseng: `exercises` means this script is only runnable from the top level of this repo, which is inconvenient to force someone to only be able to run it from one location.

thus, it may be preferable to use `File.join(__dir__, '..', 'exercises')` instead so that the script is runnable from anywhere.

bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+genny_path = ->(slug) {
+  File.join 'exercises', slug, '.meta', 'generator'
```
petertseng: here, too same comment https://github.com/exercism/ruby/pull/991#discussion_r312620485 about `exercises` forcing only being able to run it from one directory

bin/needs-a-genny
guygastineau: I didn't realize I could get `File.join` to do that. Question though: can the other tools in `bin/` be successfully run anywhere? If so then this should definitely conform to them.

bin/needs-a-genny
guygastineau: Good to know.

guygastineau:
The ruby-track already requires that the problem-specifications repository be cloned parallel to this repository.  Therefore, I can make this script check there for the existence of `canonical-data.json` .

Thanks for the great idea.

guygastineau:
Seems like Travis is broken...

guygastineau:
@Insti  @kotp If you all are in support of the extended functionality proposed by @petertseng I can implement that too.

I am actively working on generators for multiple exercises, and it would be nice to use this script without having to change branches everytime.  Are you amenable to merging with current functionality and creating an issue to extend it?

Also, I will add the parentheses back in I guess.  I like how it looks without them on calls, but I really prefer in the stabby-lambda syntax. Anyway, maybe another commit will get Travis to come back to life :smile:

bin/needs-a-genny
guygastineau: Are you saying that `File.join` will magically find `exercises/` by including `'..'`, or would it make it simply go to the parent like in shell?

bin/needs-a-genny
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+genny_path = ->(slug) {
+  File.join('exercises', slug, '.meta', 'generator')
+}
+
+no_genny_path = ->(slug) { !Dir.exist?(genny_path[slug]) }
+
+puts Dir.children('exercises')
+  .filter(&no_genny_path).sort
```
Insti: Methods with negatively phrased names can be confusing.

What about `reject`ing rather than `select`/`filter`ing ?

`reject(&generator_exists?)`

Insti:
>  it would be nice to use this script without having to change branches everytime

You can merge it in to your work in progress branches and then `rebase` the merge PR out again when you're done. (before pushing the generator branch.)

guygastineau:
Good idea about the rebase.

bin/needs-a-genny
guygastineau: I was thinking there must be some method that would do the opposite of filter :smile:

Let me make that change before merging.

guygastineau: I wanted a `?` in the filename too, but unfortunately it won't let me do that for a lambda, because it is illegal in variable names :frowning_face: 

I can't figure out how to pass a method defined with `def` in though.  `filter`/`reject` won't pass arguments to it.

bin/needs-a-genny
kotp: >can the other tools in bin/ be successfully run anywhere? If so then this should definitely conform to them.

`generate` will run from any folder, for example.  `local-status-check` will not, so it varies.

`setup-git-hooks` will not run from just anywhere, but it may make sense to be in the repository root while running this, at least minimally to be in the project folder somewhere.

kotp:
> @Insti @kotp If you all are in support of the extended functionality proposed by @petertseng I can implement that too.

Yes, the extended functionality is good.

> I am actively working on generators for multiple exercises, and it would be nice to use this script without having to change branches everytime. Are you amenable to merging with current functionality and creating an issue to extend it?

Yes, get it in so we can use it.  We can always expand on it.

> Also, I will add the parentheses back in I guess. I like how it looks without them on calls, but I really prefer in the stabby-lambda syntax. Anyway, maybe another commit will get Travis to come back to life

I don't mind, either way, the problem that was pointed out was inconsistency of style, I believe.  I don't care, as long as it is consistent.  Minimally, consistency in the same file, ideally, consistency within all of the tooling.

bin/needs-a-genny
kotp: You can pass methods as objects, if you make the method an object, with `method`.

guygastineau: I guess I would prefer not having to wrap it like `method(:generator?)` while passing to `reject`.  If I save it in a variable, then it once again will not have `?` on the end when passed.

I guess I think it is pretty clear now.

bin/needs-a-genny
petertseng: There is noting magical about File.join and `..`. `File.join(__dir__, '..', 'exercises')` will evaluate to someting like `"/path/to/repo/bin/../exercises"`, and `Dir.exercises("/path/to/repo/bin/../exercises")` will do as expected.

petertseng: There is nothing magical about File.join and `..`. `File.join(__dir__, '..', 'exercises')` will evaluate to something like `"/path/to/repo/bin/../exercises"`, and `Dir.children("/path/to/repo/bin/../exercises")` will do as expected.

guygastineau: I tested it in irb, and it treats `..` just like bash would.  That is, if I do it from the repo root it makes a non-existent path string.  It seems like there is a way to do it, but I don't think the method you are suggesting works from anywhere (only if you are 1 directory deeper than the root).

guygastineau: I needed `File.expand_path('../../exercises', __FILE__)`

guygastineau: @petertseng FYI, `__dir__` seems to evaluate to `.` in my repl.

petertseng: > `__dir__` seems to evaluate to . in my repl.

that seems correct.

bin/needs-a-genny
guygastineau: And thank you for your suggestions.

bin/needs-a-genny
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env ruby
 
+EXERCISES = File.expand_path('../../exercises', __FILE__)
```
petertseng: a reminder that this is equivalent to the following:

```ruby
EXERCISES = File.expand_path('../exercises', __dir__)
```

a suggestion that it is better to use this because the extra set of `..` can otherwise be confusing. If you will allow the use of the logical fallacy called "appeal to authority", note that rails did this in https://github.com/rails/rails/commit/5b8738c2df003a96f0e490c43559747618d10f5f, citing that using `__FILE__` hurts the person's soul and `__dir__` was done to "restore order"

bin/needs-a-genny
kotp: Since technically this line is evaluated as:

`puts(Dir.children('exercises').filter().sort.join())`  The dots should be aligned together, it is inclusive of the argument given, not at the level of puts.

```suggestion
        .filter(&genny_path_exists).sort
```

bin/needs-a-genny
kotp:
```suggestion
EXERCISES = File.expand_path('../exercises', __dir__)
```

bin/needs-a-genny
```diff
@@ -0,0 +1,12 @@
...
+
+genny_path = ->(slug) {
+  File.join(EXERCISES, slug, '.meta', 'generator')
+}
+
+genny_path_exists = ->(slug) { Dir.exist?(genny_path[slug]) }
+
+puts Dir.children(EXERCISES)
+  .reject(&genny_path_exists).sort
```
kotp:
```suggestion
        .reject(&genny_path_exists).sort
```

kotp: The argument given to `puts` might be better named as to what it represents.
`puts doohicky` where "doohicky" is a good name for whatever that argument to puts is doing.

bin/needs-a-genny
guygastineau: Thank you. Not a reminder, but rather the first time I've seen this. I sure wish the docs for File highlighted this feature...

guygastineau: I believe it is 'appeal to irrelevant authority'.  In this case your authority is not irrelevant to code, but they are probably not a relevant authority on the soul (but who knows... porbably depends on how much coffee they drink) :smiling_imp:

guygastineau: PS. I was very confused by the double `..`, so I much prefer this option.

guygastineau: Implemented.

bin/needs-a-genny
guygastineau: gave it a name

guygastineau:
You are probably right.

F3PiX:
> While the current name is cute, generators_needed may be a better file name.

In case it helps (and please ignore otherwise:)
How about `needs_a_generator`? To me `needs_a_*` feels more actionable than `needed`. Because (non native speaker here) now I'm thinking: `used to need one OR is in need of`?

kotp:
As a command from the terminal, I feel like "generators" works here, similar to how "git-*" is the pattern for git commands.  This is intended as a generator tool.  If something is needed, the "is" indicates the current need, if something was needed, "was" indicates the past need.  This tool indicates the current need... and so "generators-needed" is a good name, I believe, as it uses both forms, the tooling that is used, and the inquiry/command.   The dash may be a good idea here, though the underscore works as well to make it clear that it isn't a command/flag pair.

guygastineau:
I forgot to do this the other day 😅

I think I'll go with `generators-needes` (though I appreciate your perspective @F3PiX 😁).

I will use kebab-case so that this confirms with the filename convention already present in `bin/`.

guygastineau:
Made the changes and rebased it onto the new master updates.  fixedup and squashed some of the commit (there was no reason to include my initial work in shell ;) )

Let me know if everything looks good.

PS. I know we changed the name, but I didn't change the abbreviation `genny` inside the script for anything.  Let me know if this is okay.  I think it is a common enough abbreviation that no one would thi9nk it is talking about someone named `Genny` :smiling_imp:

guygastineau:
Made the changes and rebased it onto the new master updates.  fixedup and squashed some of the commit (there was no reason to include my initial work in shell ;) )

Let me know if everything looks good.

PS. I know we changed the name, but I didn't change the abbreviation `genny` inside the script for anything.  Let me know if this is okay.  I think it is a common enough abbreviation that no one would thi9nk it is talking about someone named `Genny` :smi

Insti:
Thanks for your work on this @guygastineau :heart:

PR #995, opened 2019-08-25
resistor-color-trio: New exercise

### LINKED ISSUE #979: Implement new exercise: Resistor Color Trio
Resistor Color Trio is [ready!](https://github.com/exercism/problem-specifications/tree/master/exercises/resistor-color-trio) now. 

Let's implement it in the Ruby track. 

See the ProbSpecs for the meaning of the 3 bands. 
The canonical data are intentionally generic, listing only the `value` that is calculated from 3 bands,  and the `unit`, which can be either 'ohms' or 'kiloohms'.

For Ruby, and because of the Track Anatomy and the position of the exercise in the track, there's a few requirements:
-  the student should use overriding `to_s`, without revealing that in the tests; so the tests should test the `label`.
-  the `label` method shouldn't  just return  `to_s`. Ideally, the label would return something like `"Resistor value: #{to_s}`
- It would be nice to add error handling. For instance an Argument Error for an input with less then 3 color bands. (Not provided in the canonical data.)
- In general, the student should take care of organising the code (i.e. separating the calculations of the value and returning the label) themselves, without the tests leading them. So, we don't want tests for an intermediate `value` method. 

This is still a mid level exercise, so we also don't want to force students to use multiple classes or mixins. (That will be on a later level, with a later Resistor Exercise.)

pgaspar:
Closes #979. I was waiting for https://github.com/exercism/problem-specifications/pull/1568 to be merged and didn't see the notification when it got merged a week ago. See the canonical data [here](https://github.com/exercism/problem-specifications/blob/master/exercises/resistor-color-trio/canonical-data.json).

The solution is based on code @F3PiX kindly gave me.

**TODO:**

- [x] Add to config.json

### DIFF
--- a/config.json
+++ b/config.json
@@ -1208,6 +1208,16 @@
       "topics": [
         "arrays"
       ]
+    },
+    {
+      "slug": "resistor-color-trio",
+      "uuid": "2df14b68-75c8-4984-8ae2-ecd2cb7ed1d4",
+      "core": false,
+      "unlocked_by": "high-scores",
+      "difficulty": 4,
+      "topics": [
+        "loops"
+      ]
     }
   ]
 }
--- a/exercises/resistor-color-trio/README.md
+++ b/exercises/resistor-color-trio/README.md
@@ -4,21 +4,12 @@ If you want to build something using a Raspberry Pi, you'll probably use _resist
 
 - Each resistor has a resistance value.
 - Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
-  To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values.
+  To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. Each band has a position and a numeric value. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
 - Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
   In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take 3 colors as input, and outputs the correct value, in ohms.
-  The color bands are encoded as follows:
+  The colors are mapped to the numbers from 0 to 9 in the sequence:
 
-* Black: 0
-* Brown: 1
-* Red: 2
-* Orange: 3
-* Yellow: 4
-* Green: 5
-* Blue: 6
-* Violet: 7
-* Grey: 8
-* White: 9
+Black - Brown - Red - Orange - Yellow - Green - Blue - Violet - Grey - White
 
 In `resistor-color duo` you decoded the first two colors. For instance: orange-orange got the main value `33`.
 The third color stands for how many zeros need to be added to the main value. The main value plus the zeros gives us a value in ohms.
--- a/exercises/resistor-color-trio/README.md
+++ b/exercises/resistor-color-trio/README.md
@@ -1,0 +1,75 @@
+# Resistor Color Trio
+
+If you want to build something using a Raspberry Pi, you'll probably use _resistors_. For this exercise, you need to know only three things about them:
+
+- Each resistor has a resistance value.
+- Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
+  To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values.
+- Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
+  In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take 3 colors as input, and outputs the correct value, in ohms.
+  The color bands are encoded as follows:
+
+* Black: 0
+* Brown: 1
+* Red: 2
+* Orange: 3
+* Yellow: 4
+* Green: 5
+* Blue: 6
+* Violet: 7
+* Grey: 8
+* White: 9
+
+In `resistor-color duo` you decoded the first two colors. For instance: orange-orange got the main value `33`.
+The third color stands for how many zeros need to be added to the main value. The main value plus the zeros gives us a value in ohms.
+For the exercise it doesn't matter what ohms really are.
+For example:
+
+- orange-orange-black would be 33 and no zeros, which becomes 33 ohms.
+- orange-orange-red would be 33 and 2 zeros, which becomes 3300 ohms.
+- orange-orange-orange would be 33 and 3 zeros, which becomes 33000 ohms.
+
+(If Math is your thing, you may want to think of the zeros as exponents of 10. If Math is not your thing, go with the zeros. It really is the same thing, just in plain English instead of Math lingo.)
+
+This exercise is about translating the colors into a label:
+
+> "... ohms"
+
+So an input of `"orange", "orange", "black"` should return:
+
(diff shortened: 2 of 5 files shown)


pgaspar:
@F3PiX where do we want to put this exercise on config.json?

F3PiX:
Awesome Pedro! I'm looking into it now.

exercises/resistor-color-trio/README.md
```diff
@@ -0,0 +1,75 @@
+# Resistor Color Trio
+
+If you want to build something using a Raspberry Pi, you'll probably use _resistors_. For this exercise, you need to know only three things about them:
+
+- Each resistor has a resistance value.
+- Resistors are small - so small in fact that if you printed the resistance value on them, it would be hard to read.
+  To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values.
```
F3PiX:
```suggestion
  To get around this problem, manufacturers print color-coded bands onto the resistors to denote their resistance values. Each band has a position and a numeric value. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
```

exercises/resistor-color-trio/README.md
```diff
@@ -0,0 +1,75 @@
...
+- Each band acts as a digit of a number. For example, if they printed a brown band (value 1) followed by a green band (value 5), it would translate to the number 15.
+  In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take 3 colors as input, and outputs the correct value, in ohms.
```
F3PiX:
```suggestion
  In this exercise, you are going to create a helpful program so that you don't have to remember the values of the bands. The program will take 3 colors as input, and outputs a label with the correct value, in ohms.
```

exercises/resistor-color-trio/README.md
```diff
@@ -0,0 +1,75 @@
...
+  The color bands are encoded as follows:
```
F3PiX:
```suggestion
  The colors are mapped to the numbers from 0 to 9 in the sequence:

 Black - Brown - Red - Orange - Yellow - Green - Blue - Violet - Grey - White
- Black: 0	
```
(instead of the list below)

exercises/resistor-color-trio/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+- orange-orange-black would be 33 and no zeros, which becomes 33 ohms.
+- orange-orange-red would be 33 and 2 zeros, which becomes 3300 ohms.
+- orange-orange-orange would be 33 and 3 zeros, which becomes 33000 ohms.
+
+(If Math is your thing, you may want to think of the zeros as exponents of 10. If Math is not your thing, go with the zeros. It really is the same thing, just in plain English instead of Math lingo.)
+
+This exercise is about translating the colors into a label:
+
+> "... ohms"
```
F3PiX:
```suggestion
> "Resistor value: ... ohms"
```

exercises/resistor-color-trio/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+So an input of `"orange", "orange", "black"` should return:
+
+> "33 ohms"
+
+When we get more than a thousand ohms, we say "kiloohms". That's similar to saying "kilometer" for 1000 meters, and "kilograms" for 1000 grams.
+So an input of `"orange", "orange", "orange"` should return:
+
+> "33 kiloohms"
```
F3PiX:
```suggestion
> "Resistor value: 33 kiloohms"
```

exercises/resistor-color-trio/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+(If Math is your thing, you may want to think of the zeros as exponents of 10. If Math is not your thing, go with the zeros. It really is the same thing, just in plain English instead of Math lingo.)
+
+This exercise is about translating the colors into a label:
+
+> "... ohms"
+
+So an input of `"orange", "orange", "black"` should return:
+
+> "33 ohms"
```
F3PiX:
```suggestion
> "Resistor value: 33 ohms"
```

pgaspar:
@F3PiX I would prefer option 2 - waiting for the new description to be merged. Here's what we should strive for, IMO: the instructions are generated from the canonical readme + any hint we add to the hints file. Ideally that alone would be enough.

Changing the description manually will probably mean that someone generates it from the canonical data data in the future and the manually added changes will be lost - so we should avoid them if possible.

F3PiX:
> Changing the description manually will mean that when someone regenerates it from the canonical data in the future, the manually added changes will be lost - so we should avoid them if possible.

In general, I agree, but in this case I updated the description in the same way it will be in the new ProbSpec description. So, it's meant to be overwritten by the ProbSpec version. In this case, it's a good thing: if things change in the final version on ProbSpec, we just regen instead of update the manual changes separately.

pgaspar:
> In general, I agree, but in this case I updated the description in the same way it will be in the new ProbSpec description. So, it's meant to be overwritten by the ProbSpec version. In this case, it's a good thing: if things change in the final version on ProbSpec, we just regen instead of update the manual changes separately.

I agreed with @F3PiX that for now we'd only include README changes that weren't specific to this track's specific handling of the `label` method. The canonical `label` returns a `{ value: 100, unit: 'ohms' }` hash, whilst the current Ruby track approach is to return a string: `Resistor value: 100 ohms`. In summary, the only changes we're including in the README are changes that are going to be done in the canonical instructions file. Historically, track-specific instructions are part of the `hints.md` file, and we'll try to do the same here.

> In the config.json, the Trio can go at the end of the file, directly behind Resistor Color.
>
> * core: false (it should be able to replace Matrix, but let's test it as a side first)
> * difficulty: 4
> * topics: loops

I made it `unlocked_by` high-scores - was that what you wanted?

I'm wondering why we're choosing `loops` as the topic 🤔 I can see some `text_formatting` and maybe some very light `math`?

exercises/resistor-color-trio/README.md
pgaspar: These changes are ruby-track specific and will not be part of the canonical description. They would be overwritten next time we regenerate the README. I'm leaving these out to be handled in the hints file.

exercises/resistor-color-trio/README.md
pgaspar: These changes are ruby-track specific and will not be part of the canonical description. They would be overwritten next time we regenerate the README. I'm leaving these out to be handled in the hints file.

exercises/resistor-color-trio/README.md
pgaspar: These changes are ruby-track specific and will not be part of the canonical description. They would be overwritten next time we regenerate the README. I'm leaving these out to be handled in the hints file.

exercises/resistor-color-trio/README.md
pgaspar: These changes are ruby-track specific and will not be part of the canonical description. They would be overwritten next time we regenerate the README. I'm leaving these out to be handled in the hints file.

config.json
```diff
@@ -1208,6 +1208,16 @@
       "topics": [
         "arrays"
       ]
+    },
+    {
+      "slug": "resistor-color-trio",
+      "uuid": "2df14b68-75c8-4984-8ae2-ecd2cb7ed1d4",
+      "core": false,
+      "unlocked_by": "high-scores",
```
F3PiX: Yes, great!

PR #997, opened 2019-08-31
Refactor columns method to use .transpose

luminousbeam:


kotp:
This seems to include more than just the title.  Commit fdd209b8 doesn't belong here.

kotp:
The example solutions aren't required to be idiomatic, as they are here for "proof of concept" more than anything else, never delivered to the student, only to validate the tests.  That being said, the change looks good.

luminousbeam:
> This seems to include more than just the title. Commit [fdd209b](https://github.com/exercism/ruby/commit/fdd209b8df7d998579b07ec53f4e370024ba2bfe) doesn't belong here.

What should I do here? 1) Close the PR 2) Remove my space-age changes 3) Tweak and edit my changes to space-age?

kotp:
Probably remove the commit that effects the other.  Let's say that the refactor on one exercise is perfect, and we want it.  The other needs changes.  This would delay being able to merge in the good, because of an unrelated bad (or unwanted, not even necessarily bad) change.

PR #999, opened 2019-09-01
Refactor Matrix columns solution

luminousbeam:


kotp:
Looks good.  I think we may have decided not to make this change some years ago because it is somewhat idiomatic. (If someone looks for a solution to use we can't really tell that they copy pasted it from here.)  But if there aren't any objections from the mentor team, I don't have any objection other than that.

These solutions are here mostly to "prove the exercise" not to provide idiomatic solutions.

luminousbeam:
That makes sense. I suppose that I have seen a lot of one liner type solutions using the most suitable methods available, which is why I used something like .transpose for this solution.

kotp:
I believe this is probably one of the mentor notes that should exist as an "approved this solution" type thing, if you want to check there.

PR #1000, opened 2019-09-04
Move Resistor Trio up in the track

F3PiX:
Trio seems to be too hard on its current position (start Level 2). So, not much fun to mentor, and struggling students.

As a side after HighScores, the gap is too big. In HighScores, the design of the class is forced by the tests. In Trio, a lots of decisions to make at once.


F3PiX:
Thanks Pedro!

Issue #1001, opened 2019-09-15
Scrabble score: test against nil forces in wrong direction


giulp:
The code,imho, should really not expect to be given 'nil' forceably, and should fail if that happens. 

Indeed, there should probably be some programming in there that purposefully raises an error. 

this test is misleading

```ruby

  def test_nil_scores_zero
    skip
    assert_equal 0, Scrabble.new(nil).score
  end

```

maybe it could be substituted with something like

```ruby

  def test_nothing_given_nothing_scores_zero
    skip
    assert_equal 0, Scrabble.new.score
  end

```

kotp:
I agree, the exercise is not about what to do when you are given bad data, and this is just bad data, in my opinion.

I am going to assign this to you, and label it.

kytrinyx:
This test disappeared the last time we synced with problem-specifications (see #1411).

PR #1002, opened 2019-09-16
Scrabble-score: Testing for nil is bad design

giulp:
The user of the ScrabbleScore library should be responsible and
responding if they hand bad data to the library.  We want to give a
reasonable "nothing given" response in the library, which intuitively
seems to be a score of 0.

closes: #1001

### DIFF
--- a/exercises/scrabble-score/.meta/solutions/scrabble_score.rb
+++ b/exercises/scrabble-score/.meta/solutions/scrabble_score.rb
@@ -4,7 +4,7 @@ class Scrabble
   end
 
   attr_reader :term
-  def initialize(term)
+  def initialize(term = '')
     @term = term.to_s.downcase
   end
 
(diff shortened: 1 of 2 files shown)


geeknees:
Hello, I am a student of exercsim.

I agree with removing test for nil. In my experience, I just add like `term ||= ''`. This is not good for handing invalid data.

In addition to it how about adding test for passing letters that is not include Letter Values. For example, `mid-summer` is not allowed scrabble, but we can pass it.

IMO, this test case expects 14 is fine, but some people may say this should raise errors. Please ignore this if this is off-topic.

Thanks.

kotp:
@geekness Thank you for your input.  I think we have decided that it may be better to accept the flexibility, dash being a 0 score regardless of discarding that character, or scoring literally any "nontile" thing a 0.  But I think this issue is strictly concerned with testing that suggests that we would accept a "nil" literally.

I hope you enjoyed the Scrabble exercise though!

geeknees:
@kotp Thank you! I really enjoyed exercise. Yes I feel nil for Scrabble is invalid data. Removing nil-test looks good to me. Otherwise nil should raise error. Thanks!

kotp:
I am going to close this.  Please create an issue if we wish to revisit this issue.

giulp:
@kotp thanks for the update about the decision ! have a nice day :)

PR #1007, opened 2019-09-23
Add tests for non-square matrix

goalaleo:
I mentored a solution which relied on the matrixes to be size 3x3 or 2x2. While this doesn't test matrixes of all sizes, it adds tests for a non-square matrix of size 2x3

Related solution:
https://exercism.io/mentor/solutions/c06d5e96459b407495564e6de2fbf6ec

### DIFF
--- a/exercises/matrix/matrix_test.rb
+++ b/exercises/matrix/matrix_test.rb
@@ -36,4 +36,16 @@ class MatrixTest < Minitest::Test
     matrix = Matrix.new("89 1903 3\n18 3 1\n9 4 800")
     assert_equal [1903, 3, 4], matrix.columns[1]
   end
+  
+  def test_asymmetric_matrix_rows
+    skip
+    matrix = Matrix.new("1 2\n3 4\n5 6")
+    assert_equal [1, 2], matrix.rows[0]
+  end
+  
+  def test_asymmetric_matrix_columns
+    skip
+    matrix = Matrix.new("1 2\n3 4\n5 6")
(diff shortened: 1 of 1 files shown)


kotp:
Do you think this should be elevated to canonical at some point?

goalaleo:
mm, probably? I can see this being relevant to all tracks, but to be honest
I'm not sure what the process is for adding new specifications.

Are all parts of the specifications shared between tracks? Can the specification be changed before the tests (for each track) are added?

kotp:
> mm, probably? I can see this being relevant to all tracks, but to be honest
> I'm not sure what the process is for adding new specifications.

To introduce the change we would [open an issue on problem-specifications](https://github.com/exercism/problem-specifications/issues) to talk about it, and once discussed, open a PR there that makes the change.


> Are all parts of the specifications shared between tracks? Can the specification be changed before the tests (for each track) are added?

The specifications are specifically to share between the tracks, and some tracks remove things that may not be applicable to their tracks, and like is done here, some tracks add some things that may be relevant to that specific track.

kotp:
To go a bit further, this track, has test generators that exist, that when set up, can use the specification data to generate the tests that we use.  Look at [generated test suites](https://github.com/exercism/ruby#generated-test-suites) which will get you started on what is used to generate the test, and looking at [implmementing a generator](https://github.com/exercism/ruby#implementing-a-generator) for information on how to do this.

goalaleo:
Thanks @kotp I'll take a look at those 👍

PR #1009, opened 2019-10-03
Binary search generator

fredrb:
- Add missing generator to `binary-search` as per #396.
- Fixes `binary_search.rb` according to canonical tests.

### DIFF
--- a/exercises/binary-search/.meta/generator/binary_search_case.rb
+++ b/exercises/binary-search/.meta/generator/binary_search_case.rb
@@ -1,0 +1,25 @@
+require 'generator/exercise_case'
+
+class BinarySearchCase < Generator::ExerciseCase
+  def workload
+    [binary_search, assertion].join
+  end
+
+  private
+
+  def binary_search
+    "binary = BinarySearch.new(#{array})\n"
+  end
+
+  def assertion
+    expected.is_a?(Numeric) ? standard_assertion : error_assertion
+  end
+
+  def standard_assertion
+    "assert_equal #{expected}, binary.search_for(#{value})"
+  end
+
+  def error_assertion
+    assert_equal(nil, "binary.search_for(#{value})")
+  end
+end
--- a/exercises/binary-search/.meta/solutions/binary_search.rb
+++ b/exercises/binary-search/.meta/solutions/binary_search.rb
@@ -2,21 +2,22 @@ class BinarySearch
   attr_reader :list
 
   def initialize(data)
-    fail ArgumentError unless data.sort == data
     @list = data
   end
 
   def search_for(datum)
+    return nil if list.empty?
     return middle if list[middle] == datum
 
     if list[middle] > datum
-      sublist = list[0..middle]
-      fail 'Not Found' if sublist == list
+      sublist = list[0...middle]
+      return nil if sublist == list
       return BinarySearch.new(sublist).search_for(datum)
     else
       sublist = list[middle..-1]
-      fail 'Not Found' if sublist == list
-      return BinarySearch.new(sublist).search_for(datum) + middle
+      return nil if sublist == list
+      result = BinarySearch.new(sublist).search_for(datum)
+      return result.nil? ? nil : result + middle
     end
   end
 
--- a/exercises/binary-search/binary_search_test.rb
+++ b/exercises/binary-search/binary_search_test.rb
@@ -1,49 +1,71 @@
 require 'minitest/autorun'
 require_relative 'binary_search'
 
+# Common test data version: 1.3.0 bfb218f
 class BinarySearchTest < Minitest::Test
-  def test_it_has_list_data
-    binary = BinarySearch.new([1, 3, 4, 6, 8, 9, 11])
-    assert_equal [1, 3, 4, 6, 8, 9, 11], binary.list
+  def test_finds_a_value_in_an_array_with_one_element
+    # skip
+    binary = BinarySearch.new([6])
+    assert_equal 0, binary.search_for(6)
   end
 
-  def test_it_raises_error_for_unsorted_list
-    skip
-    assert_raises ArgumentError do
-      BinarySearch.new([2, 1, 4, 3, 6])
-    end
-  end
-
-  def test_it_raises_error_for_data_not_in_list
-    skip
-    assert_raises RuntimeError do
-      BinarySearch.new([1, 3, 6]).search_for(2)
-    end
-  end
-
-  def test_it_finds_position_of_middle_item
+  def test_finds_a_value_in_the_middle_of_an_array
     skip
     binary = BinarySearch.new([1, 3, 4, 6, 8, 9, 11])
-    assert_equal 3, binary.middle
+    assert_equal 3, binary.search_for(6)
   end
 
-  def test_it_finds_position_of_search_data
+  def test_finds_a_value_at_the_beginning_of_an_array
     skip
     binary = BinarySearch.new([1, 3, 4, 6, 8, 9, 11])
-    assert_equal 5, binary.search_for(9)
+    assert_equal 0, binary.search_for(1)
   end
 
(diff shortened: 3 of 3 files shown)


exercises/binary-search/.meta/solutions/binary_search.rb
```diff
@@ -2,21 +2,21 @@ class BinarySearch
...
-  def search_for(datum)
-    return middle if list[middle] == datum
+  def search_for(value)
+    fail 'Not Found' if list.empty?
+    return middle if list[middle] == value
 
-    if list[middle] > datum
-      sublist = list[0..middle]
+    if list[middle] > value
+      sublist = list[0..middle-1]
```
kotp:
```suggestion
      sublist = list[0...middle]
```

exercises/binary-search/.meta/generator/binary_search_case.rb
```diff
@@ -0,0 +1,25 @@
...
+  def assertion
+    expected.is_a?(Fixnum) ? standard_assertion : error_assertion
+  end
+
+  def standard_assertion
+    "assert_equal #{expected}, binary.search_for(#{value})"
+  end
+
+  def error_assertion
+    assert_raises(RuntimeError, "binary.search_for(#{value})")
```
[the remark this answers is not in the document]
kotp: We may not want to raise an error... the specifications state:

> "comments": [
      "The error object is used to indicate that the value is not included in the array.",
      "It should be replaced with the respective expression that is idiomatic",
      "for the language that implements the tests.",

kotp:
@fredrb can we restore the `datum` name so that the diffs are more relevant to the generator being put in place?

kotp:
Thank you very much @fredrb for tackling this.  Not one of the simplest tasks to do, for sure!

exercises/binary-search/.meta/generator/binary_search_case.rb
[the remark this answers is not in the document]
fredrb: What would you recommend? I checked a few other implementations and [some return -1](https://github.com/exercism/javascript/blob/master/exercises/binary-search/example.js) and other [return value exception](https://github.com/exercism/python/blob/master/exercises/binary-search/example.py). 

`ArgumentError` might make sense, but then one could argue that the argument itself is valid.

fredrb:
@kotp thanks for the review. I reverted the `datum` to `value` change.

fredrb:
@kotp let me know if you want the commits squashed.

kotp:
The PR isn't quite close enough, I think to worry about squashing and rebasing interactively to build your history story yet.

exercises/binary-search/.meta/generator/binary_search_case.rb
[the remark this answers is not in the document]
kotp: I think we should return what makes sense for Ruby.

Given the description, I think the failure point is giving a collection that is not sorted, we should raise an error, as an argument error.  This should be the only error that we draw out for the tests.

Given the description in the Readme there is no indication of a failure for finding no value:

> If the remaining array to be searched is empty, then the key cannot be found in the array and a special "not found" indication is returned.

That special "not found" indication is probably `nil` for Ruby.   This fits in with idiomatic Ruby.  We get something truthy or falsey, the value found, or `nil`.  Further, this is expected "normal operation" rather than "exceptional behavior".  We shouldn't raise an exception for this, in my opinion.

exercises/binary-search/.meta/solutions/binary_search.rb
```diff
@@ -2,15 +2,15 @@ class BinarySearch
   attr_reader :list
 
   def initialize(data)
-    fail ArgumentError unless data.sort == data
```
kotp: I think this is meant to be here, since the entire idea of binary search doesn't work if the data is not sorted.

exercises/binary-search/.meta/solutions/binary_search.rb
```diff
@@ -2,15 +2,15 @@ class BinarySearch
   attr_reader :list
```
kotp: If `list` is not longer tested, this should probably be private, however, we are generating the exercise tests, the PR isn't about the "proof of exercise", so perhaps this is a different PR.

exercises/binary-search/binary_search_test.rb
```diff
@@ -1,49 +1,81 @@
...
+  end
+
+  def test_finds_a_value_in_the_middle_of_an_array
+    skip
     binary = BinarySearch.new([1, 3, 4, 6, 8, 9, 11])
-    assert_equal [1, 3, 4, 6, 8, 9, 11], binary.list
+    assert_equal 3, binary.search_for(6)
   end
 
-  def test_it_raises_error_for_unsorted_list
```
kotp: This test should still exist somewhere.

kotp: I take that back.  This test should not exist... 

>    "Following https://github.com/exercism/problem-specifications/issues/234 the exercise",
    "should NOT include checking whether the array is sorted as it defeats",
    "the point of the binary search.",

exercises/binary-search/.meta/solutions/binary_search.rb
kotp: Further inspection of the problem-specifications this should be gone... so this is a good change.

exercises/binary-search/.meta/generator/binary_search_case.rb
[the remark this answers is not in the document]
fredrb: Agreed that value not found is not "exceptional behavior". I'll make the necessary changes to return `nil`.

fredrb:
Sorry, had to force push cause I switched machines and git e-mail configuration was wrong. 

@kotp I changed the tests to expect `nil` other than `RuntimeError`. However the solution code had to change a bit. Since the original solution was returning the recursive call of `search_for` and it would *always* yield a valid output (i.e. a `Numeric` value) except when an exception is raised - this bit of the code worked fine:
```ruby
return BinarySearch.new(sublist).search_for(datum) + middle
```
Now that we are returning `nil` when value is not found, we get an exception that `nil` does not have `+` operation implemented. Which is expected.
The way I solved this (to avoid major changes on the example) was to simply check for `nil`:
```ruby
result = BinarySearch.new(sublist).search_for(datum)
return result.nil? ? nil : result + middle
```
I'm not entirely happy with this solution, specially cause we lost symmetry from left/right handling blocks.
Anyway, let me know what do you think.

exercises/binary-search/.meta/solutions/binary_search.rb
```diff
@@ -2,21 +2,22 @@ class BinarySearch
...
+      sublist = list[0...middle]
+      return nil if sublist == list
       return BinarySearch.new(sublist).search_for(datum)
     else
       sublist = list[middle..-1]
-      fail 'Not Found' if sublist == list
-      return BinarySearch.new(sublist).search_for(datum) + middle
+      return nil if sublist == list
+      result = BinarySearch.new(sublist).search_for(datum)
+      return result.nil? ? nil : result + middle
```
kotp: The example solution doesn't need to be idiomatic, or even clean.  Better ways to solve this will come out with students solving the new version... and if we see this solution, we may be able to realize the student has found it on github, and has submitted it, and so we can talk about the ways to improve this code with them at this time.  So I don't usually look too closely at the example solutions for this, as long as they do pass the tests.  The tests are more important to be exemplary Ruby, as that is delivered to the students.

kotp:
@fredrb I think now it is close enough to do an interactive rebase and formulate your commits how you want them to look.

Don't worry about force pushing to the PR branch, and changing history in this branch.  That's what it is here for.

fredrb:
Here you go @kotp - I've squashed all commits into one.

kotp:
Thank you for the work done on this @fredrb

PR #1012, opened 2019-11-04
Delete unnecessary instructions in Twelve Days test

### LINKED ISSUE #1005: Twelve Days not showing button to update exercise
The 'Twelve Days' exercise was updated [here](https://github.com/exercism/ruby/commit/a7b05fa3facf485efb7125f558eb0cdb61472ccc), however the button to update the exercise doesn't show up because the test file wasn't changed.

One possible solution would to make a trivial change to the test file as suggested by @iHiD [here](https://github.com/exercism/exercism/issues/4480#issuecomment-533922513),

PatrickMcSweeny:
Closes #1005. I figured this wording is better, and it should make the button to update the exercise visible.

### DIFF
--- a/exercises/twelve-days/twelve_days_test.rb
+++ b/exercises/twelve-days/twelve_days_test.rb
@@ -2,14 +2,6 @@ require 'minitest/autorun'
 require_relative 'twelve_days'
 
 class TwelveDaysTest < Minitest::Test
-  # This test is an acceptance test.
-  #
-  # If you find it difficult to work the problem with so much
-  # output, go ahead and add a `skip`, and write whatever
-  # unit tests will help you. Then unskip it again
-  # to make sure you got it right.
-  # There's no need to submit the tests you write, unless you
-  # specifically want feedback on them.
   def test_the_whole_song
     song_file = File.expand_path('../song.txt', __FILE__)
     expected = IO.read(song_file)


exercises/twelve-days/twelve_days_test.rb
```diff
@@ -5,10 +5,9 @@ class TwelveDaysTest < Minitest::Test
   # This test is an acceptance test.
   #
   # If you find it difficult to work the problem with so much
-  # output, go ahead and add a `skip`, and write whatever
-  # unit tests will help you. Then unskip it again
-  # to make sure you got it right.
-  # There's no need to submit the tests you write, unless you
+  # output, you can temporarily add a `skip` and write whatever
+  # unit tests will help you. There's no need
+  # to submit the tests you write, unless you
```
kotp: We can do even better, as there is only one test here.  And we really shouldn't suggest submitting the test file that is delivered to the student.  We can remove the instruction to skip the test, as it is not skipped, and is only one test.

What do you think?

PatrickMcSweeny: Sounds good to me. I pushed the change.

kotp: Thanks.  I am not sure it is the best solution, but it should achieve your immediate goal (well, to some definition of "immediate", sorry about the wait!).  The generator is likely where we need to make the change so that when generated this doesn't show back up.

Now that I am paying attention to your merge requests (on this exercise anyway) it should go faster, if you are interested.

PR #1014, opened 2019-12-08
stub_generator comes from Exercism tooling.

kotp:
This should be in the project for maintainers, but not version
controlled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ bin/configlet.exe
 coverage/
 .last_generator_run
 node_modules/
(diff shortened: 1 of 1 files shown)


kytrinyx:
@kotp Do we still have tooling that adds this stub_generator file? Doing a quick search, I couldn't find anything in `configlet`.

kotp:
I would have to look at the history of this file, but this has been waiting for 2.75 years and like teeth, if we ignore it, this may be the proof that it will probably go away.

Closing it.  Thanks for the review @kytrinyx !

PR #1016, opened 2019-12-11
Add editor settings to config.json

iHiD:
**(Updated from discussion)**

For our in-browser coding UI, we need need some settings that are track-specific. This is my first pass at what that might look like. 

There are two settings:
- `indent_style`: Should presses of the tab key add a tab character or x spaces. Can be "tab" or "space".
- `indent_size`: How far to indent when when you press the tab key.

So the net result of these two options in Ruby is that we add two spaces when someone presses tab, which I believe to be the normal default.

I'd appreciate feedback from @exercism/track-maintainers on the keys/structure, and also @exercism/ruby on the values for the Ruby track.

Note: I will probably merge this PR early, just so we can have *something* on master to develop against, but I don't intend that to be the end of the discussion, and would be happy to amend things afterwards if people disagree with my first pass.

### DIFF
--- a/config.json
+++ b/config.json
@@ -7,7 +7,7 @@
   "editor": {
     "tab_width": 2,
     "soft_tabs": true
-  }
+  },
   "exercises": [
     {
       "slug": "hello-world",
--- a/config.json
+++ b/config.json
@@ -4,6 +4,10 @@
   "blurb": "Ruby is a dynamic, open source programming language with a focus on simplicity and productivity. It has an elegant syntax that is natural to read and easy to write.",
   "gitter": "ruby",
   "solution_pattern": "[Ee]xample|\\.meta/solutions/[^/]*\\.rb",
+  "editor": {
+    "tab_width": 2,
+    "soft_tabs": true
+  }
   "exercises": [
     {
       "slug": "hello-world",
--- a/config.json
+++ b/config.json
@@ -4,9 +4,9 @@
   "blurb": "Ruby is a dynamic, open source programming language with a focus on simplicity and productivity. It has an elegant syntax that is natural to read and easy to write.",
   "gitter": "ruby",
   "solution_pattern": "[Ee]xample|\\.meta/solutions/[^/]*\\.rb",
-  "editor": {
-    "tab_width": 2,
-    "soft_tabs": true
+  "online_editor": {
+    "indent_style": "space",
+    "indent_size": 2
   },
   "exercises": [
     {


neenjaw:
From my perspective for elixir, as it follows ruby's style, those settings would be good too.  @devonestes and @Cohen-Carlisle may be able to offer more as they develop in elixir professionally.

kotp:
I would propose setting something for trailing white space as well.

config.json
```diff
@@ -4,6 +4,10 @@
   "blurb": "Ruby is a dynamic, open source programming language with a focus on simplicity and productivity. It has an elegant syntax that is natural to read and easy to write.",
   "gitter": "ruby",
   "solution_pattern": "[Ee]xample|\\.meta/solutions/[^/]*\\.rb",
+  "editor": {
+    "tab_width": 2,
+    "soft_tabs": true
```
ErikSchierboom: I had never heard of "soft" tabs. Is that a common term?

SleeplessByte: Yep. 

> Hard tabs are the ASCII tab character (9, or \t in many languages) that advance the cursor to the next tab stop. Soft tabs are the use of spaces to simulate this effect for indentation.

I've seen it in many many configuration / settings, but recently I've seen the much clearer "use spaces instead of tabs".

config.json
```diff
@@ -4,6 +4,10 @@
   "blurb": "Ruby is a dynamic, open source programming language with a focus on simplicity and productivity. It has an elegant syntax that is natural to read and easy to write.",
   "gitter": "ruby",
   "solution_pattern": "[Ee]xample|\\.meta/solutions/[^/]*\\.rb",
+  "editor": {
```
coriolinus: Perhaps this should be called `online_editor` or `recommended_editor_settings` or something. If it's just called `editor`, I expect to see at least a few bug reports per year surprised that this hasn't overridden someone's VS Code defaults.

iHiD:
> I would propose setting something for trailing white space as well.

I don't believe the online editor (Ace) we're using has any support for that.

eparovyshnaya:
It will be hard to form a set of settings applicable to any language we have, but these two are definitely make sense for any one.

devonestes:
I worry that defining "default" editor settings for folks is going to end up in nearly endless flamewars about what is the "right" way to do something. I'd just leave this up to the user entirely.

If you want to have something, then start with the absolute minimal configuration possible and add only if there's a really, truly compelling reason to do so. While I've seen Ruby indented with 4 spaces (or hard tabs), it's exceptionally rare so that might be a good place to start if you just want something to develop against.

SleeplessByte:
> I worry that defining "default" editor settings for folks is going to end up in nearly endless flamewars about what is the "right" way to do something. I'd just leave this up to the user entirely.

This. 

But I think it's fine to have _these_ as default, which you can then change in the UI. cc @iHiD

yawpitch:
As it’s the most common “standard” out there maybe we should lift the names and values of the properties from [editorconfig](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties)? The root key itself could be `interactive.editorconfig` or something similar.

Also to @devonestes point, since this is the config for the online interactive editor I don’t think we need to worry too much about flame wars, people will adapt. Also anyone coming to a new language with particular strong style preferences is likely missing the point of there being a per-track configuration.

Perhaps, long run, we might allow students to have and set their own preferences for these, if they truly feel the need to override the sensible defaults?

iHiD:
For clarity, in case I wasn't clear, this is for the online browsing UI. (@coriolinus I'll rename the key as you suggest to make this super-explicit)

I don't really intend to make this customisable - if people want to develop using their own settings, they can still use the CLI. I imagine it will be mainly newcomers to a language that choose to use in-browser coding environment, so what I think we need is a set of sensible defaults that you would teach a new user if they were coming to your language.

Re @yawpitch, I'd be happy with `indent_style` and `indent_size` as alternative keys taken from `editorconfig`?

SleeplessByte:
> I don't really intend to make this customisable - if people want to develop using their own settings, they can still use the CLI. I imagine it will be mainly newcomers to a language that choose to use in-browser coding environment, so what I think we need is a set of sensible defaults that you would teach a new user if they were coming to your language.

I can get behind that.

config.json
ErikSchierboom: Ah. I like "use spaces instead of tabs" better to be honest :)

iHiD: I think we use `indent_style` instead. "soft tabs" is a pretty standard phrase though :p

mpizenberg:
I don't know how common this is in other languages but the Elm community is pretty much using elm-format, a 0-config code formatter, usually triggered on save. It gives this warm feeling of seeing everything click together on save, very satisfying and comforting for newcomers. I think go has the same thing. Is this a feature embeddable inside an online editor? I'm also fine setting a default spacing config.

coriolinus:
Rust has the same thing, `rustfmt`... but I think autoformatting code is likely outside the scope of this PR. It seems likely that autoformatting will be like test running: each track will create a Docker container which can do the job according to some spec, and the editor will call into that container.

iHiD:
@kntsoriano The canonical keys to develop against are:
```
  "online_editor": {
    "indent_style": "space", # or "tab"
    "indent_size": 2
  }
```

sshine:
How did this PR pass `configlet fmt .` during CI?

iHiD:
@sshine It looks like there were no CI checks. Maybe CI isn't working properly in this repo?

sshine:
Okay, because it seems that performing a similar change in exercism/haskell#886 causes the `configlet fmt .` check to break on the Haskell track. This makes it hard to replicate the effort and preserve CI. Perhaps, along with config.json, `configlet` should be patched as well?

iHiD:
Yep. That would be good. One for @exercism/configlet :)

config.json
kotp: "expand tabs" is another one that seems common.

PR #1019, opened 2020-01-12
darts: Port Darts Exercise

Thrillberg:
This is my first exercise port! 👋 I think I followed the written instructions correctly, but one thing to note is that I, personally, came to doubt the educational value of this particular exercise as I worked through it. I'm submitting this PR in case others find it valuable but if someone wants to make an editorial decision to exclude this exercise, I would definitely not protest! 😄


Thrillberg:
Thanks for taking a look, @kotp! Didn't realize it required an entry in `config.json` but I'd be happy to add one. Taking a glance at it, I can probably figure something out about how it works but how would I go about generating the `uuid` for the exercise? Or is there documentation about this that I'm missing (sorry if so!)

In terms of where it fits in, I'm not sure. In the JavaScript track it's rated as a difficulty "3" but has no `topics` listed, if we want to use that as guidance.

kotp:
```
$ bin/configlet uuid --help
Generate a UUID.

Each Exercism exercise needs a unique and unmutable UUID. This needs to be
unique across the entire platform, not just within a track. In other words, if
you have 'clock' in Go and 'clock' in Haskell, they need to have different
UUIDs, even though they are based on the same problem specification.

Usage:
  bin/configlet uuid [flags]

Examples:
  bin/configlet uuid

Flags:
  -h, --help   help for uuid
```

That should help you out.  More information is [here, specifically the README.md](https://github.com/exercism/configlet).

Thrillberg:
Thanks for the guidance, @kotp! I've added an entry to `config.json`. I also noticed that CI is failing and I'm not sure if it's due to a change I made. If so, do you happen to have any ideas on what is the cause?

kotp:
> Thanks for the guidance, @kotp! I've added an entry to `config.json`. I also noticed that CI is failing and I'm not sure if it's due to a change I made. If so, do you happen to have any ideas on what is the cause?

```
Traceback (most recent call last):
	1: from /tmp/darts20200119-6390-w47l5w/darts_test.rb:2:in `<main>'
/tmp/darts20200119-6390-w47l5w/darts_test.rb:2:in `require_relative': /tmp/darts20200119-6390-w47l5w/darts.rb:22: syntax error, unexpected => (SyntaxError)
      10.0.. => 0,
             ^~
/tmp/darts20200119-6390-w47l5w/darts.rb:29: syntax error, unexpected end-of-input, expecting keyword_end
rake aborted!
Command failed with status (1): [/home/travis/.rvm/rubies/ruby-2.5.5/bin/ru...]
```
```ruby
      10.0.. => 0,
```
This line (22) is causing a problem with the version of Ruby being ran.  Ensure that the tests are Ruby 2.5 frienly.

exercises/darts/.meta/solutions/darts.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  private
+
+  def scores
+    {
+      0.0..1.0 => 10,
+      1.0..5.0 => 5,
+      5.0..10.0 => 1,
+      10.0.. => 0,
```
kotp:
```suggestion
      10.0..1.0/0 => 0,
```

Thrillberg:
That seems to have done it! Thanks, @kotp!

exercises/darts/README.md
```diff
@@ -0,0 +1,46 @@
...
+[Darts](https://en.wikipedia.org/wiki/Darts) is a game where players
+throw darts to a [target](https://en.wikipedia.org/wiki/Darts#/media/File:Darts_in_a_dartboard.jpg).
+
+In our particular instance of the game, the target rewards with 4 different amounts of points, depending on where the dart lands:
+
+* If the dart lands outside the target, player earns no points (0 points).
+* If the dart lands in the outer circle of the target, player earns 1 point.
+* If the dart lands in the middle circle of the target, player earns 5 points.
+* If the dart lands in the inner circle of the target, player earns 10 points.
+
```
kotp:
```
If the dart lands:

*  outside the targetg 0 points.
*  in the outer circle of the targetg 1 point.
*  in the middle circle of the targetg 5 points.
*  in the inner circle of the targetg 10 points.
```

Only a suggestion.  but perhaps easier to absorb as bullet points.

exercises/darts/README.md
```diff
@@ -0,0 +1,46 @@
...
+The outer circle has a radius of 10 units (This is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1. Of course, they are all centered to the same point (That is, the circles are [concentric](http://mathworld.wolfram.com/ConcentricCircles.html)) defined by the coordinates (0, 0).
+
+Write a function that given a point in the target (defined by its `real` cartesian coordinates `x` and `y`), returns the correct amount earned by a dart landing in that point.
```
kotp:
```suggestion
Write a method that, given a point in the target (defined by its `real` Cartesian coordinates `x` and `y`), returns the correct amount earned by a dart landing in that point.
```

Cartesian is a name, and capitalized.  The use of a comma is consistent  and works because without the explanation would make a complete sentence itself.

exercises/darts/README.md
```diff
@@ -0,0 +1,46 @@
...
+throw darts to a [target](https://en.wikipedia.org/wiki/Darts#/media/File:Darts_in_a_dartboard.jpg).
+
+In our particular instance of the game, the target rewards with 4 different amounts of points, depending on where the dart lands:
+
+* If the dart lands outside the target, player earns no points (0 points).
+* If the dart lands in the outer circle of the target, player earns 1 point.
+* If the dart lands in the middle circle of the target, player earns 5 points.
+* If the dart lands in the inner circle of the target, player earns 10 points.
+
+The outer circle has a radius of 10 units (This is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1. Of course, they are all centered to the same point (That is, the circles are [concentric](http://mathworld.wolfram.com/ConcentricCircles.html)) defined by the coordinates (0, 0).
```
kotp:
```suggestion
The outer circle has a radius of 10 units (This is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1. They are all centered to the same point (That is, the circles are [concentric](http://mathworld.wolfram.com/ConcentricCircles.html)) defined by the coordinates (0, 0).
```

The "Of course, " can be removed as we are providing the course.

Thrillberg:
I entirely agree with all your comments. The reason I didn't do such proofreading on my own is that I generated the README from the problem specifications. Should I make changes to the README here, just for the Ruby implementation of the exercise, or should I propose the changes on the problem specifications description: https://github.com/exercism/problem-specifications/blob/master/exercises/darts/description.md ?

exercises/darts/README.md
kotp: The change from "function" to "method" is a local Ruby change, and not suitable for the Readme in problem-specifications, while the missing comma should be suggested at the original source.

exercises/darts/README.md
kotp: This can be suggested at the source, rather than locally here.

kotp:
You are right.  The suggestions should be at the source.  The changes here are going to be the Ruby specific change of `function` changed to `method`.  I approved this PR, but if you want to change that word to be specifically correct for the Ruby track, that would be great.

I know when the source is changed and we re-generate, that this will be "incorrect" again.

Thrillberg:
>The changes here are going to be the Ruby specific change of `function` changed to `method`

Hmm yeah this is an interesting nuance. I'm inclined not to make the change here since, as you pointed out, it will get reverted back to `function` whenever we re-generate. Totally coincidentally I was walking a beginner through setting up Exercism on their computer and encountered the word `function` in the `hello_world` exercise in Ruby, too (https://github.com/exercism/ruby/blob/master/exercises/hello-world/README.md).

So maybe it's a losing battle to try to keep the terminology different between Ruby and other languages and they should just all be called `function`s?

Anyway, I'd be happy to propose the other (imo more important) suggestions you make at the source, which should be universally helpful regardless of programming language. Does that sound alright?

kotp:
Yeah, I am for making the changes local.  As we work on the content for the Ruby track, and update, we will notice when things slip (or not) and have regressions, but hopefully we will note them, perhaps at some point make some changes that help us with track specific language so it is less manual, less likely to revert.

Thrillberg:
Great! I've just made the wording changes. What do you think?

exercises/darts/README.md
```diff
@@ -1,20 +1,22 @@
...
+* outside the target: 0 points.
+* in the outer circle of the target: 1 point.
+* in the middle circle of the target: 5 points.
+* in the inner circle of the target: 10 points.
 
-Write a function that given a point in the target (defined by its `real` cartesian coordinates `x` and `y`), returns the correct amount earned by a dart landing in that point.
+The outer circle has a radius of 10 units (This is equivalent to the total radius for the entire target), the middle circle a radius of 5 units, and the inner circle a radius of 1. They are all centered to the same point (That is, the circles are [concentric](http://mathworld.wolfram.com/ConcentricCircles.html)) defined by the coordinates (0, 0).
+
+Write a method that, given a point in the target (defined by its `real` Cartesian coordinates `x` and `y`), returns the correct amount earned by a dart landing in that point.
```
kotp: Should `real` be marked up as code?  Or is this meant to be quoted?  And how does this compare with pretend Cartesian coordinates?

kotp: I noted that is in the canonical source as well.  Other than that, I think it looks good.  Let me know what you think, I am not opposed to bringing it in.

Issue #1025, opened 2020-03-30
Improve installation instructions


porkostomus:
I found the [instructions](http://tutorials.jumpstartlab.com/topics/environment/environment.html) linked to on `INSTALLATION.md` a bit confusing, particularly from a Linux perspective - it is focused on Rails development, and has the user install Postgres. Then I was led to a strange tutorial about installing Vagrant, and finally just went to the official ruby [guide](https://www.ruby-lang.org/en/documentation/installation/) which simply told me to use my package manager which just worked! IMO the official docs do a much better job.

kotp:
Very much Jumpstart oriented.  I had a hard time seeing the Rails connection, until they mentioned it in relationship to the database they chose to suggest.

Did you really install Postgres?  Sorry about that, not necessary for Ruby and not necessary for the Ruby track.

I agree that is a lot of information to digest, and would for sure be the wrong information to give to someone that is new to programming.

Yes, I think the Ruby guide is a much more reasonable resource, as I am sure the other instructions are more for a specific development environment... and we are more concerned with pretty plain Ruby.

Please do review the pull request that is mentioned in this issue.

xxylem:
Hi, just to add that I had a similar problem to @porkostomus, the old installation instructions for Windows lead to [http://railsinstaller.org/en](url), which ends up giving you Ruby 2.3. It looks like #1026 provided by @kotp will fix this.

kotp:
If someone from @exercism/ruby will approve it, it should be in shortly after that.

PR #1026, opened 2020-03-31
Official installation guides

kotp:
closes: #1025

/cc: @porkostomus

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,6 +1,6 @@
 ## Installation
 
-Use the [installation guide at Jumpstart Lab](http://tutorials.jumpstartlab.com/topics/environment/environment.html) for instructions on installing Ruby.
+Choose your installation method at the [Ruby installation page](https://www.ruby-lang.org/en/documentation/installation/) for instructions on installing Ruby.
 
 You will need to install the Minitest gem:
 


kotp:
@ErikSchierboom Looks like this repository is getting hit with the configlet standard in isn't gzip format.  What are the steps that we need to put an authentication token in place?

ErikSchierboom:
@kotp There are two steps involved:

1. Update the script to the [latest version](https://github.com/exercism/configlet/blob/master/scripts/fetch-configlet) to have it send authentication headers to GitHub.
1. Add a `GITHUB_TOKEN` environment variable (probably as a secret) to your Travis repository that enables you to fetch data. At this point, I can't help you, but @kytrinyx can probably help you out with this.

ErikSchierboom:
@kotp Katrina and I will try and get the GH token on Travis next week.

kotp:
> @kotp Katrina and I will try and get the GH token on Travis next week.

In the meantime, it looks like the change so far is enough to still get things moving, according to the tests I see.

kotp:
@porkostomus If you can review and approve we can get this in.

kotp:
Sorry, looks like merging this is blocked until someone on the @exercism/ruby team approves it.

docs/INSTALLATION.md
```diff
@@ -1,6 +1,6 @@
 ## Installation
 
-Use the [installation guide at Jumpstart Lab](http://tutorials.jumpstartlab.com/topics/environment/environment.html) for instructions on installing Ruby.
+Choose your installation method at the [Ruby installation page](https://www.ruby-lang.org/en/documentation/installation/) for instructions on installing Ruby.
```
dvik1950, line 3: Would it make sense to also recommend one of the methods? If I was installing Ruby for the first time I'd really like for someone to tell me about `rbenv` for example.

dvik1950, line 3: Would it make sense to recommend a method here? If I was installing Ruby for the first time, I'd like for someone to recommend me `rbenv` for example

kotp: You mean on the official site?  I think it does recommend methods there for different situations, operating systems, etc.

You wouldn't recommend `rbenv` for Windows users, I think.  Maybe you would.  I wouldn't recommend `chruby` for Linux users.

The users know their systems better than we do, I think the official page should address the ways that it can be installed on various systems.  They are vested in that information.

dvik1950: > The users know their systems better than we do
Good point, agreed.

Issue #1028, opened 2020-04-02
[Infrastructure] Update configlet to use authentication


kotp:
@kotp There are two steps involved:

1. Update the script to the [latest version](https://github.com/exercism/configlet/blob/master/scripts/fetch-configlet) to have it send authentication headers to GitHub.
1. Add a `GITHUB_TOKEN` environment variable (probably as a secret) to your Travis repository that enables you to fetch data. At this point, I can't help you, but @kytrinyx can probably help you out with this.

_Originally posted by @ErikSchierboom in https://github.com/exercism/ruby/pull/1026#issuecomment-607069051_

ErikSchierboom:
@kotp This week, @kytrinyx added an environment variable to the Travis environment that is named `NOT_KYTRINYX_TOKEN` (for reasons). Could you create a PR that tries to use this token with the updated version of the configlet script and see if it works (no errors)?

kytrinyx:
I believe this can be closed. If I'm wrong, please reopen with a comment about what needs to be done.

PR #1029, opened 2020-04-02
1028 update fetch configlet

kotp:
ref: #1028


kotp:
@ErikSchierboom looks like this might be enough.  Once this is in, I can tend to the blocked pull requests.

kotp:
> This is the first step. The second step is to add the `GITHUB_TOKEN` environment variable to your Travis repo. I'll contact Katrina to ask her for help to do that.

This is passing, so I thought perhaps that was already done.  Thanks for following up on this with @kytrinyx.

PR #1030, opened 2020-05-02
Holding github token

kotp:
ref: #1028

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 ---
+github_token=$NOT_KYTRINYX_TOKEN
 language: ruby
 sudo: false
 rvm:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
 ---
+github_token=$NOT_KYTRINYX_TOKEN
```
ErikSchierboom, line 2: I don't know much about Travis, but should this definition be part of an `env` key: https://docs.travis-ci.com/user/environment-variables/#defining-public-variables-in-travisyml

ErikSchierboom: The configlet script uses uppercase:
```suggestion
GITHUB_TOKEN=$NOT_KYTRINYX_TOKEN
```

kotp:
@kytrinyx can you check Travis, last run was about a month ago for exercism/ruby from what I can tell.

kotp:
Closing, as we have moved away from Travis.

PR #1035, opened 2020-06-24
microwave: Test that minute overflow adds to input minutes

petertseng:
If I enter a nonzero number of minutes into my microwave and add a
number of seconds that exceeds 1 minute, the correct behaviour would be
to add the additional minute to the existing minutes already input.

To test this, add the input `272`, which first is understood as 2
minutes 72 seconds, and then the 72 seconds is understood as 1 minute 12
seconds. The result is 3 minutes 12 seconds.

As noted in the README, this microwave *does* convert the input to
canonical format before display, so the output we would want here is
03:12.

The repo's solution was required to be changed to pass this new test.
Before its change, it had the result of "02:72" which violates the
README's stipulation that the microwave converts to canonical.

### DIFF
--- a/exercises/microwave/.meta/solutions/microwave.rb
+++ b/exercises/microwave/.meta/solutions/microwave.rb
@@ -2,10 +2,11 @@ class Microwave
   attr_reader :minutes, :seconds
 
   def initialize(digits)
-    base = digits > 99 ? 100 : 60
-
-    @minutes = digits / base
-    @seconds = (digits - minutes * base) % base
+    @minutes, @seconds = digits.divmod(100)
+    if @seconds >= 60
+      @seconds -= 60
+      @minutes += 1
+    end
   end
 
   def timer
--- a/exercises/microwave/microwave_test.rb
+++ b/exercises/microwave/microwave_test.rb
@@ -52,4 +52,9 @@ class MicrowaveTest < Minitest::Test
     # skip
     assert_equal '10:01', Microwave.new(1001).timer
   end
+
+  def test_minute_overflow_adds_to_input_minutes
+    skip
+    assert_equal '03:12', Microwave.new(272).timer
+  end
 end


petertseng:
If this is too large a step for students to handle at once, consider the input `160`, which still tests the overflow minute addition to existing input minutes, but has no overflow seconds.

petertseng:
If this is too large a step for students to handle at once, consider adding the input `160` before the test of `272`. The input of `160` would test the overflow minute addition to existing input minutes without the student having to also keep the remaining seconds.

exercises/microwave/.meta/solutions/microwave.rb
```diff
@@ -2,10 +2,11 @@ class Microwave
...
   def initialize(digits)
-    base = digits > 99 ? 100 : 60
-
-    @minutes = digits / base
-    @seconds = (digits - minutes * base) % base
+    @minutes, @seconds = digits.divmod(100)
+    if @seconds >= 60
+      @seconds -= 60
+      @minutes += 1
+    end
```
petertseng, line 9: possible to replace these four lines with these two lines;

```ruby
extra_minute, @seconds = @seconds.divmod(60)
@minutes += extra_minute
```

However, I find such a replacement to be deceptive about the possibilities. At this point in the code, seconds can only be in range `0..99`, which means the only possible numbers of overflow minutes resulting from that are either 0 or 1. Using `divmod(60)` seems to imply to the reader of this code that it could be anything else. I prefer not to deceive the reader.

kotp:
Tagging @RoelofWobben because of expressing interest in this exercise.

exercises/microwave/.meta/solutions/microwave.rb
kotp: I like the explicitness of the code stating that.  I agree.

kotp (approved):

PR #1038, opened 2020-09-03
Fix microwave tests

crunis:
Avoid overcooking the burrito


crunis:
Looks like I don't know how microwaves work..

PR #1085, opened 2021-01-29
Fix the tests and use multiple Ruby versions

iHiD:
This PR gets the tests passing again. The track's CI still fails due to configlet failing, but it gives us sanity again for Practice Exercises. I've added inline comments to explain individual changes.


.github/workflows/exercise-tests.yml
```diff
@@ -19,29 +18,19 @@ jobs:
...
 
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-ruby@v1
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@a699edbce608a2c128dedad88e3b6a0e28687b3c
         with:
           ruby-version: ${{ matrix.ruby-version }}
+          bundler-cache: true
```
iHiD, line 30: This removes the need for all the code below.

.gitignore
```diff
@@ -1,6 +1,5 @@
 .bundle
 bin/configlet
 bin/configlet.exe
-Gemfile.lock
```
iHiD, line 4: I don't know why the `Gemfile.lock` wasn't checked in, but it normally should be, so I've added it.

.ruby-version
```diff
@@ -0,0 +1 @@
+ruby-2.6.6
```
iHiD, line 1: Ruby 2.7 gives warnings about kwargs, but these warnings are no longer valid as Ruby 3.0 changed how they work, so I've just stuck with 2.6.6 for now, in line with all the other Ruby codebases (e.g. Test Runner)

lib/generator/files/track_files.rb
```diff
@@ -81,7 +81,7 @@ def update_version(version)
 
     class MinitestTestsFile < Writable
       def generate(template:, values:)
-        content = ERB.new(template, nil, '<>').result values.get_binding
+        content = ERB.new(template, trim_mode: '<>').result values.get_binding
```
iHiD, line 84: This is a change to the ERB lib in 2.6 onwards

lib/tasks/exercise_test_tasks.rb
```diff
@@ -16,18 +16,16 @@ def initialize(options:, test_runner: ExerciseTestsRunner)
...
 
-  def exercises
-    @exercises ||= Exercise.all
-  end
-
   def define_task_for_all_exercises
     desc 'Run the tests for all exercises'
-    task exercises: exercises
+    # TODO: Also add concept exercises
+    task exercises: Exercise.practice
```
iHiD, line 28: I've tested this with `Exercise.all` and it works in terms of generating the tests correctly, but the concept exercises currently fail the tests due to the structure being different, so I've stuck to practice exercises for now.

test/tasks/exercise_test.rb
```diff
@@ -5,47 +5,61 @@
...
-      end
+      FileUtils.mkdir_p "#{dir}/concept/test"
+      FileUtils.mkdir_p "#{dir}/concept/test2"
+      FileUtils.mkdir_p "#{dir}/practice/test3"
+      FileUtils.mkdir_p "#{dir}/practice/test4"
+
+      concept_file_list = FileList["#{dir}/concept/*"]
+      practice_file_list = FileList["#{dir}/practice/*"]
+      FileList.expects(:[]).with('exercises/concept/*').returns(concept_file_list).at_least_once
+      FileList.expects(:[]).with('exercises/practice/*').returns(practice_file_list).at_least_once
```
iHiD, line 16: I added `mocha` so I can do this ^^

lib/tasks/exercise.rb
```diff
@@ -1,25 +1,36 @@
...
+    def concept_slugs
+      FileList['exercises/concept/*'].pathmap('%f')
+    end
 
-    def slugs
-      FileList['exercises/*'].pathmap('%f').exclude('TRACK_HINTS.md')
+    def practice_slugs
+      FileList['exercises/practice/*'].pathmap('%f')
     end
   end
```
iHiD, line 22: All of this needs DRYing up at some point. But let's get everything working first.

ekingery (approved):

neenjaw:
I don't know how it exactly works all together, but each change looks reasonable with CI passing and listing each exercise.

iHiD:
Thanks both. I'll merge it down and keep going 🙂

PR #1091, opened 2021-02-06
Modern (as of 2021) rubocop

iHiD:
Our existing rubocop is quite out of date, and the config is an entire rubocop config rather than the list of overrides. This makes it difficult to know what's intentional and what's not, and makes it difficult to bring keep our Ruby up to date with the latest changes/standards. I also think we should have a consistent set of things for our students to read/see/work on.

I'm going through violation-by-violation adding a commit that addresses each. The base file I've started with is what is used elsewhere within Exercism on Ruby projects and matches my preferences. There might be some where people disagree with the Rubocop standard or my preferences. Where that's the case, let's discuss, make a decision, and encode it into the .rubocop file

There will be loads of commits - I'll keep adding them when I have time and I want an "easy" job to be working on.

I also don't know how the generator works for the Ruby track - is a lot of what's getting changed here automatically generated? If so, I guess we need to update the generator to deal with this?

Feel free to comment in the interim, but I'll mark this as not-draft once it's finally ready.


exercises/practice/custom-set/custom_set_test.rb
```diff
@@ -178,105 +178,105 @@ def test_adding_an_existing_element_does_not_change_the_set
...
 
   def test_union_of_a_non_empty_set_and_empty_set_is_the_non_empty_set
     skip
-    set1 = CustomSet.new [1, 3]
-    set2 = CustomSet.new []
+    set_1 = CustomSet.new [1, 3]
+    set_2 = CustomSet.new []
     expected = CustomSet.new [1, 3]
-    assert_equal expected, set1.union(set2)
+    assert_equal expected, set_1.union(set_2)
```
kotp, line 272: The position is not important, so this could be reported, but preference is to make the position less important via communication of style. `set1, set2` does this for us, while `set_1` indicates that there is an importance of placement.

Since order does not matter  for union, we should play down the position, using it only for "uniqueness of name".

exercises/practice/list-ops/list_ops_test.rb
```diff
@@ -44,9 +44,9 @@ def test_concat_normal
 
   def test_concat_gigantic
     skip
-    input1 = (1..1_000_000).to_a
-    input2 = (1_000_001..2_000_000).to_a
-    assert_equal (1..2_000_000).to_a, ListOps.concatter(input1, input2)
+    input_1 = (1..1_000_000).to_a
+    input_2 = (1_000_001..2_000_000).to_a
+    assert_equal (1..2_000_000).to_a, ListOps.concatter(input_1, input_2)
```
kotp, line 49: See importance of position vs importance  of uniqueness communication in prior comment.

Since these are (now or later) generated, the maintainer should make this distinction, if this style is enforced, by indicating in the source that this is purposefully done, via a directive comment.  See https://stackoverflow.com/a/26350879/485864 where I talk about inline directive.

kotp (commented):
Some of the comments are likely changed via the `.rubocop.yml` file.

.rubocop.yml
```diff
@@ -1,1892 +1,87 @@
...
-  StyleGuide: https://github.com/bbatsov/ruby-style-guide#80-character-limits
-  Enabled: false
-  Max: 80
-  AllowHeredoc: true
-  AllowURI: true
-  URISchemes:
-  - http
-  - https
+  Exclude:
+    - "test/**/*"
```
kotp, line 49: If we exclude metrics for tests, and pay attention to metrics in others, then we probably don't need to `--except Metrics` when running Rubocop in CI.

.rubocop.yml
```diff
@@ -1,1892 +1,87 @@
...
-Performance/EndWith:
-  Description: Use `end_with?` instead of a regex match anchored to the end of a string.
-  Reference: https://github.com/JuanitoFatas/fast-ruby#stringmatch-vs-stringstart_withstringend_with-code-start-code-end
+Style/BlockComments:
   Enabled: false
 
-Performance/FixedSize:
-  Description: Do not compute the size of statically sized objects except in constants
+# We use Heredoc indentation in exercises
+# where variable indentiation is desirable
```
kotp, line 85: If this is not universally (in the Ruby track) the case, then this should be enabled, and a directive comment can communicate that it is intentional.

iHiD: Happy to enable it, but then there's a few places that would need fixing and I don't know how to fix them.

kotp: I can enable it, see what breaks.

kotp: Feel free to continue to push, I can work through rolling updates, not a problem.

iHiD: I'm done for now. I need your help on the remaining failures I think.

exercises/practice/binary/binary_test.rb
```diff
@@ -3,42 +3,42 @@
 
 # Common test data version: 1.1.0 dd3c07e
 class BinaryTest < Minitest::Test
-  def test_binary_0_is_decimal_0
+  def test_binary_0_is_decimal_0 # rubocop:disable Naming/VariableNumber
```
ErikSchierboom, line 6: Is it possible to have Rubocop disable this rule for this file only?

exercises/practice/connect/connect_test.rb
```diff
@@ -11,7 +11,7 @@ def test_an_empty_board_has_no_winner
       '  . . . . .',
       '   . . . . .',
       '    . . . . .'
-    ].map {|row| row.gsub(/^ */, '')}
+    ].map { |row| row.gsub(/^ */, '') }
```
ErikSchierboom, line 14: An off-topic remark: I found the way this is written using the map/gsub combination really hard to parse as someone newer to Ruby

ErikSchierboom (approved):
Big jump in Rubocop versions! Excellent.

kotp:
On 21/02/08 11:45PM, Erik Schierboom wrote:
> @ErikSchierboom approved this pull request.
> > @@ -3,42 +3,42 @@
>  
>  # Common test data version: 1.1.0 dd3c07e
>  class BinaryTest < Minitest::Test
> -  def test_binary_0_is_decimal_0
> +  def test_binary_0_is_decimal_0 # rubocop:disable Naming/VariableNumber
> 
> Is it possible to have Rubocop disable this rule for this file only?

Yes, that is what the comment does.  We would put it at the top of the
file, maybe, and then re-enable at the end.  We can also do so in the
configuration file.

> > @@ -11,7 +11,7 @@ def test_an_empty_board_has_no_winner
>        '  . . . . .',
>        '   . . . . .',
>        '    . . . . .'
> -    ].map {|row| row.gsub(/^ */, '')}
> +    ].map { |row| row.gsub(/^ */, '') }
> 
> An off-topic remark: I found the way this is written using the
> map/gsub combination really hard to parse as someone newer to Ruby

Is it because "new to Ruby" or perhaps "New to regular expressions?"  I
don't disagree though, as I suggest using transliteration (`tr`) instead, or
`scan` instead, usually because scan works by focusing on what is
wanted, rather than getting rid of what you don't want.  A better
viewpoint usually.

ErikSchierboom:
> Is it because "new to Ruby

New to Ruby :)

exercises/practice/binary/binary_test.rb
kotp: It might be better to reword the titles of test methods as well.

lib/generator/implementation.rb
```diff
@@ -35,6 +35,8 @@ class ImplementationDelegator < SimpleDelegator
   # A implementation that also logs its progress.
   class LoggingImplementation < ImplementationDelegator
     def initialize(implementation:, logger:)
+      super
```
kotp, line 38: This may not even be necessary.  But was not incorrect to use the "no pass through" directive here that is indicated with the `super()`, a quick check indicates the arguments are used locally, and not intended for consumption further up the ancestor chain.  The same quick looking suggests that we would not even need to call the parents `initialize` method.

kotp: Disregard the above for now.  Found the reason for it.

exercises/practice/connect/connect_test.rb
kotp: @ErikSchierboom do you have a way that you would write this instead?  (Getting ready to merge this, pending the failed CI corrections).

kotp (dismissed):

iHiD:
GREEN. Wahooo. Want to Squash + Merge this @kotp?

kotp:
> GREEN. Wahooo. Want to Squash + Merge this @kotp?

Yes, I applied the patch Reading for 3.0, and still green.

I still don't understand why it wants me to do another pull request for Ruby 2.6 and another pull request for Ruby 2.7.

kotp (approved):

kotp:
> > GREEN. Wahooo. Want to Squash + Merge this @kotp?
To clarify, I can't merge with the:

- Ruby 2.6 - ubuntu-latest - pull_request Expected — Waiting for status to be reported
- Ruby 2.7 - ubuntu-latest - pull_request Expected — Waiting for status to be reported

Status listing as it is.  And I don't know what it means, what it is waiting on…

iHiD:
@kotp All good to go now :)

PR #1093, opened 2021-02-11
RubyLearning .org domain not set up for classes

kotp:
The course material is no longer available on our former platform, classes link has been (temporarily) suspended for RubyLearning.  Regular content is available, and of course, I am still available for answers in regard to the RubyLearning.com site.


kotp:
@iHiD @ErikSchierboom This has been rebased on origin/main.  Can be brought in.

Issue #1083, opened 2021-01-29
Launch Tracker 🔴


ErikSchierboom:
_This issue is part of the migration to v3. You can read full details about the various changes [here](https://github.com/exercism/v3-launch/)._

To get your track ready for Exercism v3, the following needs to be done:

- [x] [Configure online editor](https://github.com/exercism/ruby/issues/1074)
- [x] [Add key features](https://github.com/exercism/ruby/issues/1075)
- [x] [Add tags](https://github.com/exercism/ruby/issues/1076)
- [ ] [Add prerequisites to Practice Exercises](https://github.com/exercism/ruby/issues/1077)
- [ ] [Update status of track](https://github.com/exercism/ruby/issues/1081)
- [ ] [Update status of Concept Exercises](https://github.com/exercism/ruby/issues/1082)
- [ ] [Pass linting checks](https://github.com/exercism/ruby/issues/1078)
- [ ] [Build Test Runner](https://github.com/exercism/ruby/issues/1079)
- [ ] [Build Analyzer and Representer](https://github.com/exercism/ruby/issues/1080) (optional)
- [ ] [Add blurb to Concept Exercises](https://github.com/exercism/ruby/issues/1108)

This issue may be automatically added to over time. While track maintainers should check off completed items, please do not add/edit items in the list.

## Tracking

https://github.com/exercism/v3-launch/issues/7

ErikSchierboom:
@kytrinyx This can be closed. We used this for the v3 launch and for tracks that weren't v3 ready afterwards.

kytrinyx:
Thank you!

PR #1095, opened 2021-02-13
V3 add key features and tags

kotp:
Don't squash, in case we need to revert either features or tags.

Merge or merge with fast forward instead.

Ref: #1083


iHiD:
@kotp I've incorporated these into #1124

PR #1097, opened 2021-02-16
add tool versions to gitignore

neenjaw:
Just a small QoL addition for me since I use asdf to manage ruby versions. Then I also can't accidentally pollute the repo with person-specific files. 🙏🏻


kotp:
Add it to your `~/.gitignore`, and manage it locally?  Since it is used "for you"?  It is the same for Windows, Mac, Linux, editor swap files, etc., they aren't project related, but personal hygiene items.

PR #1104, opened 2021-02-20
[patch] update concepts for birdwatcher exercise

neenjaw:
Currently the config says it unlocks 'strings', but I am pretty sure this is an 'arrays' or enum exercise.


neenjaw:
I don't know why my prettier settings keep messing with this file since I thought they were they same settings from v3.

PR #1111, opened 2021-03-08
[v3] new concept exercise: flight-to-mars (exceptions, a couple enum methods)

neenjaw:
This is the start of an exercise for the exceptions concept (which seems to be present already without an exercise).  It also includes a few enum methods as well.

TODO:
- test suite
- introduction
- hints
- concepts that may be incomplete


exercises/concept/flight-to-mars/.meta/exemplar.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class LaunchStatusCheck
+  class LaunchStatusError < StandardError
+  end
+
+  class AbortLaunchError < StandardError
+  end
+
+  LAUNCH_STATUS_MESSAGE = 'Launch flight team(s) have not given a go/no-go status.'
+  ABORT_LAUNCH_MESSAGE = 'Launch flight team(s) have given a no-go status.'
```
kotp, line 11:
```suggestion
  class LaunchStatusError < StandardError
    def initalize(message = 'Launch flight team(s) have not given a go/no-go status.')
      super
    end
  end

  class AbortLaunchError < StandardError
    def initialize(message = 'Launch flight team(s) have given a no-go status.')
      super
    end
  end
```

exercises/concept/flight-to-mars/.meta/exemplar.rb
```diff
@@ -0,0 +1,32 @@
...
+    @flight_teams.all? { |team| status_checks.key?(team) }
+  end
+
+  def all_teams_are_go(status_checks)
+    status_checks.values.none? { |status_check| status_check == NO_GO }
+  end
+
+  def check(status_checks)
+    raise LaunchStatusError, LAUNCH_STATUS_MESSAGE unless all_teams_checked_in(status_checks)
+    raise AbortLaunchError, ABORT_LAUNCH_MESSAGE unless all_teams_are_go(status_checks)
```
kotp, line 30:
```suggestion
    raise LaunchStatusError unless all_teams_checked_in(status_checks)
    raise AbortLaunchError unless all_teams_are_go(status_checks)
```

This continues from the change in exception classes and removes the constants.  The constants mix concerns, and were always used, but we can move the constant message to default messages in the exception classes, leaving it flexible as well.

kotp (commented):
I know it is still draft, so this is just a comment.

neenjaw:
@kotp Thanks for the suggestions, I have committed them both!

iHiD:
(I will get back to this eventually btw)

kytrinyx:
It's been 18 months since the last movement on this PR.

Does someone who was involved at the time have an idea of what it would take to get this over the finish line?

exercises/concept/flight-to-mars/flight_to_mars.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class LaunchStatusCheck
+  # create a custom `LaunchStatusError` error
+
+  # create a custom `AbortLaunchError` error
+
+  def initialize(flight_teams)
+    raise NotImplementedError, 'Please implement the LaunchStatusCheck#initialize method'
```
kotp, line 9: In other discussion, we have decided to not use the `NotImplementedError` since it implies and is intended for those things not implemented "on this platform" rather than a simple "not available to use".  I would recommend a `NoMethodError` with the message to cause the desired method to exist.

kotp (changes requested):

PR #1124, opened 2021-04-18
[WIP] Get all existing Concept Exercises in reasonable shape

iHiD:
See https://github.com/exercism/ruby/issues/1123


iHiD:
I'm going to merge this, and we can make further improvements by actually testing out the exercises. I think it's basically an impossible PR to review.

PR #1141, opened 2021-06-10
Add hiljusti as a maintainer

hiljusti:
I've been helping out on some other tracks, but would like to assist here every once in a while with reviews.


hiljusti:
Nevermind - I think I really just wanted to be in `exercism/reviewers` to help out

PR #1145, opened 2021-06-23
Getting Started for Hello World

kotp:
Steps are adjusted, and reflect closer to what should happen when a
student starts with Hello World exercise in Ruby.


SleeplessByte (commented):
Makes sense for the Web IDE.

kytrinyx (approved):
This looks good to me.

PR #1150, opened 2021-07-07
🤖 Simplify `hello-world` stub

ErikSchierboom:
See https://github.com/exercism/v3-launch/issues/46

### DIFF
--- a/exercises/practice/hello-world/hello_world_test.rb
+++ b/exercises/practice/hello-world/hello_world_test.rb
@@ -15,7 +15,7 @@ end
 class HelloWorldTest < Minitest::Test
   def test_say_hi
     # skip
-    assert_equal "Hello, World!", HelloWorld.hello
+    assert_equal "Hello, World!", HelloWorld.new.hello
   end
 end
 
--- a/exercises/practice/hello-world/hello_world_test.rb
+++ b/exercises/practice/hello-world/hello_world_test.rb
@@ -15,7 +15,7 @@ end
 class HelloWorldTest < Minitest::Test
   def test_say_hi
     # skip
-    assert_equal "Hello, World!", HelloWorld.new.hello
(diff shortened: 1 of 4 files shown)


exercises/practice/hello-world/hello_world_test.rb
```diff
@@ -15,7 +15,7 @@
 class HelloWorldTest < Minitest::Test
   def test_say_hi
     # skip
-    assert_equal "Hello, World!", HelloWorld.hello
+    assert_equal "Hello, World!", HelloWorld.new.hello
```
ErikSchierboom, line 18: I've changed this to use `.new` to allow the removal of `self.` in the stub.

SleeplessByte: Wouldn't

```ruby
def hello
  "Goodbye, Mars!"
end
```

technically be less introduced syntax?

ErikSchierboom: You mean without a class? Is that valid? I don't know :D

ErikSchierboom: @iHiD Brought up an interesting point yesterday, and that is that my change (and yours too) would break all existing solutions. So that might be something to consider.

iHiD: Having thought about this, I'd rather stick with the structure of how this used to be (so `HelloWorld.hello`), and just change the stub not the exercise. This means that we don't break everyone's solutions and also don't break the tutorials that are out there (e.g. [this with 13k views](https://www.youtube.com/watch?v=ixSJXImknnk&ab_channel=Exercism)).

SleeplessByte: Good reason :)

iHiD:
Thanks team :)

PR #1165, opened 2021-07-22
Moviegoer improvements

angelikatyborska:
Hi! I just solved this exercise for the first time and I noticed some details that I would like to improve:

1. No hints. I tried to write some but my Ruby is very rusty so please review thoroughly.
2. The text about exceptions mentioned that exercise stubs raise `NotImplementedError`. This is no longer true.
3. Small typos in the introduction.

### DIFF
--- a/concepts/exceptions/about.md
+++ b/concepts/exceptions/about.md
@@ -7,6 +7,7 @@ At any point in our code, we can "raise" an exception.
 We do this using the `raise` method, passing in an object - normally an Exception object, although we can also use basic strings.
 For example, you'll see in the exercise stubs that we use the built-in `RuntimeError` to tell Ruby that a method hasn't been implemented.
 You can also use the shorthand syntax of `raise(ExceptionObject, params)`.
+If the exception class is omitted, `RuntimeError` is used by default.
 For example:
 
 ```
--- a/concepts/exceptions/about.md
+++ b/concepts/exceptions/about.md
@@ -5,14 +5,15 @@ They are called exceptions, as they normally appear when dealing with some unexp
 
 At any point in our code, we can "raise" an exception. 
 We do this using the `raise` method, passing in an object - normally an Exception object, although we can also use basic strings.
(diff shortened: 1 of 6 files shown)


angelikatyborska:
I also noticed that this exercise teaches raising exceptions but that's not reflected in the UI. Is this still WIP?

Screen Shot 2021-07-22 at 19 38 53

concepts/exceptions/about.md
```diff
@@ -5,14 +5,15 @@ They are called exceptions, as they normally appear when dealing with some unexp
 
 At any point in our code, we can "raise" an exception. 
 We do this using the `raise` method, passing in an object - normally an Exception object, although we can also use basic strings.
-For example, you'll see in the exercise stubs that we use the built-in `NotImplementedError` to tell Ruby that a method hasn't been implemented.
-You can also use the shorthand syntax of `raise(ExceptionObject, params)`. 
+For example, you'll see in the exercise stubs that we use the built-in `RuntimeError` to tell Ruby that a method hasn't been implemented.
+You can also use the shorthand syntax of `raise(ExceptionObject, params)`.
```
iHiD, line 9:
```suggestion
You can also use the shorthand syntax of `raise(ExceptionObject, params)`.
If the exception class is omitted, RuntimeError is used by default.
```

angelikatyborska: Any here or also in introductions?

iHiD: I think in all of them.

iHiD:
> I also noticed that this exercise teaches raising exceptions but that's not reflected in the UI. Is this still WIP?

There's a mismatch between the slugs (`exceptions` vs `raising-exceptions`). I think I was undecided on whether we have a seperate exercise on catching/handling/reraising exceptions. 

I think probably let's call this `exceptions` and later add another for `exception-handling`. So renaming the concept in this exercise's config to `exceptions` should fix it.

iHiD:
Thank you! 💙

PR #1166, opened 2021-07-28
Add icons for key features

ee7:
The icons for key features are now available, and `configlet lint` now requires that each key feature has a valid `icon` value.

This PR sets an initial icon for each key feature - feel free to change them however you want. You can see the list of available icons [here](https://github.com/exercism/docs/issues/189). Note that an icon can be used regardless of its name.

With the initial state of this PR, the key features look something like the below.

---

fun Developer happiness
Ruby puts developer-happiness first. It has an elegant syntax that is natural to read and write.

extensible Metaprogramming magic
Ruby code can write and invoke Ruby code.

garbage-collected Garbage Collection
Garbage collection done via mark and sweep, stays out of your way.

powerful Large standard library
Ruby's standard library provides a wealth of classes utilities for common tasks.

tooling Flexible package manager
Packages (gems) can be centrally managed, but can also include custom or private gem stores.

dynamically-typed Strong, dynamic typing
Ruby is strong and dynamically typed and supports 'Duck Typing'. Everything in Ruby is an object

### DIFF
--- a/config.json
+++ b/config.json
@@ -1606,32 +1606,32 @@
   ],
   "key_features": [
     {
-      "icon": "",
+      "icon": "fun",
       "title": "Developer happiness",
       "content": "Ruby puts developer-happiness first. It has an elegant syntax that is natural to read and write."
     },
     {
-      "icon": "",
+      "icon": "extensible",
       "title": "Metaprogramming magic",
       "content": "Ruby code can write and invoke Ruby code."
     },
     {
-      "icon": "",
+      "icon": "garbage-collected",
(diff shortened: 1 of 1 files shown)


SleeplessByte:
I think we should just merge this so it's there, and make further changes later. What do you think @kotp?

SleeplessByte (approved):

angelikatyborska (approved):

ErikSchierboom (approved):

iHiD:
Thanks!

PR #1182, opened 2021-08-30
Use actual concepts for practices

iHiD:


config.json
```diff
@@ -1063,8 +1063,8 @@
         "slug": "diamond",
         "name": "Diamond",
         "uuid": "c55c75fb-6140-4042-967a-39c75b7781bd",
-        "practices": ["chars", "string-formatting"],
-        "prerequisites": ["strings", "string-formatting", "chars"],
+        "practices": ["chars", "strings"],
```
kotp, line 1066: Should this be "characters" instead of fish instead?  Full words used for everything else, here we use "chars" which is defined in English as a kind of fish or something partially burnt.  While `chars` is the name of a method, we aren't using method names here, it seems.

iHiD: This is copied from C#, where a char is a thing: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/char

The purpose of this PR is to remove references to chars anyway :)

kotp: > The purpose of this PR is to remove references to chars anyway

Right on.  But it seems to still be here?

PR #1183, opened 2021-08-31
Lasagna Documentation: Trailing white spaces removed.

kotp:


angelikatyborska:
The additions to the introduction document are not appropriate. The text in them cannot be CLI specific, it has to work for the web editor too. E.g. you can't mentiong `HELP.md`. The instructions have to be in a separate `instructions.md` file. Those two files will get merged together via the CLI into a `README.md` file when downloaded, and also by the platform when viewed on the website. The addition of authors doesn't follow v3's requirements for this file.

Basically I would recommend removing all the changes but the trailing whitespace removal.

iHiD (changes requested):
I agree with Angelika's comments.

kotp:
Thanks both.

angelikatyborska (approved):

exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -1,9 +1,9 @@
 # Introduction
 
 This exercise will teach the fundamental building blocks of Ruby.
-It covers a lot of ground, but don't worry if it feels overwhelming, we'll go into everything in much more depth as we continue through the track.
+It covers a lot of ground, but do not worry if it feels overwhelming, we will go into everything in much more depth as we continue through the track.
 
-One of the key things to understand about Ruby is that it is an [object-oriented language][object-oriented-programming], and that **everything in Ruby is an [object][object]**. Numbers, strings, methods - they're all objects.
+One of the key things to understand about Ruby is that it is an [object-oriented language][object-oriented-programming], and that **everything in Ruby is an [object][object]**. Numbers, strings, arrays - they're all objects.
```
iHiD, line 6: When not `method`?

kotp: Method is not an object unless you call `method(:some_method)` and then you can pass it around as an argument, because at that time it becomes a `Method` object.  See https://ruby-doc.org/core-3.0.2/Object.html#method-i-method for those details.

It is one of the very few things in Ruby that isn't an object, and I can't think of another thing that isn't an object in Ruby.

kotp: You can see it being used more directly [here](https://ruby-doc.org/core-3.0.2/Method.html) as well.

iHiD: TIL. Thanks.

kotp: Blocks… The other thing is blocks.  So, it is just a fabulatorism<sup>1</sup> to help describe the lesson, what is almost true, for the convenience of teaching, without the telling of these couple of exceptions. 

* * *
1: A made up word, but based on the word _fabulist_, a story teller of fables to give a lesson.

PR #1185, opened 2021-08-31
Contractions, abbreviations and whitespace, oh my.

kotp:
Using `string` rather than `str` to keep the style of using full words for things that we write.

### DIFF
--- a/exercises/concept/log-line-parser/.docs/introduction.md
+++ b/exercises/concept/log-line-parser/.docs/introduction.md
@@ -2,16 +2,16 @@
 
 Like everything in Ruby, a `String` is an object.
 Strings are made up of an arbitrary sequence of bytes (normally characters) and are initialized and manipulated like any other object.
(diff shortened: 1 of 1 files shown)


angelikatyborska (dismissed):

iHiD:
I think the change from `str` to `string` makes this a little more confusing. Many people will come from languages where you call methods on string as a keyword or special thing, and this might be more confusing because of it.

Could we agree on a different variable name that makes it clear this is a variable? `my_string`?

kotp:
> my_string

I like it.  I really dislike using "str" or "len" and especially "num" and "char" because these last two have full word meanings, and we are definitely dealing with English, Programming Language, plus whatever native language potential there is for confusion.  Easier to look up full words in a dictionary, and because of that some of the "programming common abbreviations" are potentially very confusing.

SleeplessByte (approved):

PR #1187, opened 2021-09-03
Update introduction for amusement-park exercise

sevenseacat:
To make the demo code work as expected, the instance variable names need to match.

Screen Shot 2021-09-03 at 6 04 47 pm

### DIFF
--- a/exercises/concept/amusement-park/.docs/introduction.md
+++ b/exercises/concept/amusement-park/.docs/introduction.md
@@ -17,14 +17,14 @@ class Repeater
   end
 
   def repeat
-    # Outputs the value of the instance variable
-    print @heard
+    # Return the value of the instance variable
+    @heard
   end
 end
 
 repeater = Repeater.new
 repeater.hear("Hello, friend!")
-repeater.repeat # Prints "Hello, friend!"
+print repeater.repeat # Prints "Hello, friend!"
 ```
 
 Objects usually set their initial state in an `initialize` method, which is automatically called when calling `new` on a class.
--- a/exercises/concept/amusement-park/.docs/introduction.md
+++ b/exercises/concept/amusement-park/.docs/introduction.md
@@ -18,7 +18,7 @@ class Repeater
 
   def repeat
     # Outputs the value of the instance variable
-    print @name
+    print @heard
   end
 end
 


kotp (approved):

kotp:
I have approved, but I do wonder if we should be using `print` in the method there.  It should be that the method name itself should indicate that it is going to match the behavior that the method has.  The method currently has the behavior that it returns `nil` while sending output to `STDOUT`.

```diff
diff --git i/exercises/concept/amusement-park/.docs/introduction.md w/exercises/concept/amusement-park/.docs/introduction.md
index 7e3b09e2..d41b9c00 100644
--- i/exercises/concept/amusement-park/.docs/introduction.md
+++ w/exercises/concept/amusement-park/.docs/introduction.md
@@ -17,14 +17,13 @@ class Repeater
   end
 
   def repeat
-    # Outputs the value of the instance variable
-    print @heard
+    @heard
   end
 end
 
 repeater = Repeater.new
 repeater.hear("Hello, friend!")
-repeater.repeat # Prints "Hello, friend!"
+print repeater.repeat
 ```
 
 Objects usually set their initial state in an `initialize` method, which is automatically called when calling `new` on a class.
```

kotp:
Today I learned that I can use three squigglies when there are three backticks in order to get the code blocks to view properly in a "nested code block" such as what may happen with a "Diff" view.

sevenseacat:
Yeah I like that idea much better! I'll make those changes

exercises/concept/amusement-park/.docs/introduction.md
```diff
@@ -17,14 +17,14 @@ class Repeater
   end
 
   def repeat
-    # Outputs the value of the instance variable
-    print @name
+    # Return the value of the instance variable
```
kotp, line 20: We don't mention that we are returning `@heard` (though we are really returning `something`) on line 14, in addition to saving the reference to `something`.  Should we?  It is an early exercise, and may be worthwhile being consistent.  Definitely a different commit though, as it is a different concern.

Generally, though, we would not comment the code.  Instead, talk about the code in prose.  This is an instructional document, so flexibility for learning and explaining sake though.

kotp (approved):

kotp:
Thank you @sevenseacat .  I just wanted to bring these in today, so the changes are available!  Good job!

PR #1188, opened 2021-09-03
Fix tests for the Microwave task

woarewe:
Hi! Thank you very much for the task!
I was playing with Ruby track and found out that the task can be finished cause tests are broken.

https://exercism.org/tracks/ruby/exercises/microwave

### DIFF
--- a/exercises/practice/microwave/microwave_test.rb
+++ b/exercises/practice/microwave/microwave_test.rb
@@ -4,57 +4,46 @@ require_relative 'microwave'
 # Common test data version: 1.7.0 cacf1f1
 class MicrowaveTest < Minitest::Test
   def test_one_second
-    # skip
     assert_equal '00:01', Microwave.new(1).timer
   end
 
   def test_fifty_nine_seconds
-    skip
     assert_equal '00:59', Microwave.new(59).timer
   end
 
   def test_sixty_seconds
-    skip
     assert_equal '01:00', Microwave.new(60).timer
   end
 
   def test_one_minute
-    skip
     assert_equal '01:00', Microwave.new(100).timer
   end
 
   def test_ninety_seconds
-    skip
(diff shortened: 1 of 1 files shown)


exercises/practice/microwave/microwave_test.rb
```diff
@@ -4,57 +4,39 @@
...
+  
   def test_one_minute_and_fifty_nine_seconds
-    skip
     assert_equal '01:59', Microwave.new(159).timer
   end
 
   def test_two_minutes
-    skip
-    assert_equal '02:00', Microwave.new(200).timer
+    assert_equal '02:00', Microwave.new(120).timer
```
kotp, line 32: This is not true for the task at hand.  Notice 159 being 1 minute and 59 seconds, and 200 should be 2 minutes.  The only part that is an exception to this is the less than 100 entered keeps the number of minutes.

exercises/practice/microwave/microwave_test.rb
```diff
@@ -4,57 +4,39 @@
...
   end
 
-  def test_over_ten_minutes
-    # skip
-    assert_equal '10:01', Microwave.new(1001).timer
+  def test_ten_minutes_and_one_second
+    assert_equal '10:01', Microwave.new(601).timer
```
kotp, line 36: The same here, it should be that 1001 enters as 10 minutes and 1 second.

kotp (changes requested):
Revert the test changes, but keep the removal of the skips.

exercises/practice/microwave/microwave_test.rb
woarewe: So, now I got the idea. Thank you very much

kotp (approved):
Thank you @woarewe !

PR #1189, opened 2021-09-04
Remove skips for TwoFer tests

woarewe:


kotp:
Looks like we should work on the test generators again... so that we don't need to do these individually.  In the meantime, I am going to approve this.

kotp (approved):

iHiD:
@kotp What's the reason we did this? Skips should be in exercises so they are solved using TDD. Has this changed somewhere that I'm not aware of?

kotp:
No, it really should go back in, it was during the time of my confusion about concept vs practice.

PR #1191, opened 2021-09-05
Bird Count: Force predicate methods to return boolean

woarewe:
Hi!
First of all, thank you very much for such an amazing platform, I like it very much!

I'm going through all the exercises and solving [the task](https://exercism.org/tracks/ruby/exercises/bird-count) I noticed that
tests are passed when the method `day_without_birds?` returns `truthy` or `falsey` values.

It would be nice to anticipate boolean values because of the following reasons:
1. It's mentioned in the description that the method should return either `true` or `false`
2. It's a good practice to return exactly boolean values if a method ends with `?` cause it's semantically better. Yes, it works in Ruby `ifs` but I think it's not a good idea to use approaches cause they actually can work

**NOTE FOR REVIEWERS**
I'm not sure how it's better to check if a method returns exactly a boolean value.
These
```ruby
assert false, method?
assert true, method?
```
haven't been passed by Rubocop.

### DIFF
--- a/exercises/concept/bird-count/bird_count_test.rb
+++ b/exercises/concept/bird-count/bird_count_test.rb
@@ -33,10 +33,10 @@ class BirdCountTest < Minitest::Test
   end
 
   def test_has_day_without_birds
-    assert BirdCount.new([5, 5, 4, 0, 7, 6]).day_without_birds?
+    assert BirdCount.new([5, 5, 4, 0, 7, 6]).day_without_birds?.equal?(true)
   end
 
   def test_has_day_without_birds_whith_no_day_without_birds
-    refute BirdCount.new([4, 5, 9, 10, 9, 4, 3]).day_without_birds?
+    assert BirdCount.new([4, 5, 9, 10, 9, 4, 3]).day_without_birds?.equal?(false)
   end
 end


exercises/concept/bird-count/bird_count_test.rb
```diff
@@ -33,10 +33,10 @@ def test_busy_days_for_busy_week
   end
 
   def test_has_day_without_birds
-    assert BirdCount.new([5, 5, 4, 0, 7, 6]).day_without_birds?
+    assert BirdCount.new([5, 5, 4, 0, 7, 6]).day_without_birds?.equal?(true)
```
SleeplessByte, line 36:
```ruby
assert_equal true, BirdCount.new([5, 5, 4, 0, 7, 6]).day_without_birds?
```

From <https://docs.seattlerb.org/minitest/Minitest/Assertions.html>. This should not pass if `day_without_birds?` returns non `true` but thruthy value.

woarewe: I have mentioned this in the description.

woarewe: https://minitest.rubystyle.guide/#assert-truthy

SleeplessByte: It's marked as bad because generally you don't want strict type checking but in this case I agree with you that we do want strict type checking.

I would still opt for assert_equal true, ... Over assert true, ...

Let's see how the rest feels 🙂😌

woarewe: @SleeplessByte The issue here is that the CI fails. I fully agree that `assert true, expected` is better and more readable.
But it would require reconfiguring the Rubocop for the entire project.

kotp:
> exactly a boolean value

If we are talking about this, then it should be **B**oolean" rather than the less strict **b**oolean reference.  `boolean` suggests truthy/falsey, while Boolean is proper, and proper Booleans are exactly true and false.

I am going to look at this closer, and try to determine if there is loss of educational content if we do change this to be strict, or if it improves that.  (Sometimes being too strict or specifying exactly, the learner suffers because they are not allowed to make those discoveries and reason them out for themselves.)

exercises/concept/bird-count/bird_count_test.rb
kotp: Reconfiguring Rubocop is not necessary, if there is an exception to the style rules, we can do something as outlined here:

[My answer on Stack Overflow](https://stackoverflow.com/a/26350879/485864)

kotp: Not a change in this commit, but for sure line 1 can go away.  There is not a single vulnerable string literal in this entire file.  So the directive may be "misleading" as I had to look through the entire file to see what we are protecting.

kotp:
I think the key is in the documentation:  "should return" rather than "must return".  And interestingly, when I solved, I did solve with a Boolean, rather than a boolean, (to communicate the difference without specifically having to write more to express what I mean), as it is the idiomatic way to solve (without looking at our exemplar solution, which reminds me I should look at that as well) this area of the exercise.

Rubocop sometimes suggests some silly things and so I do not take too much weight with the style guide as written (such as constants being frozen, where Rubocop suggests freezing constants that are not "freezable" such as the Numeric, Regular Expression and Ranges, being just a few examples.)  So I always read those with a grain of salt, and a human (rather than robotic) brain.  (After all, we do not serve the robots (yet), and "guide" does mean "guide".)

I am leaning toward not strictly checking, and when mentored, we can talk about that, leading them to why.   They also have access to the exemplar, I think after they pass the tests.  Then if they have questions why, they can ask for mentoring.

I might also be surprised to see some non-Boolean solutions for that, but am interested to see the creative ways folks solve that!

woarewe:
Got it! @kotp @SleeplessByte Thank you guys very much!

PR #1192, opened 2021-09-06
Add class interfaces to better web editor experience

woarewe:
I think it would be much better if exercises had predefined method interfaces required to implement.

https://exercism.org/tracks/ruby/exercises/series

### DIFF
--- a/exercises/practice/series/series.rb
+++ b/exercises/practice/series/series.rb
@@ -1,7 +1,7 @@
-=begin
-Write your code for the 'Series' exercise in this file. Make the tests in
-`series_test.rb` pass.
+class Series
+  def initialize(string); end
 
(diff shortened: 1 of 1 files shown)


kotp (commented):
We may or may not care that the student creates a class.  Also, if it is far enough in the track, this becomes less useful.  At that point, leave it to the student to design their solution as they would like to, without unduly influencing their decisions.

Also, `NotImplementedError` is designated for "Not on this platform", and so is not meant to be used this way.

Also, this deletes the comment, which is here to help someone that may have been away from Exercism since 2014, or even just 6 months ago, providing them the empty file, other than the comments, to help get them back on track.

kotp:
Thank you for exploring this @woarewe !

PR #1197, opened 2021-09-07
Trailing Whitespace, Contractions but most of all…

kotp:
Using references that match documentation, mailing lists and other
support channels when talking about class/module level methods.

### DIFF
--- a/exercises/concept/savings-account/.docs/introduction.md
+++ b/exercises/concept/savings-account/.docs/introduction.md
@@ -8,8 +8,6 @@ Sometimes we do not need the overhead of creating an object with state, so inste
 A module is very similar to a class (in fact, `Module` is `Class`' parent in the object hierarchy) - the main difference being that rather than using instance methods, we use class methods.
 Class methods start with `self.` and are directly called on a module.
 
-A module is very similar to a class (in fact, `Module` is `Classes` parent in the object hierarchy) - the main difference being that rather than using instance methods, we use class methods.
-Class methods start with `self.` and are directly called on a module. 
 For example:
 
 ```ruby
@@ -24,7 +22,7 @@ Speaker.echo("Hello")   #=> "Hello ... Hello"
 
 ## Loops
 
-There are several ways to write loops in Ruby, but as we tend to use enumeration rather than looping in general, the most commonly seen loop is the `while` loop:
+There are several ways to write loops in Ruby, but as we tend to use enumeration rather than looping in general, the most commonly seen loop perhaps is the `while` loop:
 
 ```ruby
 counter = 0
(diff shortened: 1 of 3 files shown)


exercises/concept/savings-account/.docs/introduction.md
```diff
@@ -21,7 +24,7 @@ Speaker.echo("Hello")   #=> "Hello ... Hello"
 
 ## Loops
 
-There are several ways to write loops in Ruby, but as we tend to use enumeration rather than looping in general, the most commonly seen loop is the `while` loop:
+There are several ways to write loops in Ruby, but as we tend to use enumeration rather than looping in general, the most commonly seen loop perhaps the `while` loop:
```
kotp, line 27:
```suggestion
There are several ways to write loops in Ruby, but as we tend to use enumeration rather than looping in general, the most commonly seen loop perhaps is the `while` loop:
```

iHiD (changes requested):
Regarding the change from `.` to `::`, I think this is a lot more confusing personally. I've always written this as `SavingsAccount.interest_rate` (and discussed it with others in the same way) as it's 1-1 with the actual code you then run. I can't really remember seeing `Integer::sqrt` in production, and if I do try it, Rubocop's defaults change it to `Integer.sqrt`.

The fact that we have to add a paragraph of explanation as to what the syntax means suggests to me that its not that obvious. So I'd suggest leaving this out, and instead adding a paragraph to the about.md of `basics` that explains how that you might see the `::` syntax in docs etc.

kotp:
For an example of where confusion comes, see #1203

iHiD:
> For an example of where confusion comes, see #1203

I don't think the confusion arose because of the fact it uses a `.` instead of `::` though. The confusion is because the exercise was originally written using class methods, then changed to instance methods, but the docs weren't updated.

iHiD:
Merging. Thanks @kotp! :)

kotp:
> > For an example of where confusion comes, see #1203
> 
> I don't think the confusion arose because of the fact it uses a `.` instead of `::` though. The confusion is because the exercise was originally written using class methods, then changed to instance methods, but the docs weren't updated.

I just knew that I wasn't seeing a consistency of our writing where developers see the `::` and `#` in the documentation when talking about module/class level methods compared to instance methods.  Further confusing things is that some methods are also capitalized names.  Fortunately, I do not remember any of those that do not take arguments.

iHiD:
> Further confusing things is that some methods are also capitalized name

This is definitely a bug. If you have an example of those we should fix them.

kotp:
I am not talking about Exercism having methods in Ruby that are capitalized.  I am saying that Ruby uses this mechanism.  See `Kernel#Integer`, `Kernel#Array`, etc.

Because the core library does this, it is not "wrong Ruby" to do so, and they are definitely useful to know.

Issue #1199, opened 2021-09-07
Unit tests (in Savings Account) use assert_equal


paulvt:
The unit tests in the Savings Account exercise tests equivalence of floats using `assert_equal` instead of `assert_in_delta` (as is suggested by the Minitest docs).
I am doing this track as a seasoned Ruby developer so I can start mentoring the track once I know the exercises but I got stuck here.

The annual balance update in this exercise can be calculated via:
```ruby
balance + (balance * rate.abs / 100.0)
```
Which, on my machine, for the test on `SavingsAccountTest#test_annual_balance_update_for_large_positive_start_balance` results in the expected `1016.2101016209999`.
However, if you do this slightly different, but mathematically the same:
```ruby
balance * (1.0 + rate.abs / 100.0)
```
It results in `1016.210101621` which then gets rejected by the test although it is obviously correct and even more precise!
The test relies on too specific float rounding here IMO, which a test should never do.

SleeplessByte:
@paulvt je hebt gelijk! 

If you have time, can you submit a PR?

kotp:
I noticed the same thing as I was reviewing this late last night as well, delta would definitely be appreciated.

paulvt:
I will have a look this week.

Steffan153:
This should be closed now.

kotp:
Thanks @Steffan153 .

PR #1201, opened 2021-09-10
Use `assert_in_delta` when testing float values

paulvt:
Set the delta as tight as possible, but allow for float approximations when performing float calculations in the exercises (i.e. `1_016.21` vs `1_016.20999999`). Still use `assert_equal` when methods should return
specific floats.

This addresses #1199.


paulvt:
Sorry, for some reason Solargraph didn't warn me about Rubocop indentation warnings? Rebased the branch.

kotp (approved):

Issue #1202, opened 2021-09-10
LogLineParser instructions and tests/stubs not synced


ntijoh-daniel-berg:
The instructions reference class methods, but the code stub include a constructor and instance methods (which the tests then use)

I propose changing the instructions as seen below:

Current: 
Implement the `LogLineParser.message` method to return a log line's message:

```ruby
LogLineParser.message('[ERROR]: Invalid operation')
# => "Invalid operation"
```

Proposed:
Implement the `LogLineParser#message` method to return a log line's message:

```ruby
LogLineParser.new('[ERROR]: Invalid operation').message
# => "Invalid operation"
```

PR #1203, opened 2021-09-10
changed instructions to be inline with tests and code stubs

ntijoh-daniel-berg:
implements changes proposed in #1202


iHiD:
Great. Thanks :)

PR #1204, opened 2021-09-10
Change number_of_wings to wings

soolodev:
I hope this helps!

### DIFF
--- a/exercises/concept/amusement-park/.docs/introduction.md
+++ b/exercises/concept/amusement-park/.docs/introduction.md
@@ -3,7 +3,7 @@
 ## Instance Variables
 
 Objects can hold their own state by setting _instance variables_.
-These variables live within an object and are normally used to store some internal state.
+These variables live within an object are normally used to store some internal state.
 
 Instance variables always start with an `@`.
 You can create a new instance variable by setting a variable starting with an `@`.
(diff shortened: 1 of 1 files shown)


SleeplessByte (approved):

SleeplessByte:
Thank you! This is great.

(Running CI now but need to go. Whomever can, merge at will).

exercises/concept/amusement-park/.docs/introduction.md
```diff
@@ -41,7 +41,7 @@ class Airplane
 end
 
 plane = Airplane.new
-plane.wings # returns 2
+plane.number_of_wings # returns 2
```
kotp, line 44: I prefer the `wings` as a method name, rather than `number_of_wings`.  Can we make the change the other way?

kotp (commented):
`@wings` communicates a plurality of wings, and so `number_of_` is redundant.  It should be, and would idiomatically be `def wings ; @wings ; end` (as multi-line, not as written here in this comment).

But the grammar change is good. 

This should also probably be two separate commits, since the grammar is one thing, and the method name correction is not, perhaps, strictly "grammar".  Besides, it would let us accept one while discussion happens with the other change.

kotp (approved):

Issue #1205, opened 2021-09-10
Practice exercises example solutions do not follow file structure/naming standard


ntijoh-daniel-berg:
Practice exercises’ example solution files do not follow the structure/naming specified in https://exercism.org/docs/building/tracks/practice-exercises

The docs use the `.meta/example.rb` (or `exemplar.rb`if it is a concept exercise) naming convention but the practice exercises use `.meta/solutions/exercise_name.rb`

(The test suite only covers practice tests, not concept exercises. I'd like to try to extend the test system to include concept exercises, but I'd rather wait until the practice exercises are updated).

angelikatyborska:
Changing the location of the example solutions for practice exercises should be easy to automate with a script. It just needs to move the files and then update the `config.json` files with new paths. But before we attempt this we would need @iHiD to confirm if he even wants the change in the Ruby track?

ntijoh-daniel-berg:
Seems like I never got notified of the previous comment, so I did not wait.
Was a learning experience even if it never gets used.

PR #1209, opened 2021-09-13
Renamed all exercise example files, added support for concept exercise tests

ntijoh-daniel-berg:
As discussed in #1205 

Lots of changes since all config.json files were updated. Also patched the test system to work with concept exercises.
Some exercises (such as amusement-park) needed manual renaming of files and "requires" (as in `require_relative`) to work with the test system.


exercises/practice/affine-cipher/.meta/config.json
```diff
@@ -18,10 +18,8 @@
...
     ],
-    "example": [
-      ".meta/solutions/affine_cipher.rb"
-    ]
+    "example": ".meta/example.rb"
   },
   "source": "Wikipedia",
   "source_url": "http://en.wikipedia.org/wiki/Affine_cipher"
-}
+}
```
kotp, line 25: A `JSON` file should end with an EOL as it is a text file.

kotp (commented):
The JSON files consistently are missing the last line.  That is to say, there is an EOL (end of line) character 0x0a at the last line, followed by additional data (text) at what should be the last line.

exercises/practice/affine-cipher/.meta/config.json
ntijoh-daniel-berg: Aha, I did not know that.  Fixed.
`configlet lint` now passes locally (but with warnings I think are outside the scope of this fix ).

kotp:
Trailing white space is definitely within the scope of this issue, otherwise it only creates more noise later.

The `has_pass?` is not idiomatic in Ruby, the `has_` being redundant in the `?` part of the name.  Personally, I would exempt the naming via the following comment (see my answer [here](https://stackoverflow.com/a/26350879):

```ruby
def has_pass? # rubocop:disable Naming/PredicateName
```

In this way, it signals to the student that the name is not idiomatic.  But also alerts the reader of that code (from the maintainers standpoint) that we might want to attend to the exercise that is causing this to happen.

kotp (approved):

kotp:
Very nice.   Thank you @ntijoh-daniel-berg !

ntijoh-daniel-berg:
Fixed rubocop errors.
I agree w/ the rubocop disable directive.

> Trailing white space is definitely within the scope of this issue, otherwise it only creates more noise later.

Yes, I agree, the linter warnings I was referencing was things like `The following concepts exist in the `prerequisites` array of a Practice Exercise in `... , but do not exist in the top-level `concepts` array`

kotp:
Can you create an issue for those, as reported?  That will put them on the radar, at least.

ntijoh-daniel-berg:
Done.

iHiD:
@ntijoh-daniel-berg Thanks for this.

Note: The changes from `attendee.rb` to `amusement_park_improvements.rb` will result in people getting duplicate files when working in the editor. I'll revert that in a separate PR.

@kotp Please can you tag me in any issues that change things like the `config.json` files, so I can analyze the consequences before we merge. Thanks 🙂

kotp:
> tag me in any issues that change things like the `config.json` files…

Ooh… yeah!

kotp:
@ihid That would be a lovely thing to have happen automatically, I think, if possible.  Any changes to a certain list of files automatically alerting specific users or group of users.  Might be a good feature request for github.

iHiD:
We could add me as a codeowner of `config.json` files. I'm not sure of the syntax to do that recursively. @ErikSchierboom Do you know?

PR #1210, opened 2021-09-14
Fix enumeration

luilver:


kotp (approved):
The numbering would render correctly, I think, automatically, but this is good for the ones that read in their local environment!  Thank you!

PR #1212, opened 2021-09-14
Fix a missconception

luilver:
A positive number is greater than 0, a non-negative balance less than 1000 dollar includes a zero balance.

### DIFF
--- a/exercises/concept/savings-account/.docs/instructions.md
+++ b/exercises/concept/savings-account/.docs/instructions.md
@@ -3,7 +3,7 @@
 In this exercise you'll be working with savings accounts. Each year, the balance of a savings account is updated based on the interest rate. The interest rate the bank gives depends on the amount of money in the accounts (its balance):
 
 - -3.213% for a negative balance.
-- 0.5% for a positive balance less than `1000` dollars.
+- 0.5% for a non-negative balance less than `1000` dollars.
 - 1.621% for a positive balance greater or equal than `1000` dollars and less than `5000` dollars.
 - 2.475% for a positive balance greater or equal than `5000` dollars.
 
--- a/exercises/concept/savings-account/.docs/instructions.md
+++ b/exercises/concept/savings-account/.docs/instructions.md
@@ -6,6 +6,7 @@ In this exercise you'll be working with savings accounts. Each year, the balance
 - 0.5% for a non-negative balance less than `1000` dollars.
 - 1.621% for a positive balance greater or equal than `1000` dollars and less than `5000` dollars.
 - 2.475% for a positive balance greater or equal than `5000` dollars.
+- 0% for a balance of `0` dollars (it neither earns interest, nor is penalised).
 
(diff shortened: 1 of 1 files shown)


kotp:
An interest rate multiplied by 0 will always be 0 (zero).  So not sure if it is applicable.

iHiD:
Yeah, I'm not sure about this. Zero bank balance gets zero interest, so the rate is sort of irrelevant. What do others think?

iHiD:
I've asked a couple of people and the consensus is that the original way its worded is fine. @luilver Could you explain a little more about what you think is wrong with the current text pls? Is it that it doesn't explicitly cater for zero?

luilver:
> I've asked a couple of people and the consensus is that the original way its worded is fine. @luilver Could you explain a little more about what you think is wrong with the current text pls? Is it that it doesn't explicitly cater for zero?

Sure @iHiD, I'm a completely noob on accounting terms, so if 0 is positive within bank account or something like that, please, completely discard my proposition. I was talking from a math point of view where Zero exists and it is not a positive number (integer actually), zero use to be trait as a non negative value.

iHiD:
I think the key here is that if you have zero in your bank, you get zero interest, but also don't get charged.

So maybe what would help would be an extra sentence that says "If you have a balance of zero, you neither accrue interest, nor get charged".

iHiD:
(To put it another way, zero is neither positive nor negative in the context of this exercise).

luilver:
That's was precisely my first approach to this exercise: Taking zero as a special case. Later I cheat it (watching inside rspec) and make my code pass on tests. You'd probably be right about zero within this exercise's scope so you may ignore my proposal.

luilver:
Again, my point of conflict here was the term `positive` where zero has definitively no way to be on.

exercises/concept/savings-account/.docs/instructions.md
```diff
@@ -3,7 +3,7 @@
 In this exercise you'll be working with savings accounts. Each year, the balance of a savings account is updated based on the interest rate. The interest rate the bank gives depends on the amount of money in the accounts (its balance):
 
 - -3.213% for a negative balance.
-- 0.5% for a positive balance less than `1000` dollars.
+- 0.5% for a non-negative balance less than `1000` dollars.
 - 1.621% for a positive balance greater or equal than `1000` dollars and less than `5000` dollars.
 - 2.475% for a positive balance greater or equal than `5000` dollars.
```
iHiD, line 8:
```suggestion
- 2.475% for a positive balance greater or equal than `5000` dollars.
- A balance of `0` dollars neither earns interest, nor is penalised.
```

kotp (approved):

kotp:
Thank you for the work on this @luilver !

luilver:
Thank you guys for such an excellent tool like exercism I'll always be glad to contribute with.

iHiD:
Thanks @luilver!

PR #1229, opened 2021-09-21
Assembly Line: Use assert_in_delta when comparing float values

s2k:
With this push request we avoid failing tests in submissions that use a slightly different order of arguments in the tested method, which causes a small difference in the calculated result.

I set the the delta to a values that makes sense for the problem being solved.

This pull requests is created following https://github.com/exercism/ruby/pull/1201, which covers the same issue for a different exercise.

### DIFF
--- a/exercises/concept/assembly-line/assembly_line_test.rb
+++ b/exercises/concept/assembly-line/assembly_line_test.rb
@@ -3,43 +3,43 @@ require_relative 'assembly_line'
 
 class AssemblyLineTest < Minitest::Test
   def test_production_rate_per_hour_for_speed_one
-    assert_in_delta 221.0, AssemblyLine.new(1).production_rate_per_hour, 0.000_000_0001
+    assert_in_delta 221.0, AssemblyLine.new(1).production_rate_per_hour, 0.0001
   end
 
   def test_production_rate_per_hour_for_speed_two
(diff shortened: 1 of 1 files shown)


exercises/concept/assembly-line/assembly_line_test.rb
```diff
@@ -3,43 +3,43 @@
 
 class AssemblyLineTest < Minitest::Test
   def test_production_rate_per_hour_for_speed_one
-    assert_equal 221.0, AssemblyLine.new(1).production_rate_per_hour
+    assert_in_delta 221.0, AssemblyLine.new(1).production_rate_per_hour, 0.000_000_0001
```
kotp, line 6: I like the delta check.  Should it be to 3 decimal places, or perhaps 4, since we seems to be working with percentages?  Low precision problem, high value delta.

kotp (dismissed):
Approving, but feel free to modify to not "compares to other exercises" but "makes sense for the problem being solved, instead.

exercises/concept/assembly-line/assembly_line_test.rb
s2k: Thanks for the feed back. I updated both, the asserts and the PR description.

kotp (approved):
Thank you for the work on this @s2k!

PR #1190, opened 2021-09-04
Savings Account: Correct average_negative_start_balance test

jasylwong:
There is an error in the Savings Account exercise, in the `test_annual_balance_update_for_avarage_negative_start_balance` test. This PR:
- makes the starting balance negative
- updates the associated expected result, since negative starting balances should be subject to the negative interest rate of 3.213%.
- corrects the misspellings of 'average' in the test method names

PR #1233, opened 2021-09-26
Savings-Account: Fix copy/paste error from positive to negative test.

k3dB:
This is a duplication of #1190.  It has been a few weeks since the original post, and over a week since the last comment was posted.  I ran across this same issue today while working on this exercise.  It appears that the typos have been fixed on the main branch already, so I just included the number fixes (i.e. making the test honest about testing a negative balance).

### DIFF
--- a/exercises/concept/savings-account/savings_account_test.rb
+++ b/exercises/concept/savings-account/savings_account_test.rb
@@ -81,7 +81,7 @@ class SavingsAccountTest < Minitest::Test
   end
 
   def test_annual_balance_update_for_average_negative_start_balance
-    assert_in_delta 1_016.21, SavingsAccount.annual_balance_update(1_000.0), 0.01
+    assert_in_delta -1_032.13, SavingsAccount.annual_balance_update(-1_000.0), 0.01
   end
 
   def test_annual_balance_update_for_large_negative_start_balance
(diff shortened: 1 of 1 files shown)


exercises/concept/savings-account/savings_account_test.rb
```diff
@@ -81,7 +81,7 @@ def test_annual_balance_update_for_small_negative_start_balance
   end
 
   def test_annual_balance_update_for_average_negative_start_balance
-    assert_in_delta 1_016.21, SavingsAccount.annual_balance_update(1_000.0), 0.01
+    assert_in_delta -1_032.13, SavingsAccount.annual_balance_update(-1_000.0), 0.01
```
kotp, line 84:
```suggestion
    assert_in_delta -1_032.13, SavingsAccount.annual_balance_update(-1_000.0), 0.01 # rubocop:disable Lint/AmbiguousOperator
```

This is how we might state explicitly that the syntax is what we prefer specifically here, even if generally we agree with using parenthesis around negative numbers.

Otherwise, we should use

```suggestion
    assert_in_delta(-1_032.13), SavingsAccount.annual_balance_update(-1_000.0), 0.01
```

kotp (changes requested):
The use of parenthesis is warranted for the delta because it is a negative number.  It communicates well when it has parenthesis.

If you do think that having no parenthesis around, then feel free to give a Rubocop directive at the end of the line for this saying it is purposeful.   The cop name is "Lint/AmbiguousOperator".

The directive can be written as:

`# rubocop:disable Lint/AmbiguousOperator`

If placed on the line before then we should re-enable after as well.

kotp (approved):

kotp:
Thank you @k3dB !  There were so many different ways to go with that last change.  That was one I had not considered.

PR #1235, opened 2021-09-28
Uses underscore consistently for fractional part of floats

s2k:
Use the underscore consistently to separate groups of 3 digits in the rational part of numbers, starting at the decimal point and moving right.

This way the decimal places are separated into groups that would be prefixed with 'milli', 'micro', 'nano' etc, as often ist used in (physical) units (µ gram, milli second, nano meter…).

### DIFF
--- a/exercises/concept/savings-account/savings_account_test.rb
+++ b/exercises/concept/savings-account/savings_account_test.rb
@@ -59,7 +59,7 @@ class SavingsAccountTest < Minitest::Test
   end
 
   def test_annual_balance_update_for_small_positive_start_balance
-    assert_in_delta 0.000_001_005, SavingsAccount.annual_balance_update(0.000_001), 0.000_000_0001
+    assert_in_delta 0.000_001_005, SavingsAccount.annual_balance_update(0.000_001), 0.000_000_000_1
   end
 
   def test_annual_balance_update_for_average_positive_start_balance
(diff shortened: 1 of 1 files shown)


exercises/concept/savings-account/savings_account_test.rb
```diff
@@ -59,33 +59,33 @@ def test_annual_balance_update_for_empty_start_balance
...
   def test_annual_balance_update_for_large_positive_start_balance
-    assert_in_delta 1_016.2_101_016_209_999, SavingsAccount.annual_balance_update(1_000.0001), 0.0_000_000_000_01
+    assert_in_delta 1_016.2_101_016_209_999, SavingsAccount.annual_balance_update(1_000.000_1), 0.000_000_000_001
   end
 
   def test_annual_balance_update_for_huge_positive_start_balance
     assert_in_delta 920_352_587.26_744_292_868_451_875,
       SavingsAccount.annual_balance_update(898_124_017.826_243_404_425),
-      0.00_000_000_000_000_001
+      0.000_000_000_000_000_01
```
kotp, line 76: Can you move this up one line, as is consistent with the others as well, and line 42 could use the underline treatment as well.

kotp (commented):
Thanks @s2k .

exercises/concept/savings-account/savings_account_test.rb
s2k: Sure.

SleeplessByte (approved):

PR #1236, opened 2021-09-29
Use assert in delta throughout

s2k:
As discussed briefly in slack, start using `assert_in_delta` consistently when comparing floats.


kotp (approved):
Looks good!  Thank you!

PR #1239, opened 2021-10-01
Fix interest rate

luilver:
According to new description:
```
0% for a balance of 0 dollars (it neither earns interest, nor is penalised).
```
expected value should be 0 instead of 0.5

### DIFF
--- a/exercises/concept/savings-account/savings_account_test.rb
+++ b/exercises/concept/savings-account/savings_account_test.rb
@@ -3,7 +3,7 @@ require_relative 'savings_account'
 
 class SavingsAccountTest < Minitest::Test
   def test_minimal_first_interest_rate
-    assert_in_delta 0.5, SavingsAccount.interest_rate(0), 0.000_1
+    assert_in_delta 0, SavingsAccount.interest_rate(0), 0.000_1
   end
 
   def test_tiny_first_interest_rate


kotp:
I thought there should be an issue linked to this PR?

luilver:
Yes, probably : ) I'm not too familiarized yet with the flow. Should I read some good docs ?

exercises/concept/savings-account/savings_account_test.rb
```diff
@@ -3,7 +3,7 @@
 
 class SavingsAccountTest < Minitest::Test
   def test_minimal_first_interest_rate
-    assert_in_delta 0.5, SavingsAccount.interest_rate(0), 0.000_1
+    assert_in_delta 0, SavingsAccount.interest_rate(0), 0.000_1
```
kotp, line 6: The interest rate for a non-negative balance less than 1000 is not 0% but 0.5%.  This test will fail, and rightly so.  The interest rate does change based on the balance, but the interest rate is not 0% in any circumstance given.  The interest amount ends up being 0.00 though, on a balance of zero.  For our purposes, non-negative includes zero.

luilver: Well, I missed something here. 

If this case includes zero:
```
0.5% for a non-negative balance less than 1000 dollars.
```
Why then is necessary this one?:
```
0% for a balance of 0 dollars (it neither earns interest, nor is penalised).
```

I guess it should be something like:
```
-3.213% for a negative balance.
0% for a balance of 0 dollars (it neither earns interest, nor is penalised).
0.5% for a positive balance less than 1000 dollars.
1.621% for a positive balance greater or equal than 1000 dollars and less than 5000 dollars.
2.475% for a positive balance greater or equal than 5000 dollars.
```

And thus, corresponding test should be changed too.

kotp: I  do not think the interest rate itself changes for 0.  The interest rate for a zero balance could be any number, once it is multiplied the effect would be a yield of 0.  But the rate is stated as a non-negative balance to 1_000 has a rate as given.

So `-3.213%` for a negative balance, for a balance above a negative balance you get `0.5%` up to, but not including 1000.

There is no need for a special rate specifically for 0, that tends to take care of itself, and can be anything, because anything times 0 is 0. The edge case is there in theory, but effectively it takes care of itself.

The one that says 0% for a balance of 0 dollars is not necessary, in my opinion.

luilver:
Closing this because it has being well defined.

kotp:
Thank you for the discussion and exploration @luilver !

PR #1242, opened 2021-10-03
Proposed new example for Palindrome Products

gemp:
This example_2 is ~~about 60%~~ more than two times faster than the present one.

* It still uses a double loop on `@range` because `repeated_combination(2)` is waaay slower.
* ~~`r1 * r2 if r2 >= r1` prevents to calculate `2*1` when we already have `1*2` mimicking `repeated_combination`~~ Redefined the range instead.
* It is much more efficient to recalculate the factors afterwards than to store them during the loops where it creates a lot of unnecessary `Palindrome` objects (or even only arrays of `[product, [i, j].sort]`).
* And the way they are calculated prevents `palindrome.factors << [i, j].sort; palindrome.factors.uniq!` just with `div <= number`.
* `n == n.to_s.reverse.to_i` is ~15% faster than `n.to_s == n.to_s.reverse`.
* Finally I found that adding a method `palindrome?` or monkey patching `Integer` slows down stuff as well as being overkill as it's used only once; it makes things clearer but it has a cost.

Mentored by @kotp ❤️

### DIFF
--- a/exercises/practice/palindrome-products/.meta/example_2.rb
+++ b/exercises/practice/palindrome-products/.meta/example_2.rb
@@ -1,8 +1,4 @@
-class Palindrome
-  attr_reader :value, :factors
-  private
-  def initialize(palindrome) @value, @factors = palindrome end
-end
+Palindrome = Struct.new(:value, :factors)
 
 class Palindromes
 
@@ -24,7 +20,7 @@ class Palindromes
   end
 
   def palindrome_and_factors(palindrome)
-    [palindrome, factors(palindrome)]
+    Palindrome.new palindrome, factors(palindrome)
   end
 
   public
@@ -41,11 +37,11 @@ class Palindromes
   end
 
   def largest
-    @largest ||= Palindrome.new palindrome_and_factors(candidates.max)
+    @largest ||= palindrome_and_factors(candidates.max)
   end
 
   def smallest
-    @smallest ||= Palindrome.new palindrome_and_factors(candidates.min)
+    @smallest ||= palindrome_and_factors(candidates.min)
   end
 
 end
--- a/exercises/practice/palindrome-products/.meta/example_2.rb
+++ b/exercises/practice/palindrome-products/.meta/example_2.rb
@@ -1,9 +1,7 @@
 class Palindrome
   attr_reader :value, :factors
   private
-  def initialize(palindrome)
-    @value, @factors = palindrome
-  end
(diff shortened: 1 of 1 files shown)


kotp:
@exercism/ruby Should this be `example.rb` or should we want this to be in mentor notes for this exercise?

exercises/practice/palindrome-products/.meta/example_2.rb
```diff
@@ -0,0 +1,52 @@
+class Palindrome
+  attr_reader :value, :factors
+  private
+  def initialize(palindrome)
+    @value, @factors = palindrome
+  end
+end
```
kotp, line 7: This class is not styled like the other class.  We should have vertical spacing and organization that is consistent, because inconsistencies draw attention (which is sometimes wanted) and suggests that there is "something different here, pay attention" when there really is not.

exercises/practice/palindrome-products/.meta/example_2.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  attr_reader :range, :candidates
+
+  def initialize(min_factor: 1, max_factor: 9)
+    @range = (min_factor..max_factor)
+  end
+
+  def factors(palindrome)
+    range.each_with_object([]) do |number, factors|
+      div, mod = palindrome.divmod(number)
```
kotp, line 21: We use full names just about everywhere else, we could use `divisor` or `divisor` or whatever it is this abbreviation means and `modulo` or, is it `modulus`, here (which can make it easier for those unfamiliar with these things to find the words in their dictionaries).  Code does not have to read cryptically like code.

gemp:
Thanks @kotp for you approval, it's been an amazing journey over a very stupid and boring problem 😘

But I'd like to mention that it's not an optimal solution as I privileged speed over memory (most of community solutions don't care about both). It's only numbers here but when you apply `.uniq` to `candidates` you got respectively for the ranges tested (plus the 9999..1000 range) from 13 to 9, 115 to 74, 1.239 to 655 and 12.276 to 5.788. It can make a huge memory difference when applied to larger objects. For a relatively small speed toll (~10%) line 36 could be:
```ruby
        if candidate == candidate.to_s.reverse.to_i && candidates.none?(candidate)
```
Or simply invoque `.uniq` at the end of the loop if it's the same for the memory (I don't know).

As we're talking about exemplarity here this seems interesting.

PR #1243, opened 2021-10-04
Update acronym_test.rb

bolandross:
There seem to be 7 in-browser-tests on Exercism, but in the CLI version there is only 1 (the other tests are skipped).

I wonder what is right now? Should there only be one easy test, since this is one of the beginner tracks? The tests seem to be directed to more advanced programmers.

### DIFF
--- a/exercises/practice/acronym/acronym_test.rb
+++ b/exercises/practice/acronym/acronym_test.rb
@@ -4,38 +4,31 @@ require_relative 'acronym'
 # Common test data version: 1.7.0 cacf1f1
 class AcronymTest < Minitest::Test
   def test_basic
-    # skip
     assert_equal "PNG", Acronym.abbreviate('Portable Network Graphics')
   end
 
   def test_lowercase_words
-    skip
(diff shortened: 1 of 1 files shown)


kotp:
In the README, does it mention to unskip the tests, or in the test file itself?  At some point, early on, when using the CLI this information is shared.  Concept exercises do not have the skips, while the other exercises do.  On the platform they are all "unskipped" for testing automatically.

They are both correct, unskip the tests as you work through it emulating TDD, or disable them all by maybe writing:

```ruby
  def skip() end
```

bolandross:
> In the README, does it mention to unskip the tests, or in the test file itself? At some point, early on, when using the CLI this information is shared. Concept exercises do not have the skips, while the other exercises do. On the platform they are all "unskipped" for testing automatically.

Hmm strange, I checked the help and readme file again, even the test file but couldn't find any information regarding that. I was confused further since other test files didn't have the skip. But yeah, the TDD emulation part of your explanation makes sense. 

I am new to exercism, so I totally believe that there might be something important I haven't read. 😅

PR #1256, opened 2021-10-21
`log-line-parser`: minor verbiage updates

sudomateo:
As I was going through the exercise, I noticed these inconsistencies in the text.


kotp:
> You're welcome! I'll submit more as I find things. Sorry for the small change, if I didn't do it as I went through the exercise I'd have forgotten about it.

Definitely nothing wrong with fixing trailing white space and using the right words.  And small is good, sometimes better than large.

PR #1259, opened 2021-10-28
Fix: year and number no longer public

kotp:
Includes reorganizing to keep all private methods together.


kytrinyx (approved):
This makes sense to me.

kotp:
Thank you for the review!

PR #1263, opened 2021-11-01
Fix count of seconds in minute

zakharoff:


iHiD:
Hi @zakharoff - The change isn't actually correct. If you type `110` into a microwave, it means `1min, 10secs`. If you type `90` into a microwave, it means `1min, 30secs`

PR #1272, opened 2021-11-26
Fix instructions typo in part 1

eepmoi:


exercises/concept/log-line-parser/.docs/instructions.md
```diff
@@ -25,7 +25,7 @@ LogLineParser.new('[ERROR]: Invalid operation').message
 Any leading or trailing white space should be removed:
 
 ```ruby
-LogLineParser.new('[WARNING]:  Disk almost full\r\n').message
+LogLineParser.new("[WARNING]:  Disk almost full\r\n").message
```
kotp, line 28: Single quotes are fine.

kotp: Indeed, not only fine, but required, as with double quotes those characters would be interpreted instead of showing as escaped.

kotp: Never mind, that is exactly the problem, is that they are escaped and not literally what they should be, in order to be stripped.  Good catch!

kotp (approved):

PR #1274, opened 2021-12-02
simple-calculator: fix instructions typo in part 2

eepmoi:
Update instructions typo, part 2 is meant to test unsupported operation vs invalid string argument.

Might cause confusion with beginners who have the sequencing of their error handling around the wrong way to start.

### DIFF
--- a/exercises/concept/simple-calculator/.docs/instructions.md
+++ b/exercises/concept/simple-calculator/.docs/instructions.md
@@ -26,7 +26,7 @@ The main method for implementation in this task will be the class method `Simple
 Update the `SimpleCalculator.calculate()` method to raise an `UnsupportedOperation` exception for unknown operation symbols.
 
 ```ruby
-SimpleCalculator.calculate(1, '2', '-')
+SimpleCalculator.calculate(1, 2, '-')
 # => Raises an UnsupportedOperation
 ```
 


exercises/concept/simple-calculator/.docs/instructions.md
```diff
@@ -26,7 +26,7 @@ The main method for implementation in this task will be the class method `Simple
 Update the `SimpleCalculator.calculate()` method to raise an `UnsupportedOperation` exception for unknown operation symbols.
 
 ```ruby
-SimpleCalculator.calculate(1, '2', '-')
+SimpleCalculator.calculate(1, 2, '-')
```
kotp, line 29: This is not an "illegal operation" in that intuitively it is correct, but indeed this is surely keeping in line with what is stated by using an integer.

However supplying a string as an argument will ensure that we still should raise the `UnsupportedOperation` even if the student decides to "stretch" and create a `-` method for subtraction.  I mean to say that even if the tests change or the scope broadens to use subtraction, then having the string should probably still raise that exception.

The part that I might find to be unclear is the mention of only `+` on line 5 and not the additional 2 operators that we are showing on the example line.

kotp (commented):
I do not believe this is a typographical error but is on purpose.

exercises/concept/simple-calculator/.docs/instructions.md
eepmoi: Fair enough - though assuming the student is incrementally coding the parts (ie hasn't done part 3 yet) having a string would not get picked up by the `UnsupportedOperation` logic as they would likely be comparing the operator only at this point against `ALLOWED_OPERATORS`.

I guess my point is that looking at the sample method call it seems we're overloading the ask without explicitly asking the student to handle both use cases - string and operator.

kotp (approved):
I could see it being more focused on the operator instead of the string
being a potential distraction, and the reason for this failure.

I am going to approve and bring it in.

PR #1276, opened 2021-12-06
Practice exercises should emulate TDD with skips

kotp:
The first skip should be commented (unskipped) while the remainders are
active.

Ref: #1189


iHiD:
Thanks :)

PR #1278, opened 2021-12-11
fix grammar error "an" -> "a"

amandarito:
noticed this grammar mistake while doing the exercise, pretty sure "whole" needs "a" before it instead of "an". thanks!


kotp (approved):

kotp:
Thank you @amandarito I have approved, and will bring the change in once checks are completed.

PR #1281, opened 2022-01-06
Moviegoer: remove unnecessary "attr_reader"

AngelGallegosE:
The `attr_reader :age, :member` line is not necessary since we don't need to expose the attributes.

### DIFF
--- a/exercises/concept/moviegoer/moviegoer.rb
+++ b/exercises/concept/moviegoer/moviegoer.rb
@@ -3,8 +3,6 @@ class NotMovieClubMemberError < RuntimeError
 end
 
 class Moviegoer
-  attr_reader :age, :member
-
   def initialize(age, member: false)
     @age = age
     @member = member


kotp:
Thank you for the patch!  It is definitely good to see someone new here making contributions!

exercises/concept/moviegoer/moviegoer.rb
```diff
@@ -3,8 +3,6 @@ class NotMovieClubMemberError < RuntimeError
 end
 
 class Moviegoer
-  attr_reader :age, :member
-
```
kotp, line 7: While it is true that these are exposed as they are public methods, but having these are probably scaffolding indicating that the student can, or should, use them.

Should we instead make them private?  Leaving that scaffolding available to the author of the exercise solution, while not exposing them, as stated in the PR description.

I am definitely willing to change this, either removing it (as we do not talk about attributes in the concept) but we might need to check if attributes are talked about prior to this concept exercise to even know if it is something appropriate to scaffold, but note somewhere that they can be used, or direct the student in another way.

kotp (commented):
While I am good in removing this code, the description of the PR indicates the problem is that they are exposed, not necessarily that they exist.  So we should explore that, why that code is here in the first place as scaffolding.

exercises/concept/moviegoer/moviegoer.rb
iHiD: Good think, @kotp! I think I'm happy to remove them as they're not mentioned in the concept and not used by the tests, I expect they got left over from some internal testing as this exercise was been written/tested.

kotp: I think we will go with removal, then, as opposed to contending with exposure only.

kotp (approved):
Approving for the removal option, as opposed to the exposure issue.

PR #1282, opened 2022-01-08
Fix typo in boutique inventory improvements

gabo-cs:
I think that "previously" is the correct word here


kotp (approved):
Since we have talked about this directly, I am approving now so feel free to rebase and merge, but also feel free to apply changes and I will review again.

gabo-cs:
> Since we have talked about this directly, I am approving now so feel free to rebase and merge, but also feel free to apply changes and I will review again.

Yes, thanks for the review and the tips, those will come in handy for future related contributions, how about we merge this as is? (I believe you have the right permissions to do so)

kotp:
Thank you for the patch!

And the chat!

PR #1288, opened 2022-01-19
.DS_Store removed from repository

kotp:
This file should be excluded in the users (or system) excludes file.

We would not add it to the project `.gitignore` file as it is not
related to anything in this project.


kotp:
Do we need to have @exercism/reviewers added here for the Ruby track?

ErikSchierboom (approved):

ErikSchierboom:
> Do we need to have https://github.com/orgs/exercism/teams/reviewers added here for the Ruby track?

Possibly.

PR #1289, opened 2022-01-23
Savings Account: Adds test for negative current balance edge case

apmiller108:
A negative `current_balance` argument for `years_before_desired_balance` will result in an endless loop for the current example. This adds a test for the edge case. We get an endless loop since a negative balance will accrue negative interest, and will therefore get further from the `desired_balance` each year (assuming the `desired_balance` is greater than the `current_balance`).

Admittedly, raising an error may not be the desired approach for this exercise depending on where on the "Learning Exercise" track this exercise falls. It seems however, that the simple calculator exercise should have already introduced the "Raising Exceptions" concept. Anyway, I'm open a different approach like just returning `Float::INFINITY` as we done in [this solution](https://exercism.org/tracks/ruby/exercises/savings-account/solutions/clairedeshayes).

I wasn't sure about what value to pass to `timeout`, but `0.1` seconds seemed like ample time :shrug:

### DIFF
--- a/exercises/concept/savings-account/.meta/exemplar.rb
+++ b/exercises/concept/savings-account/.meta/exemplar.rb
@@ -22,6 +22,7 @@ module SavingsAccount
   end
 
   def self.years_before_desired_balance(current_balance, desired_balance)
+    raise ArgumentError, 'Current balance cannot be negative' if current_balance.negative?
     years = 0
     while current_balance < desired_balance
       current_balance = annual_balance_update(current_balance)
--- a/exercises/concept/savings-account/.meta/exemplar.rb
+++ b/exercises/concept/savings-account/.meta/exemplar.rb
@@ -22,7 +22,7 @@ module SavingsAccount
   end
 
   def self.years_before_desired_balance(current_balance, desired_balance)
-    raise ArgumentError, 'Current balance cannot be negative' if current_balance.negative?
+    raise ArgumentError, 'Current balance must be positive' unless current_balance.positive?
     years = 0
     while current_balance < desired_balance
       current_balance = annual_balance_update(current_balance)
--- a/exercises/concept/savings-account/savings_account_test.rb
+++ b/exercises/concept/savings-account/savings_account_test.rb
@@ -105,10 +105,13 @@ class SavingsAccountTest < Minitest::Test
   end
 
   def test_years_before_desired_balance_for_negative_current_balance
-    assert_raises(ArgumentError, 'Current balance cannot be negative') do
+    assert_raises(ArgumentError, 'Current balance must be positive') do
       Timeout.timeout(0.1) do
         SavingsAccount.years_before_desired_balance(-1.0, 1.0)
       end
+    rescue Timeout::Error
+      raise RuntimeError, 'Impossible Solution, raise an ArugmentError if current '\
+                          'balance is not positive.'
     end
   end
 end
--- a/exercises/concept/savings-account/savings_account_test.rb
+++ b/exercises/concept/savings-account/savings_account_test.rb
@@ -1,4 +1,5 @@
 require 'minitest/autorun'
+require 'timeout'
 require_relative 'savings_account'
 
 class SavingsAccountTest < Minitest::Test
@@ -101,5 +102,13 @@ class SavingsAccountTest < Minitest::Test
 
   def test_years_before_desired_balance_for_large_difference_between_start_and_desired_balance
     assert_equal 85, SavingsAccount.years_before_desired_balance(2_345.67, 12_345.678_9)
+  end
+
+  def test_years_before_desired_balance_for_negative_current_balance
+    assert_raises(ArgumentError, 'Current balance cannot be negative') do
+      Timeout.timeout(0.1) do
+        SavingsAccount.years_before_desired_balance(-1.0, 1.0)
+      end
+    end
   end
 end


exercises/concept/savings-account/savings_account_test.rb
```diff
@@ -102,4 +103,12 @@ def test_years_before_desired_balance_for_large_start_balance
   def test_years_before_desired_balance_for_large_difference_between_start_and_desired_balance
     assert_equal 85, SavingsAccount.years_before_desired_balance(2_345.67, 12_345.678_9)
   end
+
+  def test_years_before_desired_balance_for_negative_current_balance
+    assert_raises(ArgumentError, 'Current balance cannot be negative') do
+      Timeout.timeout(0.1) do
+        SavingsAccount.years_before_desired_balance(-1.0, 1.0)
```
kotp, line 110: This also has the problem of a balance that is 0.  More than "can not be negative", it should probably state that it should be positive.

For communicating the problem. we might also rescue the timeout and raise `RuntimeError, 'Impossible Solution'`

It might look like this:

```ruby
  def test_years_before_desired_balance_for_negative_current_balance
    assert_raises(ArgumentError, 'Current balance cannot be negative') do
      Timeout.timeout(0.1) do
        SavingsAccount.years_before_desired_balance(0, 1.0)
      end
    rescue Timeout::Error
      raise RuntimeError, 'Impossible Solution, raise an ArgumentError if current balance is not positive.'
    end
  end
```

exercises/concept/savings-account/.meta/exemplar.rb
```diff
@@ -22,6 +22,7 @@ def self.annual_balance_update(balance)
   end
 
   def self.years_before_desired_balance(current_balance, desired_balance)
+    raise ArgumentError, 'Current balance cannot be negative' if current_balance.negative?
```
kotp, line 25:
```suggestion
    raise ArgumentError, 'Current balance must be positive' if current_balance <= 0
```

Because 0 will give the same problem.

kotp (changes requested):
Thank you for looking at this, and submitting the pull request.  I have noticed while reviewing this that even 0 will cause the timeout.

I also noticed the exception raised while testing this during review, and the Timeout exception is informative, but the mechanics have not been introduced yet.  So rescuing that and re-raising may be better communication for where we are at in the concepts.

exercises/concept/savings-account/savings_account_test.rb
kotp: I think I prefer the "must be positive" rather than "cannot be negative", as that includes 0.

exercises/concept/savings-account/.meta/exemplar.rb
iHiD: Maybe this reads even nicer?

```suggestion
    raise ArgumentError, 'Current balance must be positive' unless current_balance.positive?
```

kotp: @apmiller108 I do like this version better.

But we should also weigh, like @iHiD asked about, if this is a product of the solution, rather than a problem with the exercise itself.  In other words, it may be a mentoring point instead of a test (potentially out of scope of the exercise).

Should it be a mentor note instead?

apmiller108: Thanks for the feedback and good suggestions. I've gone ahead and made the changes. As to whether or not this should be a mentor note, as @iHiD pointed out, I actually didn't know there were mentor notes (first PR to exercism here). Is that the `.docs/hints.md`? I suppose this case might be more appropriate as a note; however, it does make for a good learning topic and I see many community solutions don't consider this case -- those that didn't receive mentoring, perhaps. But, no, this isn't a problem with the exercise itself, rather an edge case where the caller can get into trouble with balances less than zero, which are expected to exist.

ErikSchierboom (changes requested):
Thanks for submitting a PR @apmiller108!
While I understand the reasoning behind the change, I'd argue not to merge this change for the following reasons.

The goal of this exercise is to teach students about the `modules` and `loops` concepts (see its [config.json entry](https://github.com/exercism/ruby/blob/main/config.json#L69)).
None of its [prerequisites](https://github.com/exercism/ruby/blob/main/config.json#L70) are about exceptions/errors, so we can't expect the student to know about that concept.
They'll thus probably be confused and unsure what is expected of them if we require them to work with exceptions/errors.
There are two solutions to this problem:

1. Add [exceptions](https://github.com/exercism/ruby/blob/main/config.json#L104) as a prerequisite
2. Don't use exceptions/errors

I would strongly be in favor of the second option.
Concept Exercises (which this is one) should be focusing _solely_ on the concepts it aims to teach.
Therefore, any non-essential concepts should not be used.
My argument is that exceptions/errors are not essential to this exercise.
Another way to think of this is: our exercises don't strive to be exhaustive; they're not production-grade, hardened code.
In production code, you'd definitely want to handle a negative (or zero) balance, but for Concept (or Practice) Exercises it is totally fine (and often desirable) to simply not deal with exhaustive error checking.
The student won't ever have to deal with negative (or zero) balances simply because we design our tests to not use those values.
Does that make sense?

p.s. I selected "Request changes" to prevent this PR from being merged

apmiller108:
@ErikSchierboom Yup, that makes sense. Interesting, I hadn't seen the config.json which details the concepts and prerequisites. I'll have to remember to look at that next time. Agreed that this test case is likely beyond the learning scope of this exercise given where it sits in the learning track. Closing this PR. Thanks for detailed explanation and reviews.

iHiD:
@apmiller108 Thanks for being awesome :)

kotp:
Thanks @apmiller108 just seeing what your decision was (I hate closing these, even though that is as important a decision as bringing it in) and am happy with the result.

ErikSchierboom:
@apmiller108 My pleasure! These sort of rules are often not immediately clear.

PR #1291, opened 2022-01-25
Fix wrong classname in "Word Count" exercise

AngelGallegosE:
I think we are using the wrong class name in the unit tests. it should be `WordCount`, not `Phrase`.

- In the test file, there is `require_relative 'word_count'` line
- The name of the class in the test is: `WordCountTest`

So there is no reference to `Phrase` in any place

### DIFF
--- a/exercises/practice/word-count/.meta/example.rb
+++ b/exercises/practice/word-count/.meta/example.rb
@@ -1,4 +1,4 @@
-class Phrase
+class WordCount
   def initialize(source)
     @source = source
   end
--- a/exercises/practice/word-count/.meta/generator/word_count_case.rb
+++ b/exercises/practice/word-count/.meta/generator/word_count_case.rb
@@ -3,7 +3,7 @@ require 'generator/exercise_case'
 class WordCountCase < Generator::ExerciseCase
   def workload
     [
-      "phrase = Phrase.new(#{sentence.inspect})",
+      "phrase = WordCount.new(#{sentence.inspect})",
       "counts = #{expected}",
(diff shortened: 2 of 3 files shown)


kotp:
Though you would be correct to associate the patterns, this is indeed
correct.  But, we could bring it up in
[exercism/problem-specifications][1]
where this exercise name and file name are determined.

However, there is nothing so far that states "The exercise title and the
code solution must match directly in name".

And so this is OK, and as designed.

[1]: https://github.com/exercism/problem-specifications

kotp:
Also, there is a reference to `Phrase` when the tests first start, via
the exception that is raised.  Also, congratulations on learning that
the test class does not have to exactly match a class that exists in the
code being tested, as you have noted.

This is a practice exercise, and so learning from the tests as much as
from the exercise itself is a good thing.

This could be mentioned in the Instructions as a learning point, maybe.
I don't know.

With it matching, though, it does make it simpler to make the generators
work.  So it might be worthwhile bring up in the "canonical" repository,
mentioned in my prior message.

AngelGallegosE:
Thanks, @kotp 👍🏼

kotp:
No, thank you @AngelGallegosE .  I hope to see you around again!  Thank you for the work. As an aside, though this was a simple patch, it might be helpful to raise an issue before the work is done to find out if it is something that needs to be addressed. 

That said, it is good practice doing the patch in itself.  So really nothing lost.  And now we also have a reference for if this comes up in another exercise and a change is thought needed.  (We might later decide that it makes sense to make the change, then we could re-open this and apply it, after all!)

PR #1295, opened 2022-02-02
log-line-parser: Fix readme for `capitalize!` method

christinashan:
https://ruby-doc.org/core-2.7.0/String.html#method-i-capitalize-21
Based on the ruby docs, the `capitalize!` method should result in this: 
```
my_string.capitalize!  #=> "Hello"
puts my_string         #=> "Hello"
```

`nil` happens if the string is already capitalized. The full capitalization of the entire string is the `upcase` method.


kotp (approved):
Very nice catch.  This is the problem of not using doctests for documentation.  You  can not easily get pass/failure for your documentation.

PR #1297, opened 2022-02-20
Link to wayback machine for rubylearning.com

kotp:
The link goes to the wayback machine, for now.

Hopefully I can recover the domain record.  But if not, then the changes
will need to be made permanent, perhaps the resource moved to another
domain.

I am trying to communicate with the new registered entity to resolve
this.


kotp:
Ref: exercism/exercism#6250

SleeplessByte (approved):

PR #1304, opened 2022-03-15
Bird count exercise rewording

harry-wood:
Attempt to slightly reword the bird count exercise tasks 1 & 2 to be a little clearer. It's still confusing the way the BirdCount class actually holds a set of 7 bird counts for the "current" week (including yesterday) as well as a set of 7 counts for "last week" (previous to the current week), but hopefully this makes it a little clearer without substantially changing the exercise.

### DIFF
--- a/exercises/concept/bird-count/.docs/instructions.md
+++ b/exercises/concept/bird-count/.docs/instructions.md
@@ -1,12 +1,12 @@
 # Instructions
 
-You're an avid bird watcher that keeps track of how many birds have visited your garden in the last seven days.
+You're an avid bird watcher that keeps track of how many birds have visited your garden. Each bird count lasts seven days.
 
 You have five tasks, all dealing with the numbers of birds that visited your garden.
 
 ## 1. Check what the counts were last week
 
-For comparison purposes, you always keep a copy of last week's counts nearby, which were: 0, 2, 5, 3, 7, 8 and 4. Implement the `BirdCount.last_week` method that returns last week's counts:
+For comparison purposes, you always keep a copy of the previous week's counts nearby, which were: 0, 2, 5, 3, 7, 8 and 4. Implement the `BirdCount.last_week` method that returns previous week's counts:
 
 ```ruby
 BirdCount.last_week
@@ -15,7 +15,7 @@ BirdCount.last_week
 
 ## 2. Check how many birds visited yesterday
 
-Implement the `BirdCount#yesterday` method to return how many birds visited your garden yesterday. The bird counts are ordered by day, with the first element being the count of the oldest day, and the last element being today's count.
+Implement the `BirdCount#yesterday` method to return how many birds visited your garden yesterday. A new bird count for the current week is ordered by day, with the first element being the count of the oldest day, and the last element being today's count.
 
 ```ruby
 birds_per_day = [2, 5, 0, 7, 4, 1]


harry-wood:
I started writing an issue to describe my confusion with this exercise, but it felt easier to explain by proposing a  rewording. Feel free to reword my rewording though!

kotp:
Mentioning @pvcarrera as the concept author.

kotp:
I left it for a while, thinking I would understand the confusing parts just by thinking about what was proposed by the changes, but I am finding that I do not still understand what was confusing.  There are changes made that change the meaning, such as stating "new" when that was not stated, there being no requirement that there is a new count from what I can see there. We could as well use a different count, or the old one.

In other places, "previous" and "last" week means the same thing, so I am still not sure how it clarifies.

If you had made the changes and accompanied that by the written description, I think that would have helped a lot.  (And yes, I know it can be hard.  I struggle with communication every day.)

Can you take some time and try to write down what you are feeling and finding confusing so that I can understand the changes better and how they relate?

harry-wood:
It's clear what we need to do when we go on to look at the sample output, hence people are not failing to complete the exercise because of this, but reading things sequentially I found the description text confusing.

>  stating "new" when that was not stated, there being no requirement that there is a new count

Well the sample output goes on to show creating a new BirdCount `BirdCount.new(birds_per_day)` and on that we want to be able to call the new `yesterday` method. So there is a requirement for a new BirdCount, but I guess what you mean is the existing sentence is correct without mentioning "new" ("ordered by day" still stands regardless of whether it's new or not). True, but I'm jamming a bit of extra info into the sentence, adding some hint about creating a new BirdCount. In fact the phrase _"A new bird count for the current week"_ offers several clarifications:
* Clarify that a BirdCount object holds several... ermm... bird counts (This is confusing. I think ideally you'd rename the class BirdCounts actually)
* Clarify (or hint) that a BirdCount is going need to be constructed to have the data for the `yesterday` method
* Clarify that we're looking at the current week for the `yesterday` method as distinct from the previous/last week.

The previous/last word. I was hoping "previous week" would better convey the intended behaviour (the set of 7 days coming before the current week) while we might say "last week" to mean the last 7 days. But I don't know if that really works as an improvement. Previous/last _kind of_ mean the same thing.

kotp:
I think that if today is the first day of the week, then yesterday is indeed inclusive of the prior or last week.  But I do know that sometimes people use yesterday in a loose way, meaning "a prior day that is not always the previous day in relation to today".

And the exercise in general seems to communicate that we are doing this at the first day of a new week.

So I do not agree that we are looking at the current week for the `yesterday` method.  The data is for last week, we are in a new week doing reporting on the prior week, and one of those days would in fact be "yesterday".

This could be set up more distinctly, I think.

I have asked for someone else to look at this as well.

joshgoebel:
OK, reviewing this one.   I believe I see the confusion here now, but I personally don't find the wording changes proposed here to add any clarity or resolve any of the issues.

We could probably more clearly state that the `BirdHistory` (perhaps a better name) class is intended to track both the current week and prior week, but then it is a fair design critique that we don't simply use a second BirdHistory object for the prior week which would allow then re-using all the statistical methods for the prior week also, so I do see that point.

kotp:
What if the change were this?

```diff
diff --git i/exercises/concept/bird-count/.docs/instructions.md w/exercises/concept/bird-count/.docs/instructions.md
index 3d93270e..9d6a64c6 100644
--- i/exercises/concept/bird-count/.docs/instructions.md
+++ w/exercises/concept/bird-count/.docs/instructions.md
@@ -1,6 +1,6 @@
 # Instructions
 
-You're an avid bird watcher that keeps track of how many birds have visited your garden in the last seven days.
+It's the first day of the week, and you're an avid bird watcher that keeps track of how many birds have visited your garden each day.
 
 You have five tasks, all dealing with the numbers of birds that visited your garden.
 
```

I think the setup defines things in a way that makes the story make sense.

harry-wood:
Well "today" is the last day of the week surely? Last day of the 7 values we're passing into `new`

kotp:
> Well "today" is the last day of the week surely? Last day of the 7 values we're passing into `new`

That, yeah, that complicates it, since if today is the last day of the week, and we only keep a week worth of history, we can not ask for the value last week, but only this week.  And if today is the first day of the next week, we can get all values, except for today, unless we also store today's value for a total of days with counts.

kytrinyx:
@harry-wood @kotp It's been a few months since there's been any movement on this PR.

Since the PR was opened as a way of opening a discussion trying to pinpoint the source of confusion, and the discussion didn't seem to establish any consensus about what was confusing and why, I think my preference here would be to close the PR.

If there's still an appetite for figuring out how to improve the exercise, I think that we should discuss it in an issue (rather than a PR), with the goal of pinpointing the source of confusion before figuring out how to address it.

kotp:
Closing, without prejudice.

Issue #1315, opened 2022-04-21
`item_names` in Boutique Inventory challenge


robsdudeson:
The instructions indicate that `item_names` should only [consider items with stock](https://github.com/exercism/ruby/blob/0ccfe9c9e85eca433e23a9bcdf2227dd714b1596/exercises/concept/boutique-inventory/.docs/instructions.md?plain=1#L18) but the [tests expect names of items without stock to be in the output](https://github.com/exercism/ruby/blob/0ccfe9c9e85eca433e23a9bcdf2227dd714b1596/exercises/concept/boutique-inventory/boutique_inventory_test.rb#L9)

kotp:
Yep, I did mention this in the related #1316.  I think we clarify the README, and maybe add a test that gathers information about stock levels that are low (whatever that level may be at the moment called).  But we should also make sure that we are sticking to the concepts that are given for this concept exercise.

PR #1316, opened 2022-04-21
fix: `BoutiqueInventory#item_name` assertions

robsdudeson:
fixes: #1315

* addresses rubocops in the files touched
* updated the [instructions.md](./exercises/concept/boutique-inventory/.docs/instructions.md) to remove ambiguity about stock levels
* updated tests to reflect that stock levels do not impact `item_name` determination

### DIFF
--- a/exercises/concept/boutique-inventory/.docs/instructions.md
+++ b/exercises/concept/boutique-inventory/.docs/instructions.md
@@ -15,7 +15,7 @@ A single item in the inventory is represented by a hash, and the whole inventory
 ]
 ```
 
-## 1. Return a list of the names of the items.
+## 1. Return a list of the names of the items
 
 Implement `BoutiqueInventory.item_names`, which should return a list of the item names ordered alphabetically.
 
--- a/exercises/concept/boutique-inventory/.docs/instructions.md
+++ b/exercises/concept/boutique-inventory/.docs/instructions.md
@@ -15,7 +15,7 @@ A single item in the inventory is represented by a hash, and the whole inventory
 ]
 ```
 
-## 1. Return a list of the names of the items in the inventory
+## 1. Return a list of the names of the items.
 
 Implement `BoutiqueInventory.item_names`, which should return a list of the item names, ordered alphabetically.
 
--- a/exercises/concept/boutique-inventory/.docs/instructions.md
+++ b/exercises/concept/boutique-inventory/.docs/instructions.md
@@ -15,9 +15,9 @@ A single item in the inventory is represented by a hash, and the whole inventory
 ]
 ```
 
-## 1. Return a list of the names of the items in stock
+## 1. Return a list of the names of the items in the inventory
 
-Implement `BoutiqueInventory.item_names` which should return a list of the item names, ordered alphabetically.
+Implement `BoutiqueInventory.item_names`, which should return a list of the item names, ordered alphabetically.
 
 ```ruby
 BoutiqueInventory.new([
--- a/exercises/concept/boutique-inventory/.docs/instructions.md
+++ b/exercises/concept/boutique-inventory/.docs/instructions.md
@@ -17,7 +17,7 @@ A single item in the inventory is represented by a hash, and the whole inventory
 
 ## 1. Return a list of the names of the items.
 
-Implement `BoutiqueInventory.item_names`, which should return a list of the item names, ordered alphabetically.
+Implement `BoutiqueInventory.item_names`, which should return a list of the item names ordered alphabetically.
 
 ```ruby
 BoutiqueInventory.new([
--- a/exercises/concept/boutique-inventory/.meta/exemplar.rb
+++ b/exercises/concept/boutique-inventory/.meta/exemplar.rb
@@ -13,24 +13,23 @@ class BoutiqueInventory
 
   def out_of_stock
     items.select do |item|
-      item[:quantity_by_size].none? {|size, quantity| quantity > 0 }
+      item[:quantity_by_size].none? { |_size, quantity| quantity.positive? }
     end
   end
 
   def stock_for_item(name)
-    items.find {|i|i[:name] == name}[:quantity_by_size]
+    items.find { |i| i[:name] == name }[:quantity_by_size]
   end
 
   def total_stock
     items.sum do |item|
-      item[:quantity_by_size].sum {|_,quantity| quantity }
+      item[:quantity_by_size].sum { |_, quantity| quantity }
 
       # This would also be acceptable, but isn't explicitly
       # taught in this exercise:
-      #item[:quantity_by_size].values.sum
+      # item[:quantity_by_size].values.sum
     end
   end
-
 
   private
   attr_reader :items
--- a/exercises/concept/boutique-inventory/.meta/exemplar.rb
+++ b/exercises/concept/boutique-inventory/.meta/exemplar.rb
@@ -4,10 +4,7 @@ class BoutiqueInventory
   end
 
   def item_names
-    items.
-      reject { |item| item[:quantity_by_size].keys&.empty? }.
-      map { |item| item[:name] }.
-      sort
+    items.map { |item| item[:name] }.sort
   end
 
   def cheap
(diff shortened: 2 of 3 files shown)


robsdudeson:
`bin/local-status-check` returns :heavy_check_mark: for files impacted

exercises/concept/boutique-inventory/.meta/exemplar.rb
```diff
@@ -4,7 +4,10 @@ def initialize(items)
   end
 
   def item_names
-    items.map { |item| item[:name] }.sort
```
kotp, line 7: Does it make sense to get a listing of items that we do not have on hand?  It is an item name, not a verification that it is in stock, maybe?  I have to read the README to know if that is a consideration, but on the face of this change, we are changing this behavior considerably.

kotp: The README is confusing in this aspect, since it shows "Red Short Skirt" as part of the output,  when it has none in stock, but it does say for point 1 that it should "Return a list of the names of the items in stock."

We should probably remove "in stock" and have it not exclude 0 stock items.  I vote for a change in the README to get rid of the conflicting information.  I prefer to have `item_names` give a listing of the names of items, regardless of stock level.

exercises/concept/boutique-inventory/boutique_inventory_test.rb
```diff
@@ -6,19 +6,26 @@ def test_no_item_names
     assert_empty BoutiqueInventory.new([]).item_names
   end
 
-  def test_one_item_name
+  def test_one_item_name_out_of_stock
```
kotp, line 9: This changes the test.  And is related to the above as well.  Getting an item name should not depend on stock level, as makes sense intuitively since this test name has changed.  It may be just coincidentally that the item name happens to be out of stock.  It should be as relevant to test getting one item name given any level of stock.

exercises/concept/boutique-inventory/boutique_inventory_test.rb
```diff
@@ -6,19 +6,26 @@ def test_no_item_names
...
   end
 
   def test_three_item_names
     items = [
-      { price: 65.00, name: "Red Brown Dress", quantity_by_size: {} },
-      { price: 50.00, name: "Red Short Skirt", quantity_by_size: {} },
-      { price: 29.99, name: "Black Short Skirt", quantity_by_size: {} }
+      { price: 65.00, name: "Red Brown Dress", quantity_by_size: { s: 1 } },
+      { price: 50.00, name: "Red Short Skirt", quantity_by_size: { m: 1 } },
+      { price: 29.99, name: "Black Short Skirt", quantity_by_size: { l: 1 } }
```
kotp, line 28: Let's not give all items a stock level of 1, but at least retain one with no stock.  It should communicate the idea that it is important to return the item name, regardless of stock level.

kotp (changes requested):

exercises/concept/boutique-inventory/.meta/exemplar.rb
kotp: To be sure, though, I have made the statement "We have zero in stock."  So even none is a stock level.  We list it as something that we do stock and having 0 in stock makes sense.

exercises/concept/boutique-inventory/.meta/exemplar.rb
```diff
@@ -13,25 +16,24 @@ def cheap
 
   def out_of_stock
     items.select do |item|
-      item[:quantity_by_size].none? {|size, quantity| quantity > 0 }
+      item[:quantity_by_size].none? { |_size, quantity| quantity.positive? }
```
ErikSchierboom, line 19: I'm fairly certain that students won't know about `.positive?` at this point, and that they'd therefore be writing `quantity > 0`. As the exemplar should only be using the things the students must know about at this point, could you revert the change?

kotp: I am not sure if I agree, as it is easy to look at documentation and notice the names.  This assumes that students read documentation though.  Even if they do not, quick introspection of the methods available on the object they have at hand would expose `postive?`.  We do not have a concept for "reading documentation" nor "introspection" though.

However, if the exemplar is meant to be shown to students, after solving, for comparison, or if both are as exemplary as the other a mentor note might be made to mention what the difference or advantage may be for using the other that is not used.

But reverting this to keep the noise specifically toward what your PR wants to accomplish may be advisable, if you have no strong opinion on this.

ErikSchierboom: Well, I'm not entirely opposed to it being left as it is then, reading your comment.

iHiD: In general Exemplars should only use things that people have been taught about at this stage. I could go either way with the `positive?` method here I think.

robsdudeson: being my first PR, I hadn't considered the standpoint of limiting code to what the user might know.  This is helpful context.  

The intent here was to satisfy `rubocop`.  I can revert if that's something we'd like to ignore.

Would it make sense to also disable the `rubcop` for this bit of code then?  This starts to feel like more advanced than what a user would know at this point. :man_shrugging:

kotp: You can disable Rubocop per line, or section if it is something that is being demonstrated that goes against Rubocop.  Also, my Rubocop settings are set that Hash literals have no space after and before the delimiters, while blocks need the delimiter spacing.  But whatever is set for the project we can go with.

[See my SO answer] for hints how to do this per line or section

[my SO answer]: https://stackoverflow.com/a/26350879/485864

robsdudeson: > [See my SO answer] for hints how to do this per line or section

what does `SO` mean in this context?

exercises/concept/boutique-inventory/.docs/instructions.md
```diff
@@ -15,9 +15,9 @@ A single item in the inventory is represented by a hash, and the whole inventory
 ]
 ```
 
-## 1. Return a list of the names of the items in stock
+## 1. Return a list of the names of the items in the inventory
```
kotp, line 18:
```suggestion
## 1. Return a list of the names of the items.
```

exercises/concept/boutique-inventory/.docs/instructions.md
```diff
@@ -15,9 +15,9 @@ A single item in the inventory is represented by a hash, and the whole inventory
...
 
-Implement `BoutiqueInventory.item_names` which should return a list of the item names, ordered alphabetically.
+Implement `BoutiqueInventory.item_names`, which should return a list of the item names, ordered alphabetically.
```
kotp, line 20:
```suggestion
Implement `BoutiqueInventory.item_names`, which should return a list of the item names ordered alphabetically.
```

kotp (changes requested):
Not positive about the period in #1 level 2 heading.  But removing the "in the inventory" can be removed, since it can be assumed that we have a list of items, regardless of inventory.

exercises/concept/boutique-inventory/.meta/exemplar.rb
kotp: Sorry @robsdudeson fixed the link above.  That should answer the question.  But just in case "Stack Overflow" (I did not define it intending for the link hint to do so for me).

exercises/concept/boutique-inventory/.docs/instructions.md
```diff
@@ -15,9 +15,9 @@ A single item in the inventory is represented by a hash, and the whole inventory
 ]
 ```
 
-## 1. Return a list of the names of the items in stock
+## 1. Return a list of the names of the items.
```
kotp, line 18: I was right about being unsure about the period.  This is the only heading with a period.  I think the period does not fit the style for this document in the heading.

robsdudeson: yeah, the linter in my IDE flagged it and I overlooked it.  good catch!

kotp (approved):

PR #1319, opened 2022-04-29
Fix: Make Scrabble last test instantiate and call instance instead static

JadziaHax:
All other tests methods instantiate Scrabble and then call .score on the instance, but the last test called .score as a static method. This change makes the last test consistent with the rest.

### DIFF
--- a/exercises/practice/scrabble-score/scrabble_score_test.rb
+++ b/exercises/practice/scrabble-score/scrabble_score_test.rb
@@ -43,6 +43,6 @@ class ScrabbleTest < Minitest::Test
 
   def test_convenient_scoring
     skip
-    assert_equal 13, Scrabble.score('alacrity')
+    assert_equal 13, Scrabble.new('alacrity').score
   end
 end


exercises/practice/scrabble-score/scrabble_score_test.rb
```diff
@@ -43,6 +43,6 @@ def test_scores_are_case_insensitive
 
   def test_convenient_scoring
     skip
-    assert_equal 13, Scrabble.score('alacrity')
+    assert_equal 13, Scrabble.new('alacrity').score
```
kotp, line 46: Does the solution change because of this?

kotp:
Thank you @JadziaHax for the work on this.

I have the question posed for you, and am looking forward to hearing from you!

exercises/practice/scrabble-score/scrabble_score_test.rb
JadziaHax: I suppose so. The core logic required to solve doesn't change, and any solutions to the current form would still pass as the class method would now just be ignored. You just wouldn't need to have a separate both a class and instance method. It occurs to me now that might have been intentional to highlight the difference. I didn't think of it that way before.

kotp: Yes, indeed, I think you have found the right path.

PR #1330, opened 2022-06-16
fix class's typo

SichangHe:
It seems like a typo but I'm not sure. Please have a look.

### DIFF
--- a/concepts/modules/introduction.md
+++ b/concepts/modules/introduction.md
@@ -3,7 +3,7 @@
 Some times you don't need the overhead of creating an object with state.
 In these cases, a `module` can be used.
 
-A module is very similar to a class (in fact, `Module` is `Classes` parent in the object hierarchy) - the main difference being that rather than using instance methods, we use class methods.
+A module is very similar to a class (in fact, `Module` is `Class`'s parent in the object hierarchy) - the main difference being that rather than using instance methods, we use class methods.
 Class methods start with `self.` and are directly called on a module. 
 For example:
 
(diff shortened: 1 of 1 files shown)


siaw23 (dismissed):

concepts/modules/introduction.md
```diff
@@ -3,7 +3,7 @@
 Some times you don't need the overhead of creating an object with state.
 In these cases, a `module` can be used.
 
-A module is very similar to a class (in fact, `Module` is `Classes` parent in the object hierarchy) - the main difference being that rather than using instance methods, we use class methods.
+A module is very similar to a class (in fact, `Module` is `Class`'s parent in the object hierarchy) - the main difference being that rather than using instance methods, we use class methods.
```
kotp, line 6:
```suggestion
A module is very similar to a class (in fact, `Module` an ancestor to `Class` in the object hierarchy) - 
```
```suggestion
A module is very similar to a class (in fact, `Module` is an ancestor of `Class` in the object hierarchy) - the main difference being that rather than using instance methods, we use class methods.
```

kotp (changes requested):

concepts/modules/introduction.md
kotp: I think this simplifies the language as well as corrects the spelling.

kotp:
Thank you for the PR @SichangHe you are right, it is incorrect.  I have made a suggestion that hopefully makes the reading easier for non-native speakers as well.

kotp (approved):

kotp:
Thank you @SichangHe for the PR!

SichangHe:
@kotp You're welcome.

PR #1332, opened 2022-06-20
Make the desired class methods private

kotp:
The `private_class_method` directive needs a list of method names (strings
are converted to symbols for that list) in order to make the methods
private.  It does not work the same as the familiar `private` method
used at the instance level.

This change creates meta-class instance methods as a group and applies
the familiar `private` directive to the methods that follow that
directive.


kytrinyx (approved):
I think this is fine. If we decide we would like to take the other approach later (for philosophical reasons or whatever) then we can iterate.

PR #1333, opened 2022-06-22
[Lasagna]: Update instructions.md

Crucibl:
I spent 3 days stuck on this problem not understanding why my solution was incorrect after I solved this challenge in 3 other languages. Very infuriating and frustrating as a new coder to bang your head against the wall thinking you're doing something wrong solving the problem. When all you needed to do was remove the raise command and your efforts to solve the problem were correct. I didn't need to do this in the other languages. My other proposal is to comment out the raise the command initially.

### DIFF
--- a/exercises/concept/lasagna/lasagna.rb
+++ b/exercises/concept/lasagna/lasagna.rb
@@ -1,13 +1,13 @@
 class Lasagna
   def remaining_minutes_in_oven(actual_minutes_in_oven)
-    raise 'Please implement the Lasagna#remaining_minutes_in_oven method'
+    raise 'Please implement the Lasagna#remaining_minutes_in_oven method, remove this line of code when implemented'
   end
 
   def preparation_time_in_minutes(layers)
-    raise 'Please implement the Lasagna#preparation_time_in_minutes method'
+    raise 'Please implement the Lasagna#preparation_time_in_minutes method, remove this line of code when implemented'
   end
 
   def total_time_in_minutes(number_of_layers:, actual_minutes_in_oven:)
-    raise 'Please implement the Lasagna#total_time_in_minutes method'
+    raise 'Please implement the Lasagna#total_time_in_minutes method, remove this line of code when implemented'
   end
 end


kotp:
That can be frustrating indeed.

The code that is delivered for this concept exercise is definitely different that the ones for the practice exercise where we have a comment that is in the file to direct a student as to what should be do.

If we have not covered exceptions, I wonder why we are raising exception in the code. This is a concept exercise, and so it may be inappropriate on several levels to do this.

I would be for removing the raising of the exceptions and let the tests do the job of guiding to a solution, rather than raising and using a concept that is not learned yet in the solution file.

---

The solution might be to just add `and remove this line` to the end of the message that comes up when the errors are raised.

---

The update may be misleading, as it will not avoid argument errors but `RuntimeError` instead.

---

The commit history story does not say why we chose to raise exceptions.

exercises/concept/lasagna/.docs/instructions.md
```diff
@@ -42,3 +42,4 @@ lasagna = Lasagna.new
 lasagna.total_time_in_minutes(number_of_layers: 3, actual_minutes_in_oven: 20)
 # => 26
 ```
+Dont forget to delete or comment out the instructions in the method before you run your tests to avoid argument errors!
```
kotp, line 45: My suggestion would be to remove this line, and put the message in the exception message as it is raised.  That is, if we decide to keep the exception being raised, as this is a concept exercise and there has been no training at this point for exceptions, from what I can see.

kotp (changes requested):

iHiD:
@Crucibl Could you paste the code you ran that didn't work please so I can understand more what's happening. Thanks.

Crucibl:
> That can be frustrating indeed.
> 
> The code that is delivered for this concept exercise is definitely different that the ones for the practice exercise where we have a comment that is in the file to direct a student as to what should be do.
> 
> If we have not covered exceptions, I wonder why we are raising exception in the code. This is a concept exercise, and so it may be inappropriate on several levels to do this.
> 
> I would be for removing the raising of the exceptions and let the tests do the job of guiding to a solution, rather than raising and using a concept that is not learned yet in the solution file.
> 
> The solution might be to just add `and remove this line` to the end of the message that comes up when the errors are raised.
> 
> The update may be misleading, as it will not avoid argument errors but `RuntimeError` instead.
> 
> The commit history story does not say why we chose to raise exceptions.

Ok sounds great I think and yes sure. Or you can include a brief paragraph on "Raise" and what it means/does in Ruby

Crucibl:
```  1) Error:
LasagnaTest#test_total_time_in_minutes_for_multiple_layer:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /Users/u/Exercism/ruby/lasagna/lasagna.rb:22:in `total_time_in_minutes'
    lasagna_test.rb:31:in `test_total_time_in_minutes_for_multiple_layer'

  2) Error:
LasagnaTest#test_remaining_minutes_in_oven:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /Users/u/Exercism/ruby/lasagna/lasagna.rb:7:in `remaining_minutes_in_oven'
    lasagna_test.rb:12:in `test_remaining_minutes_in_oven'

  3) Error:
LasagnaTest#test_preparation_time_in_minutes_with_multiple_layers:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /Users/u/Exercism/ruby/lasagna/lasagna.rb:14:in `preparation_time_in_minutes'
    lasagna_test.rb:20:in `test_preparation_time_in_minutes_with_multiple_layers'

  4) Error:
LasagnaTest#test_preparation_time_in_minutes_with_one_layer:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /Users/u/Exercism/ruby/lasagna/lasagna.rb:14:in `preparation_time_in_minutes'
    lasagna_test.rb:16:in `test_preparation_time_in_minutes_with_one_layer'

  5) Error:
LasagnaTest#test_total_time_in_minutes_for_one_layer:
ArgumentError: wrong number of arguments (given 1, expected 0)
    /Users/u/Exercism/ruby/lasagna/lasagna.rb:22:in `total_time_in_minutes'
    lasagna_test.rb:24:in `test_total_time_in_minutes_for_one_layer' ```
```

Crucibl:
``` class Lasagna

  EXPECTED_MINUTES_IN_OVEN = 40
  PREPARATION_TIME = 2 

  def remaining_minutes_in_oven(actual_minutes_in_oven)
   
    return EXPECTED_MINUTES_IN_OVEN - actual_minutes_in_oven


  end

  def preparation_time_in_minutes(number_of_layers)
  
  
 return number_of_layers * 2
  end


  def total_time_in_minutes(number_of_layers, actual_minutes_in_oven)
  return number_of_layers * 2 + actual_minutes_in_oven
    
  end
end ```
```

Crucibl:
``` class Lasagna

  EXPECTED_MINUTES_IN_OVEN = 40
  PREPARATION_TIME = 2 

  def remaining_minutes_in_oven(actual_minutes_in_oven)
   
    return EXPECTED_MINUTES_IN_OVEN - actual_minutes_in_oven


  end

  def preparation_time_in_minutes(number_of_layers)
  
  
 return number_of_layers * 2
  end


  def total_time_in_minutes(number_of_layers, actual_minutes_in_oven)
  return number_of_layers * 2 + actual_minutes_in_oven
    
  end
end
```

Crucibl:
```
class Lasagna

  EXPECTED_MINUTES_IN_OVEN = 40
  PREPARATION_TIME = 2 

  def remaining_minutes_in_oven(actual_minutes_in_oven)
   
    return EXPECTED_MINUTES_IN_OVEN - actual_minutes_in_oven


  end

  def preparation_time_in_minutes(number_of_layers)
  
  
 return number_of_layers * PREPARATION_TIME
  end


  def total_time_in_minutes(number_of_layers, actual_minutes_in_oven)
  return number_of_layers * PREPARATION_TIME + actual_minutes_in_oven
    
  end
end
```

Crucibl:
Now the only error I get is 
```

  1) Error:
LasagnaTest#test_total_time_in_minutes_for_multiple_layer:
ArgumentError: wrong number of arguments (given 1, expected 2)
    /Users/u/Exercism/ruby/lasagna/lasagna.rb:20:in `total_time_in_minutes'
    lasagna_test.rb:31:in `test_total_time_in_minutes_for_multiple_layer'

  2) Error:
LasagnaTest#test_total_time_in_minutes_for_one_layer:
ArgumentError: wrong number of arguments (given 1, expected 2)
    /Users/u/Exercism/ruby/lasagna/lasagna.rb:20:in `total_time_in_minutes'
    lasagna_test.rb:24:in `test_total_time_in_minutes_for_one_layer'
```

exercises/concept/lasagna/lasagna.rb
```diff
@@ -1,13 +1,13 @@
 class Lasagna
   def remaining_minutes_in_oven(actual_minutes_in_oven)
-    raise 'Please implement the Lasagna#remaining_minutes_in_oven method'
+    raise 'Please implement the Lasagna#remaining_minutes_in_oven method, remove this line of code when implemented'
```
ErikSchierboom, line 3: What about the following:

```suggestion
    raise 'Please implement the Lasagna#remaining_minutes_in_oven method, remove this line of code to start implementing'
```

This makes it clear that you need to remove the statement and suggests to do it when you start implementing the method, which makes the most sense I feel.

That being said, there is indeed the fact that the student won't know what `raise` is when they start this exercise. The only real alternative would be to use a comment instead, which _is_ explained to the student. That might actually be a better solution.

kotp: Doesn't the tests fail if this is not implemented?  I really think that the tests could communicate what needs to be done instead, rather than having this information in the supplied file.  We might give the scaffolding, the method names, the fact that they are not designed to pass should give us the communication the student needs, rather than raising explicitly from here.

ErikSchierboom: Which aligns with the second part of my comment: using comments.

kotp: I was thinking to not even necessarily do the comments.  The tests would be "in your face" from the test file.

But it is another chance to remind the student to remove stale comments, and that is never a bad thing.

kytrinyx:
Was there a consensus on how to move forward here?

kytrinyx:
As we've not heard anything further here, I'm going to go ahead and close this. Thanks, all, for taking the time to discuss it.

PR #1334, opened 2022-06-22
Bumps Kindergarten Garden exercise difficulty level

NeimadTL:
This was consider as an easy one, however this exercise involves
being knowledgeable in metaprogramming concepts as mentioned
by an user. Therefore we decided to bump up its difficulty level
to 4.

### DIFF
--- a/config.json
+++ b/config.json
@@ -934,7 +934,14 @@
         "name": "Kindergarten Garden",
         "uuid": "04bde625-e363-4d8b-880f-db7bf86286eb",
         "practices": ["symbols"],
-        "prerequisites": ["strings", "symbols", "enumerable"],
+        "prerequisites": [
+          "strings",
+          "symbols",
+          "enumerable",
+          "dynamic-programming",
+          "recursion",
+          "meta-programming"
+        ],
         "difficulty": 4,
         "topics": ["parsing", "records", "searching", "strings", "structs"]
       },
--- a/config.json
+++ b/config.json
@@ -935,7 +935,7 @@
         "uuid": "04bde625-e363-4d8b-880f-db7bf86286eb",
         "practices": ["symbols"],
         "prerequisites": ["strings", "symbols", "enumerable"],
-        "difficulty": 3,
+        "difficulty": 4,
         "topics": ["parsing", "records", "searching", "strings", "structs"]
       },
       {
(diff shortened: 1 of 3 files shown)


kotp:
The issue mentions hints that are missing as well, let's make this PR about that to round it out for that issue.

@edgerunner might have some ideas for hints as well.

kotp:
It very well might.  I think my favorite approach uses dynamic programming (not really metaprogramming) with `define_singleton_method`.

NeimadTL:
@kotp, @edgerunner , how do you like those prerequisites :
```
"prerequisites": [
  "strings",
  "symbols",
  "enumerable",
  "dynamic-programming", <- newly added
  "recursion" <- newly added
],
```

and a new instruction.append file with the following in it:

```
In this exercise you're going to be using an advanced concept which is called dynamic programming. https://en.wikipedia.org/wiki/Dynamic_programming
Your solution could use the class method `#define_singleton_method` or the combination of `#method_missing` and `#respond_to_missing`.
``` 
?

kotp:
I would leave out the "an advanced concept" and just say "a concept called …"  But the concepts have to come from a list that is already approved, and modifications to that list are going to be few and far between any additions.

I believe the list is located at https://github.com/exercism/problem-specifications/blob/main/TOPICS.txt 

Someone may correct me on that though.

edgerunner:
Hmm, *meta-programming* is very much a *concept* for Ruby and more (e.g. Elixir). Maybe it should be somewhere in that list. I hope this isn't snowballing out of control 😆 

Is there a language-specific concept list somewhere as well?

kotp:
I asked, and we do not have a canonical list of concepts.  So I think using that is fine.

kotp (dismissed):
There may be more that can be done, but approving to give the benefit of the difficulty increase.

config.json
```diff
@@ -934,8 +934,15 @@
         "name": "Kindergarten Garden",
         "uuid": "04bde625-e363-4d8b-880f-db7bf86286eb",
         "practices": ["symbols"],
-        "prerequisites": ["strings", "symbols", "enumerable"],
-        "difficulty": 3,
+        "prerequisites": [
+          "strings",
+          "symbols",
+          "enumerable",
+          "dynamic-programming",
```
kotp, line 941: Are we sure about this one?  What hints would you write concerning this if you were to write some mentor notes about dynamic programming concerning this exercise?

kotp (approved):
Let's get this in.

NeimadTL:
Great ! Thanks @kotp 🎉

Issue #1308, opened 2022-04-01
stock_for_item method in boutique Inventory challenge


tundeiness:
The **stock_for_item** method in the Boutique Inventory challenge looks like this before implementation:
```
def stock_for_item(name) 
end
```
After several iterations, I found out that there is no need for the `name` argument in the `stock_for_item` method because I could solve the challenge without supplying the `name` argument and it could be solved like below:

```
def stock_for_item(name)
    items[0][:quantity_by_size]
end
```
`items` being an instance variable.

So it got me suspecting that maybe the instructions are not being enforced by the tests or something else that I may not know.

tundeiness:
Also, removing `name` from the stock_for_item methods throws an error making the test not pass.

kotp:
The removal of `name` from the method is enforced by the tests, so that is not surprising.

The instructions state this:

    ## 4. Return a single item's stock
    
    I'mplement `BoutiqueInventory.stock_for_item(name)` to return the stock of an item by its name:
    
    ```ruby
    BoutiqueInventory.new([
      {price: 65.00, name: "Maxi Brown Dress", quantity_by_size: {s: 3, m: 7, l: 8, xl: 4}},
      {price: 50.00, name: "Red Short Skirt", quantity_by_size: {}},
      {price: 29.99, name: "Black Short Skirt", quantity_by_size: {s: 1, xl: 4}},
      {price: 20.00, name: "Bamboo Socks Cats", quantity_by_size: {s: 7, m: 2}}
    ]).stock_for_item("Black Short Skirt")
    
    #=> {s: 1, xl: 4}
    ```

And the example is good, but unfortunately it only tests the first position, so if the student solution is only checking for the first entry, then it will pass.

This suggests, then, that we need a second test, with an item that is selected by name in a position other than the first position.

Are you up for a patch for this?

tundeiness:
yes, I hope the timing will be right?
-------- Original message --------From: Victor Goff ***@***.***> Date: 02/04/2022  09:36  (GMT+01:00) To: exercism/ruby ***@***.***> Cc: Tunde Oretade ***@***.***>, Author ***@***.***> Subject: Re: [exercism/ruby] stock_for_item method in boutique Inventory challenge (Issue #1308) 
The removal of name from the method is enforced by the tests, so that is not surprising.
The instructions state this:
## 4. Return a single item's stock

I'mplement `BoutiqueInventory.stock_for_item(name)` to return the stock of an item by its name:

```ruby
[...]

PR #1335, opened 2022-06-28
Add new test for boutique inventory exercise

NeimadTL:
The second test `test_stock_for_item_for_some_in_stock` is testing
the stock for the "Shoes" item. However, this item is in the first
position in the items list, which will make the test to pass if a
student's solution is checking for the first entry. In order to
make sure that the solution is correct, a new test has been added
so that no matter which position the item is in, the correct stock
is returned.

fixes: #1308


kotp (approved):

NeimadTL:
Thanks @kotp 🚀

PR #1337, opened 2022-07-01
protein translation private class method

kotp:
The class was not private, since `private` only effects the methods at
the `instance` level.

Includes the following changes:

- Detach methods from class definition
- remove redundant _key part of variable name
- No need for `.nil?` check explicitly
- Make class method private


kytrinyx (approved):
This looks fine to me.

PR #1346, opened 2022-09-12
Savings-Account: Add a test for zero balance case

Adiis:
## Issue

If `current_balance` is zero, there is a chance for an infinite loop.

## Proposal

We're preventing the infinite loop when `current_balance` is zero by adding a test scenario that returns `0` for such a case.


kotp (changes requested):
The PR description states that this test prevents an infinite loop, but it is causing one.  The solution should prevent the infinite loop, the test should only verify.

Can you update the solution file?

You can run the tests locally before submitting, by running `rake test:savings-account`.

Adiis:
> The PR description states that this test prevents an infinite loop, but it is causing one. The solution should prevent the infinite loop, the test should only verify.
> 
> Can you update the solution file?
> 
> You can run the tests locally before submitting, by running `rake test:savings-account`.

@kotp, oups, my bad! Thank you very much for pointing that out! 🙏
I added a [`fixup!`](https://github.com/exercism/ruby/pull/1346/commits/623e89b506ef0462a4f77dbe3d9abafb5b886679), can you, please, check it out?
Thank you!

kotp:
What if the current balance is zero but the desired balance is zero?  This should not raise.  The example code disallows this.

exercises/concept/savings-account/savings_account_test.rb
```diff
@@ -102,4 +102,8 @@ def test_years_before_desired_balance_for_large_start_balance
   def test_years_before_desired_balance_for_large_difference_between_start_and_desired_balance
     assert_equal 85, SavingsAccount.years_before_desired_balance(2_345.67, 12_345.678_9)
   end
+
+  def test_years_before_desired_balance_for_zero_balance
+    assert_raises(RuntimeError) { SavingsAccount.years_before_desired_balance(0, -Float::INFINITY) }
```
kotp, line 107: Why negative infinity and not positive infinity?  We could use `1` here as easily, and I think it would not look to be so "special".   We can communicate this better, perhaps by doing something like this:

```ruby
  def test_years_before_desired_balance_for_zero_balance
    arbitrary_goal = rand(10) + 1
    assert_raises(RuntimeError) do
      SavingsAccount.years_before_desired_balance(0, arbitrary_goal)
    end
  end
```

kotp: This arbitrary goal is made so that it is not possible to be 0, because a balance of 0 with a goal of 0 is valid and achievable.

exercises/concept/savings-account/.meta/exemplar.rb
```diff
@@ -22,6 +22,8 @@ def self.annual_balance_update(balance)
   end
 
   def self.years_before_desired_balance(current_balance, desired_balance)
+    raise RuntimeError if current_balance.zero?
```
iHiD, line 25: If this gets added, then this exercise needs to rely on `exceptions` in the config.json. I think I'd also want a specific exception class etc, not just a RuntimeError.

kytrinyx:
@Adiis What is currently the problem with the infinite loop, from a student's point of view? Is it that they might end up with code that implements an infinite loop and not know how to deal with it? Will the test runner on the site fail in a way that is too confusing?

Adiis:
> @Adiis What is currently the problem with the infinite loop, from a student's point of view? Is it that they might end up with code that implements an infinite loop and not know how to deal with it? Will the test runner on the site fail in a way that is too confusing?

@kytrinyx, that was precisely my goal as I run into an infinite loop while working on this exercise and I wanted to create something to prevent that in the first place, but I'm having trouble with how to include it in this exercise. This exercise doesn't need to deal with Exceptions but I need something to prevent the infinite loop.

exercises/concept/savings-account/.meta/exemplar.rb
Adiis: Is it OK if this exercise relies on Exceptions? Sorry for asking this, but this is my first PR on Exercism. 😄

kytrinyx:
@Adiis Okay, that's really helpful to know!

What happened when you had an infinite loop? Did the test runner crash? Or did it fail with an error message? (Can you recreate the problem, by any chance, and take a screenshot of what you see?)

exercises/concept/savings-account/.meta/exemplar.rb
kotp: Exceptions are introduced in the concept exercise Simple Calculator and used in Moviegoer exercise.

It could be OK to use exceptions in this exercise, but we should put that as a prerequisite, if we do, as stated in [the above comment](https://github.com/exercism/ruby/pull/1346#discussion_r975232373).

kytrinyx:
@Adiis as we haven't heard anything further on this, and don't have enough information to decide how to proceed I'm going to go ahead and close this. If you have time to pick it back up, please holler, and we can reopen it!

Adiis:
@kytrinyx,
> @Adiis as we haven't heard anything further on this, and don't have enough information to decide how to proceed I'm going to go ahead and close this. If you have time to pick it back up, please holler, and we can reopen it!

Sure, thank you! 
I'm not sure what's the correct way and how to communicate on this _issue_.
I was stuck when I added a test (from this [commit](https://github.com/exercism/ruby/pull/1346/commits/2de50048d540e1290bc3d377ffbcd7595693951f)) that made `rake test:savings-account` run indefinitely.
I will try to figure out how to add a testing scenario and inform you about my ideas.

Thank you! 🙏

kytrinyx:
Would you be able to recreate the problem and take a screenshot of what the UI showed you?

In order to solve this we don't necessarily need to prevent an infinite loop, we could potentially give a useful error message if one occurs.

PR #1350, opened 2022-09-15
Change negative interest rate tests

### LINKED ISSUE #1347: [SavingAccounts]: Suggestion: change negative interest rate tests to return positive value
Currenty we have four tests of negative interest rates:

```
  def test_minimal_negative_interest_rates
    assert_in_delta(-3.213, SavingsAccount.interest_rate(-0.000_1), 0.000_1)
  end
  
  def test_small_negative_interest_rates
    assert_in_delta(-3.213, SavingsAccount.interest_rate(-0.123), 0.000_1)
  end

  def test_regular_negative_interest_rates
    assert_in_delta(-3.213, SavingsAccount.interest_rate(-300.0), 0.000_1)
  end

  def test_large_negative_interest_rates
    assert_in_delta(-3.213, SavingsAccount.interest_rate(-55_555.444), 0.000_1)
  end
```

They require an interest rate on a negative balance to always be *-3.213*, no matter what its value, as long as it is less than 0

But a negative interest rate can cause bugs and possibly 'hamper' a student's learning, mainly because it's not very common to see banks and financial institutions using negative interest rates.

An interest rate of *3.123* (positive) applied to a negative balance should give the right result and avoid possible misunderstandings.

murichristopher:
PR for the issue:
* #1347 

#### Issue: 
It's not very common to see banks and financial institutions using a negative interest rates.

#### Proposal:
Using a interest rate of 3.123 (positive) applied to a negative balance should give the right result and avoid possible misunderstandings.

Closes #1347


kotp:
You will find that if you use `rake test:savings-account` locally, that this also fails the tests.

This is the patch that can get those up and running:

```diff
diff --git a/exercises/concept/savings-account/.meta/exemplar.rb b/exercises/concept/savings-account/.meta/exemplar.rb
index 8ffc22bb..6b7eedf5 100644
--- a/exercises/concept/savings-account/.meta/exemplar.rb
+++ b/exercises/concept/savings-account/.meta/exemplar.rb
@@ -1,6 +1,6 @@
 module SavingsAccount
 
-  NEGATIVE_RATE = -3.213
+  NEGATIVE_RATE = 3.213
   SMALL_POSITIVE_RATE = 0.5
   MEDIUM_POSITIVE_RATE = 1.621
   LARGE_POSITIVE_RATE = 2.475
@@ -32,6 +32,6 @@ module SavingsAccount
 
   def self.annual_yield(balance)
     multiplier = interest_rate(balance) / 100
-    balance.abs * multiplier
+    balance * multiplier
   end
 end
```

kotp (changes requested):
```diff
diff --git a/exercises/concept/savings-account/.meta/exemplar.rb b/exercises/concept/savings-account/.meta/exemplar.rb
index 8ffc22bb..6b7eedf5 100644
--- a/exercises/concept/savings-account/.meta/exemplar.rb
+++ b/exercises/concept/savings-account/.meta/exemplar.rb
@@ -1,6 +1,6 @@
 module SavingsAccount
 
-  NEGATIVE_RATE = -3.213
+  NEGATIVE_RATE = 3.213
   SMALL_POSITIVE_RATE = 0.5
   MEDIUM_POSITIVE_RATE = 1.621
   LARGE_POSITIVE_RATE = 2.475
@@ -32,6 +32,6 @@ module SavingsAccount
 
   def self.annual_yield(balance)
     multiplier = interest_rate(balance) / 100
-    balance.abs * multiplier
+    balance * multiplier
   end
 end
```

This is the minimum change to make the tests pass.

kotp:
Any updates needed for the README.md file?

muricchristopher:
So I changed the Instructions.md file, where a '-3,213%' still remained. What exactly about the README.md should I update?

kotp:
Go ahead and squash the commits and add the directive in the commit message: `closes: #1347` so that when this is brought in the issue will automatically close as well.

kotp (changes requested):
Looking good.  The tests, the exemplar, and the documentation all seem to match up now.

Squash the commits, add a directive to close the issue, and then I can merge it.

iHiD (dismissed):
Thanks for the PR.

I'm not a huge fan of it though for two reasons:
1. While I understand the idea that banks don't generally have negative interest rates, they do charge you a percentage fee for having an overdraft, which is what this basically represents.
2. I don't really understand how this change helps someone learning Ruby. We've lost:
working with negative numbers and teaching `abs`, which are part of the point of this.

I can see how it could be argued that (1) makes things clearer, but I'd like to not lose the negative number and `abs`. So maybe you can come up with an alternative that incorporates those teaching aims?

kotp:
I think the alternative is more where absolute is applied, such as problems concerning speed and distance.  It feels wrong here, in my opinion.

kotp:
Quick thought about distance that benefits from absolute is probably the darts exercise, since score is dependent on distance from the center.

kotp:
However, this is the argument against this being about `abs`, really.
I find nothing in the README or introduction.md file, and it is not
listed as a goal, though it is mentioned in the hints file (under the
`.docs` directory).  It may be hinted because the concept solution used
it?

The config file says:

    "concepts": ["loops", "modules"],

But nothing about absolute numbers.

The hints file can (should) be updated if we decide to not encourage the `abs`
method use, and I find it is not appropriate if the percentage on the
balance when that balance is less than 0, is 3.213 percent.  There is no
need for it, and I think it is atypical.  In this case, I might argue it is contrived
only to force a solution that uses the `abs` method.

iHiD:
OK, I'll go with it. Maybe you could reword it to be clear that it should result in negative interest. See [F# for example](https://github.com/exercism/fsharp/blob/main/exercises/concept/interest-is-interesting/.docs/instructions.md)

> 3.213% for a negative balance (results in negative interest).

exercises/concept/savings-account/.docs/instructions.md
```diff
@@ -2,10 +2,10 @@
 
 In this exercise you will be working with savings accounts. Each year, the balance of a savings account is updated based on the interest rate. The interest rate the bank gives depends on the amount of money in the accounts (its balance):
 
-- -3.213% for a negative balance.
 - 0.5% for a non-negative balance less than `1000` dollars.
 - 1.621% for a positive balance greater or equal than `1000` dollars and less than `5000` dollars.
 - 2.475% for a positive balance greater or equal than `5000` dollars.
+- 3.213% for a negative balance.
```
kytrinyx, line 8:
```suggestion
- 3.213% for a negative balance (results in negative interest).
```

kytrinyx:
I made the suggested change in the latest comment from @iHiD.

@kotp would you review this one final time and merge it if you think it's good to go? And if it is not good to go, would you help clarify what needs to happen in order to get it merged?

kotp (approved):
I think that was the last change that would make this good to go.

PR #1352, opened 2022-09-21
fix test 3 in Boutique Inventory

kingshalaby1:
the expected array should follow the same order as the hash


kingshalaby1:
I was wrong. it should be sorted alphabetically

PR #1354, opened 2022-09-29
collatz-conjecture: test invalid positive values

gabo-cs:
From a mentoring discussion with @kotp :

> The tests that exist are not enforcing that we use only valid numbers, it only touches on zero and negative numbers, but it is not inclusive of the numbers that are not valid.

### DIFF
--- a/exercises/practice/collatz-conjecture/collatz_conjecture_test.rb
+++ b/exercises/practice/collatz-conjecture/collatz_conjecture_test.rb
@@ -36,7 +36,7 @@ class CollatzConjectureTest < Minitest::Test
       CollatzConjecture.steps(-15)
     end
   end
-  
+
   def test_positive_value_but_less_than_one_is_an_error
     skip
     assert_raises(ArgumentError) do
--- a/exercises/practice/collatz-conjecture/collatz_conjecture_test.rb
+++ b/exercises/practice/collatz-conjecture/collatz_conjecture_test.rb
@@ -36,4 +36,11 @@ class CollatzConjectureTest < Minitest::Test
       CollatzConjecture.steps(-15)
     end
   end
+  
+  def test_positive_value_but_less_than_one_is_an_error
+    skip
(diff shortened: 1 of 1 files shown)


gabo-cs:
Sorry about the extra commit, I guess we could _Squash and Merge_ should this be approved.

kotp:
Check https://github.com/exercism/problem-specifications and see if this should be added there, since what is here often comes from the community exercises.

gabo-cs:
> Check https://github.com/exercism/problem-specifications and see if this should be added there, since what is here often comes from the community exercises.

I don't see a tests section there, where should I look exactly?

kotp:
I think the first thing to do on Problem Specifications is to open an issue (if the exercise exists there) and ask if this is generally something that the tracks want.  So that should happen before a pull request would happen there, to find out if it is valuable for more than just Ruby.

gabo-cs:
> I think the first thing to do on Problem Specifications is to open an issue (if the exercise exists there) and ask if this is generally something that the tracks want. So that should happen before a pull request would happen there, to find out if it is valuable for more than just Ruby.

Got it! Opened an [issue](https://github.com/exercism/problem-specifications/issues/2103)! Should I close this PR?

kotp:
I converted it to draft in case it is adopted in general.

Once that is determined, then whatever needs to happen here can happen.

gabo-cs:
> I converted it to draft in case it is adopted in general.
> 
> Once that is determined, then whatever needs to happen here can happen.

Thanks!

kytrinyx:
Thanks for raising the question, @gabo-cs.

The collatz conjecture is about integers, so I don't think that a proposal to add a test case for a float will be accepted in problem-specifications.

As such, my preference would be to close this PR.

kytrinyx:
Since I've not heard anything further in the past three weeks, I'm going to close this, per the comment above.

Thanks for taking the time to raise your concern.

PR #1355, opened 2022-09-30
Add approach introduction for two-fer

ErikSchierboom:


exercises/practice/two-fer/.approaches/introduction.md
```diff
@@ -0,0 +1,33 @@
...
+
+```ruby
+module TwoFer
+  module_function
+​
+  def two_fer(name = 'you')
+    format('One for %<name>s, one for me.', name: name)
+  end
+end
+```
```
kotp, line 33: Since it is mentioned that "You can use", we might add a singleton approach:

    ## Singleton
    
    You can use a constant, instead of `class` or `module` and define the
    method on this singleton object.
    
    ```ruby
    TwoFer = 'Sharing is fun.'
    
    def TwoFer.two_fer(name = 'you')
      'One for %<name>s, one for me.' % {name: name}
    end
    ```

kotp (commented):
Because it was mentioned that we _can_, I thought that showing that the tests really do not drive more than is necessary for the student to explore the language to solve the problem successfully.

exercises/practice/two-fer/.approaches/introduction.md
kytrinyx: While there are a whole lot of things we _could_ do, I think that it's worth keeping the default "approaches" document focused on the more common solutions.

For more playful approaches I think we need to have more commentary around this, as it would otherwise be quite misleading as to what we would expect to see out in the wild.

kotp: Then should we remove/change the word `can` and say instead that _the idiomatic approaches use things like_ because words like "can", "should", "must", "may" are powerful words.

How do we avoid discouraging the playful, where it is often where deeper understanding stems from?

kotp: Or perhaps we that while we can do the listed things, it is not a complete list?  That is probably the thing that is missing here.

kytrinyx: These are good questions, @kotp. I believe the intention here is that we can do the listed things but that it's not a complete list.

@ErikSchierboom Should we tweak the wording to make this a bit clearer?

exercises/practice/two-fer/.approaches/introduction.md
```diff
@@ -0,0 +1,33 @@
...
+module TwoFer
+  def self.two_fer(name = 'you')
+    "One for #{name}, one for me."
+  end
+end
+```
+
+## Method definition
+
+You can use `class` instead of `module` and use metaclasses instead of `self`, `module_function`, or extend `self`.
```
ErikSchierboom, line 19:
```suggestion
You may use `class` instead of `module` and use metaclasses instead of `self`, `module_function`, or extend `self`.
```

exercises/practice/two-fer/.approaches/introduction.md
```diff
@@ -0,0 +1,33 @@
...
+
+## String method
+
+Solutions can also use string templates with `Kernel#sprintf`, `Kernel#format` or `String#%`
```
ErikSchierboom, line 23:
```suggestion
Solutions may also use string templates with `Kernel#sprintf`, `Kernel#format` or `String#%`
```

exercises/practice/two-fer/.approaches/introduction.md
ErikSchierboom: I'd be fine with tweaking the wording a bit.
I've added some wording suggestions.

exercises/practice/two-fer/.approaches/introduction.md
kytrinyx: I honestly don't think that changing from `can` to `may` adds anything here. Seeing as this is under the subheading "## String method" it sounds like these things are all referring to different ways of doing string formatting, rather than different ways of solving the exercise. Am I mis-reading this?

exercises/practice/two-fer/.approaches/introduction.md
kytrinyx: I think the bit about this being an idiomatic approach solves it for me.

exercises/practice/two-fer/.approaches/introduction.md
kotp: For a practice exercise, perhaps.  This is not a concept exercise, so I am unsure we have to be as careful about "prior knowledge" since we may assume they are coming from other languages.

exercises/practice/two-fer/.approaches/introduction.md
kotp: I think so too.

exercises/practice/two-fer/.approaches/introduction.md
kotp: I think it takes away the implied ability connotation, and gives permission and possibility.  The word "may" gives both permission and possibility, while "can" implies ability (and until they do so, we really do not know if they can or can not).

I am positive that "may" is better for this, because we are suggesting and encouraging the possibility to use these, rather than implying only the ability to do so.

exercises/practice/two-fer/.approaches/introduction.md
kytrinyx: We're assuming that people are coming from other languages when they do concept exercises, as well.

I'm not worried about mentioning things that haven't been introduced, I think that's a given in the approaches documents. What I'm wondering is whether we should be a bit more verbose or explanatory, since this is one of the very first practice exercises, and people might be coming from languages that don't have these concepts.

kotp: That's definitely fair.  I think in their research of what a module is, they should have encountered the extend/include as that is pretty core to what modules are used for.  But linking to a resource about module and class and what the differences are would probably be welcome, especially early on.

Do we expect people to go through the concept exercises before they go to practice mode, if they are really new to the language?  So in the case of "new to language" rather than "practicing" would we expect this information be covered already?

kytrinyx: This would be unlocked to practice with almost as soon as you get started with the concept exercises, so I think it could go either way.

My gut feeling here is that it's _good enough_ for now, and that we can take a second pass once the whole approaches thing is in production. We'll probably find out soon enough if people get confused.

kytrinyx (approved):
I think this is good enough to get into production so we can do testing.

ErikSchierboom:
As Katrina mentioned, this document is mostly for having _something_ there for testing. I'm happy to take follow-up PRs to further improve things. Thanks all!

PR #1356, opened 2022-10-03
Tweak RuboCop configuration

kytrinyx:
Our test suites are not optimized for being the best possible style for Ruby, but rather to give students a good experience while doing exercises, as well as reducing the maintenance burden for maintainers.

As such, I think it's worth being fairly lenient in terms of which rubocop cops are enabled for the exercise tests.

In this PR I disable the `ClassLength`, `MethodLength`, and `AbcSize` metrics cops for the exercise tests across the board.
I also disable the `VariableNumber` naming cop for the exercise tests.

Finally, it removes unnecessary disable directives in the tests for all directives that are now covered in the top-level config.


kytrinyx:
I'm also tempted to change `Metrics/LineLength` to exclude exercise test suites, because if we generate a test and the line is too long I'm going to add an exclusion for it.

kotp (approved):
Approving with the nod that there will likely be another change coming. (re: Metrics/LineLength)

PR #1361, opened 2022-10-04
Sync all-your-base with problem-specifications

kytrinyx:


kotp (changes requested):
Perhaps use code block instead?  Or 4 space indent (no language highlighting then) to avoid reading this in plain text with the escaped asterisks.

kotp (approved):

kotp:
I think this gets fixed with recent changes in Problem Specifications.

kotp:
Will this need a "re-sync" @kytrinyx ?

kytrinyx:
Indeed! I've re-synced, and force pushed.

PR #1362, opened 2022-10-04
Sync atbash-cipher with problem-specifications

kytrinyx:


kotp (approved):
Approving but not merging in case you want to do the Single Sentence per Line thing

kytrinyx:
I can't change that here if we want to be synced to the problem-specifications repository. I am going to look at how widespread the problem is there.

PR #1363, opened 2022-10-04
Sync crypto-square with problem-specifications

kytrinyx:

### DIFF
--- a/exercises/practice/crypto-square/.docs/instructions.md
+++ b/exercises/practice/crypto-square/.docs/instructions.md
@@ -5,9 +5,9 @@ Implement the classic method for composing secret messages called a square code.
 Given an English text, output the encoded version of that text.
 
 First, the input is normalized: the spaces and punctuation are removed
-from the English text and the message is downcased.
+from the English text and the message is down-cased.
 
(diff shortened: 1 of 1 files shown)


exercises/practice/crypto-square/.docs/instructions.md
```diff
@@ -5,9 +5,9 @@ Implement the classic method for composing secret messages called a square code.
 Given an English text, output the encoded version of that text.
 
 First, the input is normalized: the spaces and punctuation are removed
-from the English text and the message is downcased.
+from the English text and the message is down-cased.
 
-Then, the normalized characters are broken into rows.  These rows can be
+Then, the normalized characters are broken into rows. These rows can be
```
kotp, line 10: Single sentence per line, but approving.

kotp (approved):

exercises/practice/crypto-square/.docs/instructions.md
kytrinyx: Good catch. I will make a note that this needs to be fixed in the problem-specifications repository.

PR #1396, opened 2022-10-05
Simple Calculator: replace `**` operator with a truly bogus operator

JaKXz:
In [my solution](https://exercism.org/mentoring/external_requests/29bd4863cb604480b2b73a10ada25bfe) wanted to make simple calculator a bit more generic, but this test required me to hardcode an unnecessary condition.

### DIFF
--- a/exercises/concept/simple-calculator/simple_calculator_test.rb
+++ b/exercises/concept/simple-calculator/simple_calculator_test.rb
@@ -27,7 +27,7 @@ class SimpleCalculatorTest < Minitest::Test
   end
 
   def test_raises_exception_for_non_valid_operations
-    assert_raises(SimpleCalculator::UnsupportedOperation) { SimpleCalculator.calculate(1, 2, '**') }
+    assert_raises(SimpleCalculator::UnsupportedOperation) { SimpleCalculator.calculate(1, 2, '*******') }
   end
 
   def test_raises_exception_when_operation_is_nil


kotp:
So, why did you change the test?  Also, before raising an issue, it is probably good to have the exercise mentored first, and discuss the things that you "had to do" before creating an issue here.  They can help determine if it is a problem with understanding Ruby, the exercise, or the tests, and then we can open an issue to update the exercise if necessary.

At this point, because the exercise passes the tests (unmodified), I am unsure what the issue is.

JaKXz:
Hey @kotp thanks for the reply. I'll try to answer your questions here.

> So, why did you change the test?

This is just a suggestion to allow for more flexibility in the solution. This test in particular enforces that you cannot use the `**` operator, but it is a valid operator for `Numeric`s. So, if you're able to see my solution, you'll see I had to add this line as a "workaround" so that my more generic solution would pass all tests:

```rb
    raise UnsupportedOperation if operation == '**'
```

This line does not actually affect the "correctness" of the solution, but, is just to pass this one test that [perhaps intentionally] limits the functionality of the simple calculator.

I'm happy to hear your suggestions here or on the exercism platform on why you think this is the wrong approach. Absolutely no worries if you feel this is the wrong change and want to close this PR.

exercises/concept/simple-calculator/simple_calculator_test.rb
```diff
@@ -27,7 +27,7 @@ def test_no_number_second_operand_raises_exception
   end
 
   def test_raises_exception_for_non_valid_operations
-    assert_raises(SimpleCalculator::UnsupportedOperation) { SimpleCalculator.calculate(1, 2, '**') }
+    assert_raises(SimpleCalculator::UnsupportedOperation) { SimpleCalculator.calculate(1, 2, '*******') }
```
kotp, line 30: This test should not be changed, and it is the test that says "Our library does not support this operation." 

However, instead of changed, I could see an argument for removing it, since yes, we could expand supported operations.

For the goals of this exercise, since we are not specifically aiming to support "ruby like" operators, and only generic, this test still allows for the exploration of allowing something like `'^'` to be added to the supported list of operators, while disallowing this specifically for the library (but not for Ruby itself, a different subject, really.)

kotp:
I am closing this, until such time that we determine that this exercise should be changed.

exercises/concept/simple-calculator/simple_calculator_test.rb
JaKXz: > This test should not be changed, and it is the test that says "Our library does not support this operation."

I don't follow this reasoning... the test title says that `the program should raise an exception for invalid operations` - not quite "our library does not support _this_ operation". This change more closely reflects the behaviour implied by the test's title, IMO. It'd be clearer to change the title of the test if the quoted statement is the intention, and perhaps add more assertions of non-standard/generic operations.

> For the goals of this exercise, since we are not specifically aiming to support "ruby like" operators, and only generic,

This is fair enough, so I won't say any more. Thanks for feedback and conversation.

kotp: Yes, but the README and test are talking about our library, not Ruby specifically (We aren't testing Ruby, we are writing a program using Ruby).

For the change of the title of the test, that should be discussed in a way that makes sense for the tracks using the exercise, I think.  That would be in https://github.com/exercism/problem-specifications where the discussion and change would happen, hopefully, to benefit the other tracks as well.

And yes, of course, you are welcome to say more.  (And especially in the track exercise, during mentoring.)

PR #1398, opened 2022-10-09
Sync all practice exercise docs

kytrinyx:
This updates all the descriptions for the practice exercises based on the changes in the problem specification.

Most (but not all) of the changes are implementing the one-sentence-per-line formatting, as well as moving inline links to be reference links.

If there are specific exercises that you think we should not update until we also update the tests, let me know and I'll remove them from this PR.

### DIFF
--- a/exercises/practice/accumulate/.docs/instructions.md
+++ b/exercises/practice/accumulate/.docs/instructions.md
@@ -1,9 +1,6 @@
 # Instructions
 
-Implement the `accumulate` operation, which, given a collection and an
-operation to perform on each element of the collection, returns a new
-collection containing the result of applying that operation to each element of
-the input collection.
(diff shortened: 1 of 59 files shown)


kotp (approved):
Not sure about the question about "merge before test updates", though.

kytrinyx:
How about if I remove docs changes for the exercises where there are test changes to make?

kytrinyx:
Having thought about this a bit, I think that I prefer to update only the docs for exercises that don't have any pending test updates.

I'm going to remove changes for those exercises from this PR.

kytrinyx:
That removed the changes to 70 exercises, leaving instructions changed for 118 exercises.

kotp:
Good choice!  Otherwise, the "update exercise" irritant will happen twice, and once only for formatting, rather than test updates and formatting and content.

PR #1400, opened 2022-10-12
Sync space-age with problem-specs

kytrinyx:
There was a new test upstream for an error case.

We have not previously introduced error handling for this exercise, and I do not think that adding error handling here makes the exercise more interesting.

I've therefore chosen to exclude the new test on this track.


ErikSchierboom (approved):
> We have not previously introduced error handling for this exercise, and I do not think that adding error handling here makes the exercise more interesting.
> 
> I've therefore chosen to exclude the new test on this track.

I did the exact same thing on the C# track.

PR #1402, opened 2022-10-12
Sync triangle with problem specifications

kytrinyx:
This has two new test cases that both make sense for our implementation.

### DIFF
--- a/exercises/practice/triangle/triangle_test.rb
+++ b/exercises/practice/triangle/triangle_test.rb
@@ -2,25 +2,25 @@ require 'minitest/autorun'
 require_relative 'triangle'
 
 class TriangleTest < Minitest::Test
-  def test_triangle_is_equilateral_if_all_sides_are_equal
+  def test_equilateral_triangle_all_sides_are_equal
     # skip
     triangle = Triangle.new([2, 2, 2])
     assert triangle.equilateral?, "Expected 'true', triangle [2, 2, 2] is equilateral."
   end
 
-  def test_triangle_is_not_equilateral_if_any_side_is_unequal
+  def test_equilateral_triangle_any_side_is_unequal
     skip
     triangle = Triangle.new([2, 3, 2])
     refute triangle.equilateral?, "Expected 'false', triangle [2, 3, 2] is not equilateral."
   end
 
-  def test_triangle_is_not_equilateral_if_no_sides_are_equal
+  def test_equilateral_triangle_no_sides_are_equal
     skip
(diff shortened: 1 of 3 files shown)


exercises/practice/triangle/triangle_test.rb
```diff
@@ -86,28 +86,40 @@ def test_isosceles_triangle_sides_may_be_floats
...
+    skip
+    triangle = Triangle.new([4, 3, 3])
+    refute triangle.scalene?, "Expected 'false', triangle [4, 3, 3] is scalene."
+  end
+
+  def test_scalene_triangle_may_not_violate_triangle_inequality
     skip
     triangle = Triangle.new([7, 3, 2])
-    refute triangle.scalene?, "Expected 'false', triangle [7, 3, 2] is not scalene."
+    refute triangle.scalene?, "Expected 'false', triangle [7, 3, 2] is scalene."
```
kotp, line 122: If we are refuting that a triangle is scalene, with then should it not be scalene?

exercises/practice/triangle/triangle_test.rb
```diff
@@ -86,28 +86,40 @@ def test_isosceles_triangle_sides_may_be_floats
...
-    refute triangle.scalene?, "Expected 'false', triangle [4, 4, 4] is not scalene."
+    refute triangle.scalene?, "Expected 'false', triangle [4, 4, 4] is scalene."
   end
 
-  def test_triangle_is_not_scalene_if_two_sides_are_equal
+  def test_scalene_triangle_first_and_second_sides_are_equal
     skip
     triangle = Triangle.new([4, 4, 3])
-    refute triangle.scalene?, "Expected 'false', triangle [4, 4, 3] is not scalene."
+    refute triangle.scalene?, "Expected 'false', triangle [4, 4, 3] is scalene."
```
kotp, line 104: If we are refuting that triangle.scalene?, then isn't it false, because 4, 4, 3 is not scalene?  Otherwise, if it is scalene, t hen scalene? would not be refuted, but it would be true.

kotp (changes requested):

exercises/practice/triangle/triangle_test.rb
kytrinyx: My bad! Thanks for catching this.

kotp (approved):
Hopefully the scalene/not scalene errors were not coming from upstream.  If so, we should fix them there too.

kytrinyx:
No, it was all me. I am working on a generator and I got the template wrong.

PR #1404, opened 2022-10-12
Sync anagram with problem-specifications

kytrinyx:
This sync applied some reimplemented exercises that changed the data being used.

Two notes on the choices I made:

1. I generated this with double quotes rather than single quotes. We do not have the Rubocop StringLiterals enabled, so I could not autocorrect to use single quotes. I personally don't think it makes a difference, but thought I should call it out.
2. I explicitly did not call the 'sort' method on the values returned. Our reference implementation still passes, but it does mean that we're changing the design of the exercise to no longer require that people return the values in sorted order, but rather to return them in the order that the candidates were passed in. I think this makes more sense, but it could break some people's earlier solutions.

### DIFF
--- a/exercises/practice/anagram/.docs/instructions.md
+++ b/exercises/practice/anagram/.docs/instructions.md
@@ -1,8 +1,13 @@
 # Instructions
 
-An anagram is a rearrangement of letters to form a new word.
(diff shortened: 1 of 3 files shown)


kotp:
Either choice for sorting is a choise made, so I find it is OK.

PR #1405, opened 2022-10-12
Sync dominoes with problem-specifications

kytrinyx:
This added one new test and changed the name of a couple of others.

### DIFF
--- a/exercises/practice/dominoes/dominoes_test.rb
+++ b/exercises/practice/dominoes/dominoes_test.rb
@@ -14,7 +14,7 @@ class DominoesTest < Minitest::Test
     assert Dominoes.chain?(dominoes)
   end
 
-  def test_singleton_that_can_not_be_chained
+  def test_singleton_that_cant_be_chained
     skip
     dominoes = [[1, 2]]
     refute Dominoes.chain?(dominoes)
@@ -32,7 +32,7 @@ class DominoesTest < Minitest::Test
     assert Dominoes.chain?(dominoes)
   end
 
-  def test_can_not_be_chained
+  def test_cant_be_chained
     skip
     dominoes = [[1, 2], [4, 1], [2, 3]]
     refute Dominoes.chain?(dominoes)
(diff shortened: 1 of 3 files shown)


exercises/practice/dominoes/dominoes_test.rb
```diff
@@ -32,7 +32,7 @@ def test_can_reverse_dominoes
     assert Dominoes.chain?(dominoes)
   end
 
-  def test_can_not_be_chained
+  def test_cant_be_chained
```
kotp, line 35: This may be coming from somewhere, but _cant_ is a thrust or motion that tilts something, or to talk in a sing-song way, among other things.

This is obviously supposed to be a contraction, based on the change though, and since we can not represent the contraction, the easiest fix is to simply not use a contraction and instead use the two words.

kytrinyx: > This may be coming from somewhere

This is coming from the `tests.toml` that is on main right now, which in turn came from problem-specifications

<details>
<summary>Permalinks</summary>

https://github.com/exercism/ruby/blob/d491cd5a1b73f00760e59de4bfd899715ebc8bd2/exercises/practice/dominoes/.meta/tests.toml#L19
https://github.com/exercism/ruby/blob/d491cd5a1b73f00760e59de4bfd899715ebc8bd2/exercises/practice/dominoes/.meta/tests.toml#L28
https://github.com/exercism/problem-specifications/blob/b5bfcdad6f53f62ee0f92c0c5a536db7b215aa80/exercises/dominoes/canonical-data.json#L59
https://github.com/exercism/problem-specifications/blob/b5bfcdad6f53f62ee0f92c0c5a536db7b215aa80/exercises/dominoes/canonical-data.json#L86

</details>


> This is obviously supposed to be a contraction

Yes, this is a contraction. In this context I believe that it is clear that this is the contraction, can't, rather than a thrust or motion that tilts something, or talking in a sing-song way. Words such as "can’t" are common enough that I have no problem with them being in Exercism’s codebase, and actively prefer them in docs as I find them friendlier and less formal to read. 

I deliberately use contractions like _can't_ over cannot or can not in a number of contexts, just as I have learned to use emojis to help soften communication in a number of channels. I, personally, find emojis confusing and bewildering, but I know that the vast majority of people I communicate with find that they add more emotion and humanity. So this is a stylistic choice that I have made my peace with.

kotp: Good to know it was purposeful, rather than accidental.  And of course, when in Rome…

kotp (approved):

PR #1421, opened 2022-10-13
Sync resistor-color-duo with problem-specifications

kytrinyx:
We got new docs and two new tests.

### DIFF
--- a/exercises/practice/resistor-color-duo/.docs/instructions.md
+++ b/exercises/practice/resistor-color-duo/.docs/instructions.md
@@ -17,18 +17,17 @@ The program will take color names as input and output a two digit number, even i
 
 The band colors are encoded as follows:
 
-- Black: 0
-- Brown: 1
-- Red: 2
-- Orange: 3
-- Yellow: 4
-- Green: 5
-- Blue: 6
-- Violet: 7
-- Grey: 8
-- White: 9
+* Black: 0
+* Brown: 1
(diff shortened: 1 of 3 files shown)


exercises/practice/resistor-color-duo/.docs/instructions.md
```diff
@@ -17,18 +17,17 @@ The program will take color names as input and output a two digit number, even i
...
+* Black: 0
+* Brown: 1
+* Red: 2
+* Orange: 3
+* Yellow: 4
+* Green: 5
+* Blue: 6
+* Violet: 7
+* Grey: 8
+* White: 9
```
iHiD, line 29: I wonder why these changed from `-` to `*`? I much prefer `-` but I guess they've changed upstream. (This isn't me blocking anything to be clear - I'm just confused why it changed).

iHiD (approved):

exercises/practice/resistor-color-duo/.docs/instructions.md
kotp: @wolf99 linted this on October 14th, 2021.  It seems to have been a personal choice, their linter is set to do this, apparently.

There is no apparent standard for bullets in the description files I just looked at there.

Bring it in for the additional changes, and deal with the noise if/when they are normalized upstream?

Bringing it in, not blocking on this, as well.

PR #1427, opened 2022-10-14
Sync gigasecond with problem-specifications

kytrinyx:
I chose to not bring in the new test for immutable datetimes, as the DateTime and Time classes in Ruby create new copies of themselves when making changes to them, as far as I can tell.

### DIFF
--- a/exercises/practice/gigasecond/.docs/instructions.md
+++ b/exercises/practice/gigasecond/.docs/instructions.md
@@ -1,6 +1,5 @@
 # Instructions
 
-Given a moment, determine the moment that would be after a gigasecond
-has passed.
+Given a moment, determine the moment that would be after a gigasecond has passed.
 
 A gigasecond is 10^9 (1,000,000,000) seconds.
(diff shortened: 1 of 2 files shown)


kotp (approved):

kotp:
Good choice, but for me, very different reasons.

kytrinyx:
I'd love to hear your reasons.

kotp:
The base unit of `Time` is seconds, represented as an integer.  Presumably, for an imaginary `Degree` object, the base unit of `seconds` is also represented as an integer.  The problem is giga-seconds, the base unit being `seconds` and as we generally think of them is as an integer. So, the solution is to take a moment (perhaps even an argument of "seconds" for `from`, as a number of seconds (er, an integer).

The most straight forward solution being a gigasecond plus those seconds, which is reduced to math.

And integers are immutable.

Reducing it to math, means that we may be able to use any object that has a base unit of seconds, and that implements the `+` method, and therefore, we can find out what a gigasecond more of seconds would be for that object.

Also, it reveals that perhaps we used `Time` in the tests for the convenience of the test itself, avoiding outright giving away "gigasecond + seconds" as a solution.

kytrinyx:
That's a much smarter reason than mine! :-D

PR #1433, opened 2022-10-15
Sync high-scores with problem-specifications

kytrinyx:
There were a slew of new tests for methods that were not previously a part of the API of the class.

None of the methods are mentioned in the instructions for the exercise, and they also don't seem very interesting, so I didn't bring them in.

When regenerating the test suite the test for the #latest_is_personal_best? method got removed, so I removed the corresponding code from the reference solution as well.

### DIFF
--- a/exercises/practice/high-scores/.docs/instructions.md
+++ b/exercises/practice/high-scores/.docs/instructions.md
@@ -2,4 +2,5 @@
 
 Manage a game player's High Score list.
 
-Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write methods that return the highest score from the list, the last added score and the three highest scores.
+Your task is to build a high-score component of the classic Frogger game, one of the highest selling and most addictive games of all time, and a classic of the arcade era.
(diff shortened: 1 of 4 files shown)


exercises/practice/high-scores/.meta/example.rb
```diff
@@ -18,8 +18,4 @@ def latest
   def personal_top_three
     scores.max(3)
   end
-
-  def latest_is_personal_best?
-    latest == personal_best
-  end
```
kotp, line 24: That is unfortunate, as it provides a space for me to talk about attributes macros in Ruby, and how to maintain the ability to have a method generated in C, while still having a question mark at the end of the name of the method…

Not the goal of the exercise, of course, but a "So you want to dig deeper into how the language works, because the problem is a simple one?" kind of side stretch.

kotp (approved):
Approving, but right in the middle of mentoring someone that is taking that side stretch at the moment, and don't want to have the tests change and the "problem" disappear out from under us.

Delaying patching this in until after that student is done.

kytrinyx:
I reverted the change to the sample solution. I've created a custom inclusion for this test suite so that the tests for `latest_is_personal_best?` get added to the bottom.

PR #1437, opened 2022-10-16
Sync tournament with problem-specifications

kytrinyx:
The sync brought in changes to the docs and tests.toml

The example solution failed the new test due to a small formatting difference in the spec. The new test is the first one that shows a score that has double-digits.

I updated the example solution to pass the new test, but did not spend any time thinking about how to make the code nice.

### DIFF
--- a/exercises/practice/tournament/.docs/instructions.md
+++ b/exercises/practice/tournament/.docs/instructions.md
@@ -24,7 +24,7 @@ A win earns a team 3 points.
 A draw earns 1.
 A loss earns 0.
 
-The outcome should be ordered by points, descending.
+The outcome is ordered by points, descending.
 In case of a tie, teams are ordered alphabetically.
 
 ## Input
--- a/exercises/practice/tournament/.docs/instructions.md
+++ b/exercises/practice/tournament/.docs/instructions.md
@@ -2,8 +2,7 @@
 
 Tally the results of a small football competition.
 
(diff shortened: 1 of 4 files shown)


exercises/practice/tournament/.docs/instructions.md
```diff
@@ -21,9 +20,12 @@ What do those abbreviations mean?
 - L: Matches Lost
 - P: Points
 
-A win earns a team 3 points. A draw earns 1. A loss earns 0.
+A win earns a team 3 points.
+A draw earns 1.
+A loss earns 0.
 
-The outcome should be ordered by points, descending. In case of a tie, teams are ordered alphabetically.
+The outcome should be ordered by points, descending.
```
kotp, line 27: This wording has not changed, _should_ is different than _must_, but now there is a test that is enforcing that this must happen.

kotp (approved):
There is a new test that ensures that the points are sorted numerically, descending.  But the wording should match the new test.

We can change it here and pursue it in Problem Specifications (or leave the test out, for now).

Approving now, but delaying applying the patch for consideration of the language change.

exercises/practice/tournament/.docs/instructions.md
kytrinyx: https://github.com/exercism/problem-specifications/pull/2135

iHiD (approved):

PR #1483, opened 2023-01-16
Update documentation to fix numbering issue.

tekwizz123:
The amusement-park exercise has incorrect numbering and repeats the 4th instruction as an instruction number when it should be 5 instead.


kotp:
Approving.

kotp (approved):

PR #1485, opened 2023-01-16
Update explanation of floating point numbers for Ruby Assembly Line exercise

tekwizz123:
A floating point number has at least one number after the decimal point, not zero or more.

This explanation conflicts with the integers explanation which states its "numbers with no digits behind the decimal separator".

### DIFF
--- a/exercises/concept/assembly-line/.docs/introduction.md
+++ b/exercises/concept/assembly-line/.docs/introduction.md
@@ -5,7 +5,7 @@
 The two most common types of numbers in Ruby are:
 
 - **Integers:** numbers with no digits behind the decimal separator (whole numbers). Examples are `-6`, `0`, `1`, `25`, `976` and `500000`.
-- **Floating-point numbers:** numbers with zero or more digits behind the decimal separator. Examples are `-2.4`, `0.1`, `3.14`, `16.984025` and `1024.0`.
+- **Floating-point numbers:** numbers with one or more digits behind the decimal separator. Examples are `-2.4`, `0.1`, `3.14`, `16.984025` and `1024.0`.
 
 They are implemented through the `Integer` and `Float` classes.
 


kotp:
@tekwizz123 I clicked on the "this link" and made the new topic [here].

[here]: https://forum.exercism.org/t/update-explanation-of-floating-point-numbers/2757

iHiD:
Verifying that here (I didn't actually know if `5.` was an `Integer` or a `Float`:

```
irb(main):010:0> (5. + 3).class
=> Integer
irb(main):011:0> (5.0 + 3).class
=> Float
```

kotp:
> Verifying that here (I didn't actually know if `5.` was an `Integer` or a `Float`:
> 
> ```
> irb(main):010:0> (5. + 3).class
> => Integer
> irb(main):011:0> (5.0 + 3).class
> => Float
> ```

It is not, it is really not `5.` since the `.` is really attached to the `+`.  As we can see here:

```irb
>> 5.+(3)
=> 8
```

So it is a bit tricky, as what we think we are seeing as a number `5.` it is only the `5` that is the number.

iHiD:
Excellent. Thanks for clarifying. This explains why my brain had no record for `5.` syntax! 😁

kotp:
Let's expand this to be better in whole, since the Integer description talks about some digits behind the decimal point, but there is no decimal point in an Integer.

Also, the commit message is not quite in compliance with git recommendations.  The commit summary is at a length of 77, when it should be less than 50[¹][git-scm] . This will cause some problems with the GUI tools with displaying, and the subject line of the e-mail message for this patch being that long is problematic as well.

See the [git-scm] discussion section on _commit_ for more information., or if you prefer not to go out to the web, you can see the **DISCUSSION** section from `man git-commit` for the same information.

[git-scm]: https://git-scm.com/docs/git-commit#_discussion

kytrinyx:
@kotp what is the status of this PR? It looks like it has been approved, but it's not merged. Is there a reason why it's being held further?

kotp:
In terms of me not lighting the fire, was hoping @tekwizz123 would confirm.  But it has been long enough.  And things are noted, so it surely can come in.

PR #1494, opened 2023-02-27
Sync gigasecond docs with problem-specifications

kytrinyx:
The gigasecond exercise has been overhauled as part of a project
to make practice exercises more consistent and friendly.

For more context, please see the discussion in the forum, as well as
the pull request that updated the exercise in the problem-specifications
repository:

- https://forum.exercism.org/t/new-project-making-practice-exercises-more-consistent-and-human-across-exercism/3943
- https://github.com/exercism/problem-specifications/pull/2206

### DIFF
--- a/exercises/practice/gigasecond/.docs/instructions.md
+++ b/exercises/practice/gigasecond/.docs/instructions.md
@@ -1,5 +1,8 @@
 # Instructions
 
-Given a moment, determine the moment that would be after a gigasecond has passed.
+Your task is to determine the date and time one gigasecond after a certain date.
 
-A gigasecond is 10^9 (1,000,000,000) seconds.
+A gigasecond is one thousand million seconds.
+That is a one with nine zeros after it.
+
+If you were born on _January 24th, 2015 at 22:00 (10:00:00pm)_, then you would be a gigasecond old on _October 2nd, 2046 at 23:46:40 (11:46:40pm)_.
--- a/exercises/practice/gigasecond/.docs/introduction.md
+++ b/exercises/practice/gigasecond/.docs/introduction.md
@@ -1,0 +1,24 @@
+# Introduction
+
+The way we measure time is kind of messy.
+We have 60 seconds in a minute, and 60 minutes in an hour.
+This comes from ancient Babylon, where they used 60 as the basis for their number system.
+We have 24 hours in a day, 7 days in a week, and how many days in a month?
+Well, for days in a month it depends not only on which month it is, but also on what type of calendar is used in the country you live in.
+
+What if, instead, we only use seconds to express time intervals?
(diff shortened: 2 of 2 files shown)


exercises/practice/gigasecond/.docs/instructions.md
```diff
@@ -1,5 +1,8 @@
 # Instructions
 
-Given a moment, determine the moment that would be after a gigasecond has passed.
+Your task is to determine the date and time one gigasecond after a certain date.
```
kotp, line 3: Does this too strongly suggest that we should get a Time or Date object back?  It is not required, but used for a few reasons in the tests themselves.

If the unit were not so large, a Degree would also work since the base unit of a degree is a second, and integer solutions work well (as do float solutions) because the base unit of Time is the second, and it is represented as a `Numeric`.

kotp (approved):
Approved with a query.  It probably won't change mentoring for me, but it could end up driving the tests later in the track.

If this is coming from upstream, then of course, this is not the place to talk about it, exactly, but the generator may need to deal with the differences.

exercises/practice/gigasecond/.docs/instructions.md
kytrinyx: No, I don't think this says anything about how the date and time is implemented.

kotp: That is true, but now uses date and time language specifically, where before it suggested only the _simple_ idea of a gigasecond more than a number of seconds given.

So, does it suggest that we should use date or time specifically?  Humans tend to easily read too much into statements as it is.  Rather than discarding irrelevant information, which often helps in solving problems.

But it does not really matter, since these things can be addressed via mentoring.  I already talk about the potential "reading too much into" which often happens due to the tests choice to use Time here, rather than the simpler addition of an integer solution that would and does still work.

kytrinyx: I chose to say "date and time" very explicitly, because that is how we talk about these things in the real world. This is an attempt to ensure that the language used is more in the realm of day-to-day conversation than in the realm of programming syntax.

I'm completely okay with people reading too much into it. People do that all the time no matter what we do and say. It will happen in their day-to-day life as well. We do not need to take every possible opportunity to ensure to the best of our ability that this does not happen in this simple exercise.

And, as you say, if people choose to be mentored, this is something that can be raised then.

kotp: I can work with it either way.   This adds more fodder to the aspect of the mentoring in my snippets about "Boss Speak" which sometimes hold misdirection to solutions.

Sounds like more fun in mentoring this way. ;)

PR #1499, opened 2023-03-03
Clarifies instructions

rebelwarrior:
Adds clarity to what max_factor is and that you must return a triplet up


kotp:
[This discussion] is available for background.

[This discussion]: https://forum.exercism.org/t/the-ruby-pythagorean-triplet-exercise-has-something-wrong-with-it-1497/4259

PR #1502, opened 2023-03-07
Sync binary-search docs with problem-specifications

kytrinyx:
The binary-search exercise has been overhauled as part of a project to make practice exercises more consistent and friendly.

For more context, please see the discussion in the forum, as well as the pull request that updated the exercise in the problem-specifications repository:

- https://forum.exercism.org/t/new-project-making-practice-exercises-more-consistent-and-human-across-exercism/3943
- https://github.com/exercism/problem-specifications/pull/2220

### DIFF
--- a/exercises/practice/binary-search/.docs/instructions.md
+++ b/exercises/practice/binary-search/.docs/instructions.md
@@ -1,24 +1,28 @@
 # Instructions
 
-Implement a binary search algorithm.
+Your task is to implement a binary search algorithm.
 
-Searching a sorted collection is a common task.
-A dictionary is a sorted list of word definitions.
-Given a word, one can find its definition.
-A telephone book is a sorted list of people's names, addresses, and telephone numbers.
-Knowing someone's name allows one to quickly find their telephone number and address.
+A binary search algorithm finds an item in a list by repeatedly splitting it in half, only keeping the half which contains the item we're looking for.
+It allows us to quickly narrow down the possible locations of our item until we find it, or until we've eliminated all possible locations.
 
-If the list to be searched contains more than a few items (a dozen, say) a binary search will require far fewer comparisons than a linear search, but it imposes the requirement that the list be sorted.
+```exercism/caution
+Binary search only works when a list has been sorted.
+```
 
-In computer science, a binary search or half-interval search algorithm finds the position of a specified input value (the search "key") within an array sorted by key value.
+The algorithm looks like this:
 
-In each step, the algorithm compares the search key value with the key value of the middle element of the array.
+- Divide the sorted list in half and compare the middle element with the item we're looking for.
+- If the middle element is our item, then we're done.
+- If the middle element is greater than our item, we can eliminate that number and all the numbers **after** it.
+- If the middle element is less than our item, we can eliminate that number and all the numbers **before** it.
+- Repeat the process on the part of the list that we kept.
 
-If the keys match, then a matching element has been found and its index, or position, is returned.
+Here's an example:
 
-Otherwise, if the search key is less than the middle element's key, then the algorithm repeats its action on the sub-array to the left of the middle element or, if the search key is greater, on the sub-array to the right.
+Let's say we're looking for the number 23 in the following sorted list: `[4, 8, 12, 16, 23, 28, 32]`.
 
-If the remaining array to be searched is empty, then the key cannot be found in the array and a special "not found" indication is returned.
-
-A binary search halves the number of items to check with each iteration, so locating an item (or determining its absence) takes logarithmic time.
-A binary search is a dichotomic divide and conquer search algorithm.
+- We start by comparing 23 with the middle element, 16.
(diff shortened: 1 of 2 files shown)


exercises/practice/binary-search/.docs/instructions.md
```diff
@@ -1,24 +1,28 @@
...
-If the list to be searched contains more than a few items (a dozen, say) a binary search will require far fewer comparisons than a linear search, but it imposes the requirement that the list be sorted.
+```exercism/caution
+Binary search only works when a list has been sorted.
+```
 
-In computer science, a binary search or half-interval search algorithm finds the position of a specified input value (the search "key") within an array sorted by key value.
+The algorithm looks like this:
 
-In each step, the algorithm compares the search key value with the key value of the middle element of the array.
+- Divide the sorted list in half and compare the middle element with the item we're looking for.
```
kotp, line 14: We could emphasize the "sorted" as it is a requirement of the algorithm.

```suggestion
- Divide the _sorted_ list in half and compare the middle element with the item we're looking for.
```

exercises/practice/binary-search/.docs/introduction.md
```diff
@@ -0,0 +1,9 @@
+# Introduction
+
+You have stumbled upon a group of mathematicians who are also singer-songwriters.
+They have written a song for each of their favorite numbers, and, as you can imagine, they have a lot of favorite numbers.
```
kotp, line 4: Should we be more explicit in which meaning of "number" we mean?   When I first read this, I took it as the musical number, rather than a numeric number, even having seen that they were mathematicians doing this.

Then I had to think about what it meant, was it referring to the ditties that they liked, or their favorite scores, or how did they separate those arrangements.  Then, and only after, did I realize that in this context, we must mean numeric numbers, not musical numbers.

For some reason the music subject took over the context of the vocabulary used here.  Even after having read about numeric numbers in this review in the instructions file.

kotp (approved):
Approving, with a suggestion of emphasizing the sorted requirement.

And mentioning the different meanings of the word `number` and that I might be strange (hmpfh) but it took me some effort to realize what we were talking about.

exercises/practice/binary-search/.docs/introduction.md
kytrinyx: We're going to need to make this change upstream, rather than here. I will open a PR tomorrow!

exercises/practice/binary-search/.docs/instructions.md
kytrinyx: We have an `exercism/caution` block right above. I _think_ (but am not 100% confident) that this should be enough to not need emphasizing within the sentence itself.

What do you think?

kotp: It could very well be enough if it is there.  It is also repeated, and so the repetition itself is an emphasis.  A little bit of markdown for emphasis could help to bring that up as being an important requirement though.  If we were speaking, I would inflect on that single word, just a little bit.

exercises/practice/binary-search/.docs/introduction.md
iHiD: Yeah, for both this and the other comment about emphasing the word, they'd need to happen upstream not here. I don't think the emphasising one would be worth repropogating across all 65 tracks as that's lots of noise for maintainers, but I think this one might be.

kotp: If we take care of both at the same time, since the change is coming, would it be an appropriate time for it?  Perhaps in two commits, so that if a track wanted one but not the other, they could choose to do so.

kytrinyx: The changes from problem-specifications get added to the tracks by using configlet sync. They get all of the changes or none of the changes.

I won't be making hand-crafted PRs (or even lovingly scripted PRs) to allow people to cherry-pick an emphasis that does not meaningfully change the instructions.

I feel quite strongly that the only good reason to not accept a PR that syncs from problem-specs is if the instructions **contradict** the exercise in that track.

kytrinyx: @kotp I have opened the PR here: https://github.com/exercism/problem-specifications/pull/2226

If you like, feel free to suggest a PR to change the emphasis.

There is another PR open to the same exercise as well, so I am happy to wait to put in the new PRs to the tracks until all three PRs have been merged.

exercises/practice/binary-search/.docs/instructions.md
kytrinyx: This is what it looks like on the site, for context.

Screen Shot 2023-03-08 at 12 12 45 PM

exercises/practice/binary-search/.docs/introduction.md
kotp: Opened a [PR][pr] for the emphasis.  (Also, yes, the asterisks for bold rather than italics for emphasis, @iHiD .

[pr]:  exercism/problem-specifications#2227

PR #1506, opened 2023-03-10
Rewrite pythagorean-triplet

kytrinyx:
The pythagorean-triplet exercise wasn't following the spec
even a little bit.

This rewrites the test suite to follow the existing
canonical data, and rewrites the solution to make it
pass.

The last test takes a long time to run, which isn't ideal but I couldn't find a way around it without going all mathlete on it.

See discussion here: https://forum.exercism.org/t/the-ruby-pythagorean-triplet-exercise-has-something-wrong-with-it-1497/4259/6

### DIFF
--- a/exercises/practice/pythagorean-triplet/.meta/example.rb
+++ b/exercises/practice/pythagorean-triplet/.meta/example.rb
@@ -1,13 +1,11 @@
 class PythagoreanTriplet
   def self.triplets_with_sum(sum)
     triplets = []
-    (2..(sum/2)).to_a.reverse.each do |c|
+    (2..(sum / 2)).to_a.reverse.each do |c|
       a = 1
-      b = sum-c-a
+      b = sum - c - a
       while a < b
-        if pythagorean?(a, b, c)
-          triplets << [a, b, c]
-        end
+        triplets << [a, b, c] if pythagorean?(a, b, c)
         a += 1
         b -= 1
       end
@@ -16,6 +14,6 @@ class PythagoreanTriplet
   end
 
   def self.pythagorean?(a, b, c)
-    (a*a + b*b == c*c)
+    (a * a + b * b == c * c)
   end
 end
--- a/exercises/practice/pythagorean-triplet/.meta/example.rb
+++ b/exercises/practice/pythagorean-triplet/.meta/example.rb
@@ -1,52 +1,21 @@
-class Triplets
-  attr_reader :factors, :sum
-  def initialize(conditions)
-    min = conditions.fetch(:min_factor) { 1 }
-    max = conditions.fetch(:max_factor)
-    @sum = conditions[:sum]
-    @factors = (min..max).to_a
-  end
-
-  def to_a
+class PythagoreanTriplet
+  def self.triplets_with_sum(sum)
     triplets = []
-    each_triplet do |triplet|
-      triplets << triplet if select?(triplet)
+    (2..(sum/2)).to_a.reverse.each do |c|
+      a = 1
+      b = sum-c-a
+      while a < b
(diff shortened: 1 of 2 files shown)


exercises/practice/pythagorean-triplet/.meta/example.rb
```diff
@@ -1,52 +1,21 @@
...
-
-  def to_a
+class PythagoreanTriplet
+  def self.triplets_with_sum(sum)
     triplets = []
-    each_triplet do |triplet|
-      triplets << triplet if select?(triplet)
+    (2..(sum/2)).to_a.reverse.each do |c|
+      a = 1
+      b = sum-c-a
```
kotp, line 6:
```suggestion
      b = sum - c - a
```

Space around operators, this looks different than other areas of this same file.

exercises/practice/pythagorean-triplet/.meta/example.rb
```diff
@@ -1,52 +1,21 @@
...
-  end
-
-  def product
-    a * b * c
-  end
-
-  def pythagorean?
-    a**2 + b**2 == c**2
+  def self.pythagorean?(a, b, c)
+    (a*a + b*b == c*c)
```
kotp: Similar, space around `*` to be similar to `+` and `-`, but not like `**` which should not have space around it.

exercises/practice/pythagorean-triplet/.meta/example.rb
kytrinyx: I must have forgotten to run rubocop autocorrect on it.

kotp (approved):
Example is not exemplar, and so the suggestions here really do not matter, so approving.

Good improvement, and linked this to the [discussion] that prompted this change.

[discussion]: https://forum.exercism.org/t/the-ruby-pythagorean-triplet-exercise-has-something-wrong-with-it-1497/4259/8

kotp:
Thank you @kytrinyx !

exercises/practice/pythagorean-triplet/.meta/example.rb
```diff
@@ -1,52 +1,19 @@
...
-    @factors = (min..max).to_a
-  end
-
-  def to_a
+class PythagoreanTriplet
+  def self.triplets_with_sum(sum)
     triplets = []
-    each_triplet do |triplet|
-      triplets << triplet if select?(triplet)
+    (2..(sum / 2)).to_a.reverse.each do |c|
```
kotp, line 4: Funny, or interesting, that Rubocop considers the space around `/` here, but does not consider "too many parenthesis".  I wrote (and then deleted) a mention about this line with the "Perfect use of no space around `/`" because it communicates well the idea of "half of sum" as a fraction.

But `(2..sum/2).to_a` is valid, the lack of space around `/` communicating the right thing in terms of its "closeness" to `sum`.

Anyway, thought I would mention it here, since the change was made (and merged).

kytrinyx: Yeah, I did that intentionally (same with the `a*a` lower down), but I'm totally fine with just going with whatever Rubocop wants here :)

kytrinyx:
Thanks for taking the time to work through this, @kotp

rebelwarrior:
I like Katrina's non math approach. It's actually clearer to read. I like that it's only two (whatchamacallit) cycles. The while loop that reduces one an increases the other is pretty cool. 

It can be done in one but requires doing a square root. A Rust mentor helped me get it:
You cycle through `a` then 
b = ((b + c)² - a²) / (2(b + c))
and
c = √(a² + b²)
I wonder if I can parallelize it in Ruby too.

PR #1523, opened 2023-04-22
Update microwave_test.rb - Corrected initialised input for test cases

teyofnull:
Several lines in the test file treated 1 minute as 100 seconds.

This pull request corrects this and allows for the test cases to pass if supplied with valid input.

### DIFF
--- a/exercises/practice/microwave/microwave_test.rb
+++ b/exercises/practice/microwave/microwave_test.rb
@@ -15,7 +15,7 @@ class MicrowaveTest < Minitest::Test
   end
 
   def test_one_minute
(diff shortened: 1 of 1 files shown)


kotp:
This is as designed.  If you pass in 60 in your microwave you should get 1:00, and if you press 100 in  your microwave, you should also get 1:00, in both cases getting 1 minute.

teyofnull:
Thanks @kotp. Thought I had caught a bug.

kotp:
It is often good to get mentoring and talk with the mentor first.  They can often explain what is happening, and why.

PR #1525, opened 2023-04-26
Don't catch load error in hello-world

kytrinyx:
Minitest version 5.0.0 was released in May 2013.
That's 10 years ago. We don't need to worry about this anymore.

### DIFF
--- a/exercises/practice/hello-world/hello_world_test.rb
+++ b/exercises/practice/hello-world/hello_world_test.rb
@@ -1,15 +1,5 @@
-begin
-  gem 'minitest', '>= 5.0.0'
-  require 'minitest/autorun'
-  require_relative 'hello_world'
-rescue Gem::LoadError => e
-  puts "\nMissing Dependency:\n#{e.backtrace.first} #{e.message}"
-  puts 'Minitest 5.0 gem must be installed for the Ruby track.'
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):

exercises/practice/hello-world/hello_world_test.rb
```diff
@@ -1,15 +1,5 @@
-begin
-  gem 'minitest', '>= 5.0.0'
-  require 'minitest/autorun'
-  require_relative 'hello_world'
-rescue Gem::LoadError => e
-  puts "\nMissing Dependency:\n#{e.backtrace.first} #{e.message}"
-  puts 'Minitest 5.0 gem must be installed for the Ruby track.'
-rescue LoadError => e
-  puts "\nError:\n#{e.backtrace.first} #{e.message}"
-  puts DATA.read
```
kotp, line 10: If this is removed, should we also remove the source of the data?

kotp (changes requested):

exercises/practice/hello-world/hello_world_test.rb
kotp: Never mind, I see it in #1527

kotp (approved):

PR #1548, opened 2023-05-29
Update instructions.md

imechemi:


kotp (commented):
This instruction probably comes directly from upstream in https://github.com/exercism/problem-specifications .

PR #1556, opened 2023-07-25
Add better stub to armstrong numbers

iHiD:
Fixes https://github.com/exercism/ruby/issues/1555#issuecomment-1649998663

### DIFF
--- a/exercises/practice/armstrong-numbers/armstrong_numbers.rb
+++ b/exercises/practice/armstrong-numbers/armstrong_numbers.rb
@@ -1,7 +1,5 @@
-=begin
-Write your code for the 'Armstrong Numbers' exercise in this file. Make the tests in
-`armstrong_numbers_test.rb` pass.
-
-To get started with TDD, see the `README.md` file in your
-`ruby/armstrong-numbers` directory.
-=end
+module ArmstrongNumbers
+  def self.include?(number)
+    # Write your code here
+  end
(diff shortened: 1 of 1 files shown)


exercises/practice/armstrong-numbers/armstrong_numbers.rb
```diff
@@ -1,7 +1,5 @@
-=begin
-Write your code for the 'Armstrong Numbers' exercise in this file. Make the tests in
-`armstrong_numbers_test.rb` pass.
-
-To get started with TDD, see the `README.md` file in your
-`ruby/armstrong-numbers` directory.
-=end
+module ArmstrongNumbers
```
kotp, line 1: Are we removing the comment in the stubs now?  Is this no longer considered helpful?

kotp (approved):
Approving in case we are getting prepared to remove the reminder comment, but I think even with the stub, having the helpful comment of where to read for help is helpful in this file.

ErikSchierboom (approved):

exercises/practice/armstrong-numbers/armstrong_numbers.rb
iHiD: All these comments should be removed or amended now. They don't make sense in the context of the online editor. The website should be informing someone what to do - not a comment in a file they see.

PR #1557, opened 2023-07-26
[Kindergarten Garden] Fix a mention of dynamic programming.

JakubSarnik:
See a discussion over at [the forum](http://forum.exercism.org/t/dynamic-programming-in-kindergarten-garden/6609) for why the original text is not really correct. The suggested rephrasing was also discussed there.

### DIFF
--- a/exercises/practice/kindergarten-garden/.docs/instructions.append.md
+++ b/exercises/practice/kindergarten-garden/.docs/instructions.append.md
@@ -2,4 +2,4 @@
 
 The exercise requires you to write a class that provides a method for each child in the kindergarten.
 You can make your work less repetitive (and more fun!) if, instead of manually defining 12 different methods, you use a bit of metaprogramming.
-If you're not sure where to start, explore the [`method_missing`](https://www.leighhalliday.com/ruby-metaprogramming-method-missing) or [`define_method`](http://rubymonk.com/learning/books/2-metaprogramming-ruby/chapters/25-dynamic-methods/lessons/72-define-method) methods.
+If you're not sure where to start, explore the `method_missing` or `define_method` methods.
--- a/exercises/practice/kindergarten-garden/.docs/instructions.append.md
+++ b/exercises/practice/kindergarten-garden/.docs/instructions.append.md
@@ -1,4 +1,5 @@
 # Instructions append
 
-In this exercise you're going to be using a concept called dynamic programming. https://en.wikipedia.org/wiki/Dynamic_programming
-Your solution could use the class method `#define_singleton_method` or the combination of `#method_missing` and `#respond_to_missing`.
+The exercise requires you to write a class that provides a single method for each child in the kindergarten, respectively.
+You can make your work easier and less repetitive if, instead of manually defining 12 different methods, you use a bit of metaprogramming.
+Your solution could use the method `#define_method` (or `#define_singleton_method`), or alternatively a combination of `#method_missing` and `#respond_to_missing?`.
--- a/exercises/practice/kindergarten-garden/.docs/instructions.append.md
+++ b/exercises/practice/kindergarten-garden/.docs/instructions.append.md
@@ -1,5 +1,5 @@
 # Instructions append
 
 The exercise requires you to write a class that provides a method for each child in the kindergarten.
-You can make your work easier and less repetitive if, instead of manually defining 12 different methods, you use a bit of metaprogramming.
+You can make your work less repetitive (and more fun!) if, instead of manually defining 12 different methods, you use a bit of metaprogramming.
 Your solution could use the method `#define_method` (or `#define_singleton_method`), or alternatively a combination of `#method_missing` and `#respond_to_missing?`.
--- a/exercises/practice/kindergarten-garden/.docs/instructions.append.md
+++ b/exercises/practice/kindergarten-garden/.docs/instructions.append.md
@@ -1,5 +1,5 @@
 # Instructions append
 
-The exercise requires you to write a class that provides a single method for each child in the kindergarten, respectively.
+The exercise requires you to write a class that provides a method for each child in the kindergarten.
 You can make your work easier and less repetitive if, instead of manually defining 12 different methods, you use a bit of metaprogramming.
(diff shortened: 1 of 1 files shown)


exercises/practice/kindergarten-garden/.docs/instructions.append.md
```diff
@@ -1,4 +1,8 @@
 # Instructions append
 
-In this exercise you're going to be using a concept called dynamic programming. https://en.wikipedia.org/wiki/Dynamic_programming
-Your solution could use the class method `#define_singleton_method` or the combination of `#method_missing` and `#respond_to_missing`.
+The exercise requires you to write a class that provides a single method for
+each child in the kindergarten, respectively. You can make your work easier and
+less repetitive if, instead of manually defining 12 different methods, you use
+a bit of metaprogramming. Your solution could use the method `#define_method`
+(or `#define_singleton_method`), or alternatively a combination of
+`#method_missing` and `#respond_to_missing?`.
```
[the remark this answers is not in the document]
JakubSarnik: Done!

meatball133:
I suggest linking to those methods from the Ruby API docs.

exercises/practice/kindergarten-garden/.docs/instructions.append.md
```diff
@@ -1,4 +1,5 @@
 # Instructions append
 
-In this exercise you're going to be using a concept called dynamic programming. https://en.wikipedia.org/wiki/Dynamic_programming
-Your solution could use the class method `#define_singleton_method` or the combination of `#method_missing` and `#respond_to_missing`.
+The exercise requires you to write a class that provides a single method for each child in the kindergarten, respectively.
```
iHiD, line 3:
```suggestion
The exercise requires you to write a class that provides a method for each child in the kindergarten.
```

exercises/practice/kindergarten-garden/.docs/instructions.append.md
```diff
@@ -1,4 +1,5 @@
...
+You can make your work easier and less repetitive if, instead of manually defining 12 different methods, you use a bit of metaprogramming.
```
iHiD, line 4:
```suggestion
  You can make your work less repetitive (and more fun!) if, instead of manually defining 12 different methods, you use a bit of metaprogramming.
```

exercises/practice/kindergarten-garden/.docs/instructions.append.md
```diff
@@ -1,4 +1,5 @@
...
+Your solution could use the method `#define_method` (or `#define_singleton_method`), or alternatively a combination of `#method_missing` and `#respond_to_missing?`.
```
iHiD, line 5:
```suggestion
If you're not sure where to start, explore the [`method_missing`](https://www.leighhalliday.com/ruby-metaprogramming-method-missing) or [`define_method`](http://rubymonk.com/learning/books/2-metaprogramming-ruby/chapters/25-dynamic-methods/lessons/72-define-method) methods.
```

iHiD (commented):
Thanks for this. I've added a few suggestions :)

exercises/practice/kindergarten-garden/.docs/instructions.append.md
JakubSarnik: I have no problem with the rephrasing, but I kinda dislike the fact that the `method_missing` article overrides `respond_to?` (and only relatively late in the text) instead of `respond_to_missing?`, which would be more safe and idiomatic, as I am lead to believe (although I am by no means an expert on Ruby).

Problem is, I cannot find an article that would explain it more (only those that already presume the reader knows what `method_missing` is, e.g. https://thoughtbot.com/blog/always-define-respond-to-missing-when-overriding).

exercises/practice/kindergarten-garden/.docs/instructions.append.md
JakubSarnik: I have no problems with this and the above suggestions. Are you, by the way, OK with amending the commit even for such more substantial changes, or do you prefer a dedicated new commit?

iHiD: It's probably better to have a different PR for bigger changes. These appends should be small hints really, not explanations about how to do things though, so I wouldn't expect the text to be too long here :)

exercises/practice/kindergarten-garden/.docs/instructions.append.md
iHiD: I still consider `respond_to_missing?` to be the new kid on the block, although you're very right :) 

OK, Let's stick with these links for now and as soon as we have better Concept pages on Exercism for these in Ruby, we'll switch to them.

@meatball133 I don't find the Ruby docs to be clear enough for Ruby newbies really. I'd rather like to an explanation tutorial/blog (ideally our concept pages!)

kotp (commented):
I like that they are Ruby-centric links, but not happy that `respond_to_missing?` does not seem to be covered.  This is pretty essential to doing it right, with `method_missing`.  It has been in the language since around 2009, and I would say it is expected to become known in a lesson about `method_missing`.

iHiD:
Happy to have alternative blog posts suggested. But I feel strongly that the "appends" isn't the place for us to start teaching people how to do things properly. Again, ideally this points to our concept docs once they get written.

JakubSarnik:
> Happy to have alternative blog posts suggested. But I feel strongly that the "appends" isn't the place for us to start teaching people how to do things properly. Again, ideally this points to our concept docs once they get written.

Well in that case it would maybe be better to just link to the methods in API documentation, as @meatball133 suggested.

kotp:
> Happy to have alternative blog posts suggested. But I feel strongly that the "appends" isn't the place for us to start teaching people how to do things properly. Again, ideally this points to our concept docs once they get written.

We do not yet have a concept exercise for metaprogramming or dynamic programming, and would be more of a reminder, but for sure the "start" should be in a concept exercise.

> Well in that case it would maybe be better to just link to the methods in API documentation, as @meatball133 suggested.

It is a practice exercise, not a concept exercise, so I think that is the way to go.

iHiD:
I've just removed the links for now. I'll merge. Thanks all.

PR #1560, opened 2023-08-04
[New Concept]: Symbols

meatball133:
@iHiD

### DIFF
--- a/concepts/symbols/.meta/config.json
+++ b/concepts/symbols/.meta/config.json
@@ -1,0 +1,5 @@
+{
+  "blurb": "Symbols are identifiers and are created through the symbol literal. They are non mutable, which makes them useful for things like keys in a hashes.",
+  "authors": ["meatball133"],
+  "contributors": []
+}
--- a/concepts/symbols/.meta/config.json
+++ b/concepts/symbols/.meta/config.json
@@ -1,5 +1,5 @@
 {
   "blurb": "Symbols are identifiers and are created through the symbol literal. They are non mutable, which makes them useful for things like keys in a hashes.",
   "authors": ["meatball133"],
-  "contributors": []
+  "contributors": ["kotp"]
 }
--- a/concepts/symbols/.meta/config.json
+++ b/concepts/symbols/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "Symbols are identifiers and are created through the symbol literal. They are non mutable, which makes them useful for things like keys in a hashes.",
+  "blurb": "Symbols are very important in Ruby.  They are used as \"identity\".  They are immutable, guaranteed to be a single item no matter where it is used or referenced.  They are used internally as variable, method, constants, and keyword identifiers.  They are also useful, and often used, as keys in a `Hash`.",
   "authors": ["meatball133"],
   "contributors": ["kotp"]
 }
--- a/concepts/symbols/about.md
+++ b/concepts/symbols/about.md
@@ -13,7 +13,7 @@ Symbols are used in many places in the language, including as keys in hashes, to
 
 ## Identifier
 
-What makes symbols different from strings is that they are identifiers, and not data or text.
+What makes symbols different from strings is that they are identifiers, and do not represent data or text.
 This means that two symbols with the same name are always the same object.
 
 ```ruby
--- a/concepts/symbols/about.md
+++ b/concepts/symbols/about.md
@@ -27,7 +27,7 @@ This means that two symbols with the same name are always the same object.
 
 Symbols are immutable, which means that they cannot be modified.
 This means that when you "modify" a symbol, you are actually creating a new symbol.
-There are a few methods that can be used to modify symbols, but they all return new symbols.
+There are a few methods that can be used to manipulate symbols, they all return new symbols.
 All methods can be found in the [Symbol API][symbols-api].
 
 ```ruby
--- a/concepts/symbols/about.md
+++ b/concepts/symbols/about.md
@@ -43,7 +43,7 @@ The benefit of symbols being immutable is that they are more memory efficient th
 
 Symbols can be converted to strings and vice versa.
 This can be useful when you want to modify a symbol, or when you want to use a symbol as a string.
-To convert a string to a symbol, you can use the `String#to_sym` method, and to do the opposite, you can use the `Symbol#to_s` method.
+To present a symbol as a string, you can use the `String#to_sym` method, and to do the opposite, you can use the `Symbol#to_s` method.
 Due to symbols having a limited set of methods, it can be useful to convert a symbol to a string to use string methods on it, if a new symbol is needed.
 
 ```ruby
--- a/concepts/symbols/about.md
+++ b/concepts/symbols/about.md
@@ -2,7 +2,7 @@
 
 [Symbols][symbols] are named identifiers that can be used to refer to a value.
 Symbols are created through a symbol literal, which is by prefixing a name with a `:` character, e.g. `:foo`.
-They also allow for being written as strings, e.g. `:"foo"`, which allows for example spaces in the name.
+They also allow for being written with quotes, e.g. `:"foo"`, which allows, for example, spaces in the name.
 
 ```ruby
 :foo # => :foo
--- a/concepts/symbols/about.md
+++ b/concepts/symbols/about.md
@@ -56,7 +56,7 @@ Due to symbols having a limited set of methods, it can be useful to convert a sy
 In Ruby you can get names in scopes by using various methods to get names of constants, methods, and variables.
 These methods returns arrays of symbols.
 
-The most common methods are `Module#constants`, `Module#instance_methods`, `Module#class_variables`, `global_variables` and `local_variables`.
+Some methods are `Module#constants`, `Module#instance_methods`, `Module#class_variables`, `global_variables` and `local_variables`.
 
 ```ruby
 module Foo
--- a/concepts/symbols/about.md
+++ b/concepts/symbols/about.md
@@ -44,7 +44,7 @@ The benefit of symbols being immutable is that they are more memory efficient th
 Symbols can be converted to strings and vice versa.
 This can be useful when you want to modify a symbol, or when you want to use a symbol as a string.
 To convert a string to a symbol, you can use the `String#to_sym` method, and to do the opposite, you can use the `Symbol#to_s` method.
-Due to symbols having a limited set of methods, it can be useful to convert a symbol to a string to use string methods on it, if a modification is needed.
+Due to symbols having a limited set of methods, it can be useful to convert a symbol to a string to use string methods on it, if a new symbol is needed.
 
 ```ruby
 :foo.to_s # => "foo"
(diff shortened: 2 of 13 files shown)


concepts/symbols/.meta/config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "blurb": "Symbols are identifiers and are created through the symbol literal. They are non mutable, which makes them useful for things like keys in a hashes.",
```
kotp, line 2: Symbols are these things, but they are used as identity.  They are, of course, used as keys for Hash objects, but we could mention that they are used as the internal names for constants (including module and class names), variables, method names, and keywords.  They are not only immutable, but they are singletons, as well.

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
+# About
+
+[Symbols][symbols] are named identifiers that can be used to refer to a value.
+Symbols are created through a symbol literal, which is by prefixing a name with a `:` character, e.g. `:foo`.
+They also allow for being written as strings, e.g. `:"foo"`, which allows for example spaces in the name.
```
kotp, line 5:
```suggestion
They also allow for being written with quotes, e.g. `:"foo"`, which allows, for example, spaces in the name.
```

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
...
+
+```ruby
+:foo # => :foo
+:"foo boo" # => :"foo boo"
+```
+
+Symbols are used in many places in the language, including as keys in hashes, to represent method names and variable names.
```
kotp, line 12: Maybe this is where we expand on that where it is used as the identifiers for methods and local variables.

As an aside, it is interesting that constants, keywords, methods, and local variables all are the exception to the rule of "Everything is an object".

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Identifier
+
+What makes symbols different from strings is that they are identifiers, and not data or text.
```
kotp, line 16: They are not "text", but then what is "text"?  In Ruby, it is not a thing.

I think they are, and can be used as, data.  When mentoring for RubyLearning I would say "symbols are used for their identity, the data is the identity is the symbol".

Fabio Akita wrote:

> You can also think of `:id` as meaning the name of the variable `id`, and plain `id` as meaning the value of the variable.

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
...
+"foo".object_id # => 80
+:foo.object_id # => 1086748
+:foo.object_id # => 1086748
+```
+
+## Modifying Symbols
+
+Symbols are immutable, which means that they cannot be modified.
+This means that when you "modify" a symbol, you are actually creating a new symbol.
+There are a few methods that can be used to modify symbols, but they all return new symbols.
```
kotp, line 30: Should we state ways to not modify symbols by using the word modify?  None of the ways listed modify, and we just stated that they are immutable and can not be modified.

Instead, we might talk about how a symbol is specifically and always one and only that one thing.

kotp, line 30:
```suggestion
There are a few methods that can be used to manipulate symbols, they all return new symbols.
```

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
...
+
+:foo.object_id # => 1086748
+:foo.upcase.object_id # => 60
+```
+
+The benefit of symbols being immutable is that they are more memory efficient than strings, but also safer to use as identifiers.
+
+## Conversion
+
+Symbols can be converted to strings and vice versa.
```
kotp, line 44: How do we state this without suggesting that they are really converted, or modified, to be a string?

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
...
+This can be useful when you want to modify a symbol, or when you want to use a symbol as a string.
```
kotp, line 45: Since we can not modify a symbol, this should be stated differently.

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
...
+To convert a string to a symbol, you can use the `String#to_sym` method, and to do the opposite, you can use the `Symbol#to_s` method.
```
kotp, line 46:
```suggestion
To present a symbol as a string, you can use the `String#to_sym` method, and to do the opposite, you can use the `Symbol#to_s` method.
```

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
...
+Due to symbols having a limited set of methods, it can be useful to convert a symbol to a string to use string methods on it, if a modification is needed.
```
kotp, line 47:
```suggestion
Due to symbols having a limited set of methods, it can be useful to convert a symbol to a string to use string methods on it, if a new symbol is needed.
```

concepts/symbols/about.md
```diff
@@ -0,0 +1,80 @@
...
+:foo.to_s # => "foo"
+"foo".to_sym # => :foo
+```
+
+## Getting names in scopes
+
+In Ruby you can get names in scopes by using various methods to get names of constants, methods, and variables.
+These methods returns arrays of symbols.
+
+The most common methods are `Module#constants`, `Module#instance_methods`, `Module#class_variables`, `global_variables` and `local_variables`.
```
kotp, line 59:
```suggestion
Some methods are `Module#constants`, `Module#instance_methods`, `Module#class_variables`, `global_variables` and `local_variables`.
```

concepts/symbols/introduction.md
```diff
@@ -0,0 +1,55 @@
+# About
+
+[Symbols][symbols] are named identifiers that can be used to refer to a value.
+Symbols are created through a symbol literal, which is by prefixing a name with a `:` character, e.g. `:foo`.
+They also allow for being written as strings, e.g. `:"foo"`, which allows for example spaces in the name.
```
kotp, line 5:
```suggestion
They also allow for being written with quotes, e.g. `:"foo"`, which allows, for example, spaces in the name.
```

kotp (changes requested):

concepts/symbols/about.md
meatball133: "[Symbol](https://ruby-doc.org/3.2.2/Symbol.html) objects are different from [String](https://ruby-doc.org/3.2.2/String.html) objects in that [Symbol](https://ruby-doc.org/3.2.2/Symbol.html) objects represent identifiers, while [String](https://ruby-doc.org/3.2.2/String.html) objects represent text or data." - Ruby Docs, Symbols

I could likely change the wording a bit and say:

> What makes symbols different from strings is that they are identifiers, and do not represent data or text.

concepts/symbols/about.md
meatball133: I think conversion is the correct wording here since I think it is what a student links with getting another type.
Since every "conversion" when you go from one type to another, isn't actually a "conversion" in the sense that the "converted" value is the same as the original.

```irb
irb(main):001:0> a = 5
=> 5
irb(main):002:0> a.object_id
=> 11
irb(main):003:0> a.to_s.object_id
=> 26870
irb(main):004:0> a.object_id
=> 11
```

And these methods `to_s` and `to_sym` is put under the section of "Methods for Converting", in the official doc.

image

concepts/symbols/about.md
meatball133: I am inclining here that if you want to modify a symbol, you can convert it to a string, and when you have a string you can modify it. Does that need to be clarified?

concepts/symbols/.meta/config.json
meatball133: Feel free to suggest another blurb/modified one.

concepts/symbols/about.md
meatball133: I put them in quotation marks, to highlight that they are not really modified, and the later part of the sentence should also suggest the same. But we can change it to manipulate as you suggested

concepts/symbols/about.md
kotp: Yeah, you really can not modify a symbol.  So converting it to a string, modifying that string, and then coercing the string to a symbol to create a new symbol is what is happening.  It is a different object, as can be seen by using `.object_id` to check/verify.  No modification of a symbol is possible.

concepts/symbols/about.md
kotp: I like that better.  And yeah, the Ruby Docs is right, it does not say that Ruby has "text" only that String objects represent "text".

I would have worded it a little more strongly in the docs, as Symbol objects are identifiers, they do not represent identifiers.  They do represent the "non-symbol" names of those things identified though. You can see this in action when you do something like `Object.methods` and see the symbols that are the identifiers that represent the names of those methods.

concepts/symbols/about.md
kotp: I think "coerce" is the functionality of conversion from a symbol to something else, so yeah, but definitely not modify.  "conversion" is nuanced, since I can convert a hard top vehicle to a soft top vehicle, and it is the same vehicle, where here we would get a brand new object, instead of a true "conversion".

What you said about the conversion not really being a true conversion.  As long as we avoid the idea that modification is going on, I do think that "convert" and "conversion" is OK.  We should accentuate that it is a new object, though.  (Ruby is strongly typed, so having the same object change types would be weird.)

exercises/concept/port-palermo/.docs/introduction.md
```diff
@@ -0,0 +1,55 @@
+# About
+
+[Symbols][symbols] are named identifiers that can be used to refer to a value.
+Symbols are created through a symbol literal, which is by prefixing a name with a `:` character, e.g. `:foo`.
+They also allow for being written as strings, e.g. `:"foo"`, which allows for example spaces in the name.
```
kotp, line 5:
```suggestion
They also allow for being written with quotes, e.g. `:"a foo"`, which allows for example spaces in the name.
```

This also demonstrates the benefit of using strings with the symbol creation, allowing for a space in the symbol, reinforcing the example shown in the IRB session below.  (Well, it could be an IRB session, anyway...)

exercises/concept/port-palermo/.docs/introduction.md
```diff
@@ -0,0 +1,55 @@
...
+
+```ruby
+:foo # => :foo
+:"foo boo" # => :"foo boo"
+```
```
kotp, line 10:
~~~suggestion
```irb
>> :foo
=> :foo
>> :"foo boo"
=> :"foo boo"
```
~~~

concepts/symbols/.meta/config.json
kotp:
```suggestion
  "blurb": "Symbols are very important in Ruby.  They are used as "identity".  They are immutable, guaranteed to be a single item no matter where it is used or referenced.  They are used internally as variable, method, constants, and keyword identifiers.  They are also useful, and often used, as keys in a `Hash`."
```

Something like that.  Hopefully that helps set up the expectation that these things will be covered, or demonstrated, during the concept.

kotp (approved):

kotp:
Thank you @meatball133 !  Looking forward to seeing people go through the exercise!

PR #1562, opened 2023-08-18
Fix rubocop error for the constant by converting to uppercase - new

egemen-dev:
This PR fixes the rubocop issue of the [Constant Name](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/ConstantName) for the exercise concept of port-palermo. Since usage of uppercase constants are recommended by the rubocop docs and it is a good practice.

This PR is the new version of [the previous one](https://github.com/egemen-dev/ruby/tree/port_palermo_constant_fix). I mistakenly deleted the previous one with the forked repo so, I had to open a new one.

### DIFF
--- a/exercises/concept/port-palermo/.docs/instructions.md
+++ b/exercises/concept/port-palermo/.docs/instructions.md
@@ -12,10 +12,10 @@ The system has to handle identifiers for ships, but also for destinations.
 The first thing you need to do is to create the identifier for the port of Palermo.
 The identifier are the first four letters of the name of the port, in uppercase.
 
-Define the `Port.Identifier` constant to be a symbol with the value `:PALE`.
+Define the `Port::IDENTIFIER` constant to be a symbol with the value `:PALE`.
     
 ```ruby
-Port::Identifier 
+Port::IDENTIFIER 
 # => :PALE
 ```
 
--- a/exercises/concept/port-palermo/port_palermo.rb
+++ b/exercises/concept/port-palermo/port_palermo.rb
@@ -1,6 +1,4 @@
 module Port
-  # TODO: define the 'Identifier' constant
-
   def self.get_identifier(city)
     raise 'Please implement the Port.get_identifier method'
   end
--- a/exercises/concept/port-palermo/port_palermo.rb
+++ b/exercises/concept/port-palermo/port_palermo.rb
@@ -1,4 +1,6 @@
 module Port
+  # TODO: define the 'IDENTIFIER' constant
+
   def self.get_identifier(city)
     raise 'Please implement the Port.get_identifier method'
   end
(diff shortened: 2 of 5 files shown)


exercises/concept/port-palermo/.docs/instructions.md
```diff
@@ -12,10 +12,10 @@ The system has to handle identifiers for ships, but also for destinations.
 The first thing you need to do is to create the identifier for the port of Palermo.
 The identifier are the first four letters of the name of the port, in uppercase.
 
-Define the `Port.Identifier` constant to be a symbol with the value `:PALE`.
+Define the `Port::IDENTIFIER` constant to be a symbol with the value `:PALE`.
     
 ```ruby
-Port::Identifier 
+Port::IDENTIFIER
```
kotp, line 18: @meatball133 thoughts?  Should we not have it as a comment, but have it as explicit code instead?

exercises/concept/port-palermo/port_palermo.rb
```diff
@@ -1,5 +1,5 @@
 module Port
-  # TODO: define the 'Identifier' constant
+  # TODO: define the 'IDENTIFIER' constant
```
kotp, line 2: @meatball133 the comment earlier was specifically regarding this submitted version to the student.

kotp (commented):
Awaiting approval/comments/direction from @meatball133 .

exercises/concept/port-palermo/.docs/instructions.md
meatball133: Do you mean like in the stub file, that we give them the constant but let them define it by themself?

kotp: Yeah.  Because what is presented here is different than how they would likely define it, though by adding an equals sign and a value to this exact syntax, it could be assigned.  I am not sure how important it is to tell them to define the constant if we give them that line in the code... seems like "check the box" rather than "see how a constant is defined", compared to this instruction set and having them find out how to define a constant in that name space.

It might be more straight forward just to do it for them.  It is not a complex idea.

exercises/concept/port-palermo/port_palermo.rb
kotp: In response to [this](https://github.com/exercism/ruby/pull/1562#discussion_r1299237531) yes...

exercises/concept/port-palermo/.docs/instructions.md
meatball133: Yeah, I am fine either, the goal is to teach how to define a symbol. So in a way, there is no need to have the constant part, but that is some repetition on how to do it.

kotp (approved):

kotp:
Thank you @egemen-dev .

PR #1563, opened 2023-08-21
[Port Of Palermo] Remove redundant returns in exemplar

gabo-cs:
This PR removes a redundant 'return' statement

### DIFF
--- a/exercises/concept/port-palermo/.meta/exemplar.rb
+++ b/exercises/concept/port-palermo/.meta/exemplar.rb
@@ -2,7 +2,7 @@ module Port
   IDENTIFIER = :PALE
 
   def self.get_identifier(city)
-    return city[0..3].upcase.to_sym
+    city[0..3].upcase.to_sym
   end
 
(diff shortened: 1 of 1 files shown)


kotp:
I have reopened it. Normally we would want to talk about it in the forums first, but this is something that the exemplar solution should not do.  So we can bring it in.

Would you like to remove the other redundant `return` statement in the other method as well?  Might as well take care of both of them while we are here.

gabo-cs:
@kotp Thanks for that! It's done now! And sorry I "bypassed" the forum, didn't know until I saw the bot comment. Appreciate your help.

kotp:
> @kotp Thanks for that! It's done now! And sorry I "bypassed" the forum, didn't know until I saw the bot comment. Appreciate your help.

No problem. I wish there was a way to get the message out **before** that happens.  But it is what it is.

And we appreciate _your_ help!

kotp (approved):

PR #1569, opened 2023-09-13
[New Concept]: Range

meatball133:


concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+
+[Ranges][ranges] represents an interval between two values.
+The most common types that support ranges are `Int` and `String`.
+They can be used for many things like quickly creating a collection, slicing strings, checking if a value is in a range and iteration.
+They are created using the range operator `..` or `...` (inclusive and exclusive respectively).
+
+```ruby
+1..5 # => 1, 2, 3, 4, 5
+1...5 # => 1, 2, 3, 4
+```
```
kotp, line 11: The return is not what this statement returns, as we can see.  Using IRB, since IRB "evaluated" syntax is being used here in the code example:

~~~suggestion

In IRB:
```ruby
>> (1..5)
=> 1..5
>> (1...5)
=> 1...5
```
~~~

What do you think?  The return is not like it would be with an Array.

Probably not in this section, but we can coerce this to an Array to show what is originally here:

```irb
>> (1..5).to_a
=> [1, 2, 3, 4, 5]
>> (1...5).to_a
=> [1, 2, 3, 4]
```

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+
+The reason for having two range operators is to allow to create ranges that are inclusive or exclusive of the end value, which can be useful when for example working with indexes, that are zero based.
+
+Ranges can also be created using the `Range` initializer.
+
+```ruby
+Range.new(1, 5) # => 1, 2, 3, 4, 5
+```
```
kotp, line 19:
```suggestion
Ranges can also be created using the `Range` initializer.

Exploring in IRB:

```ruby
>> Range.new(1, 5)
=> 1..5
```
~~~

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+```ruby
+(1..5).sum # => 15
+1..5.sum # => Error: undefined method `sum' for 5:Integer (NoMethodError)
+```
+````
+
+## Getting substrings
+
+When wanting to slice a string, you can use the range operator to get a substring.
+That is by creating a range with the start and end index of the substring.
```
kotp, line 34:
```suggestion
That is, by creating a range with the start and end index of the sub-string.
```

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+## Range methods
+
+Ranges do have a set of methods that can be used to work with them.
+These methods when for example want to get the sum of all the values in the range, or check if the range includes a value.
+
+| Method                  | Description                                                             | Example                         |
+| ----------------------- | ----------------------------------------------------------------------- | ------------------------------- |
+| [`sum`][sum]            | Returns the sum of all the values in the range                          | `(1..5).sum # => 15`            |
+| [`size`][size]          | Returns the size of the range                                           | `(1..5).size # => 5`            |
+| [`include?`][indlude] | Returns `true` if the range includes the given value, otherwise `false` | `(1..5).include?(3) # => true` |
```
kotp, line 57:
```suggestion
| Method                  | Description                                                             | Example                         |
| ----------------------- | ----------------------------------------------------------------------- | ------------------------------- |
| [`sum`][sum]            | Returns the sum of all the values in the range                          | `(1..5).sum # => 15`            |
| [`size`][size]          | Returns the size of the range                                           | `(1..5).size # => 5`            |
| [`include?`][indlude]   | Returns `true` if the range includes the given value, otherwise `false` | `(1..5).include?(3) # => true` |
```

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Endless & Beginless ranges
+
+There are two special types of ranges, the endless and beginless ranges.
```
kotp, line 61: I wonder if they are really special ranges, or if this is more "syntactic sugar" that lets us express the normal ranges like this.

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+This means that the range has no beginning or end.
```
kotp, line 62: If we terminate them with negative infinity and positive infinity, then we can see that they do have a beginning or end.

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+The endless or beginless range has there start or end value being `nil`, but when defining the range so can nil be omitted.
```
kotp, line 63:
```irb
irb(main):004:0> Range.new(nil, 5)
ArgumentError: bad value for range
        from (irb):4:in `initialize'
        from (irb):4:in `new'
        from (irb):4
        from :0
irb(main):005:0> (nil..5)
ArgumentError: bad value for range
        from (irb):5
        from :0
irb(main):006:0> 
```

I do not think this is true, demonstrably.

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+
+Using beginless and endless ranges is useful when you want to for example slice a string from the beginning or to the end.
```
kotp, line 65:
```suggestion
Using beginless and endless ranges is useful when you want to, for example, slice a string from the beginning or to the end.
```

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+
+```ruby
+"Hello World"[0..] # => "Hello World"
+"Hello World"[4..] # => "o World"
+"Hello World"[..5] # => "Hello"
+```
+
+```exercism/caution
+If not used on a collection, the endless range can cause an infinite loop, if not used with caution.
```
kotp, line 74:
```suggestion
If not used on a collection, the endless range can cause an endless sequence, if not used with caution.
```

Because it does not loop, it is more an endless sequence than and endless loop.

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+## String ranges
+
+Strings can also be used in ranges and allow one to get an interval of strings between two strings.
+But its behavior is a bit different than with chars, when using multiple characters in a string range.
```
kotp, line 80:
```suggestion
But its behavior is a bit different than with `chars`, when using multiple characters in a string range.
```

I believe you may be referring to the method, and so that should be code formatted.

concepts/ranges/about.md
```diff
@@ -0,0 +1,125 @@
...
+Its behavior can become when doing more complex string ranges, so use it with caution.
+
+```ruby
+"aa".."az" # => "aa", "ab", "ac", ..., "az"
```
kotp, line 84:
```suggestion
In IRB:

>> "aa".."az"
=> "aa", "ab", "ac", # continues until  "az"
```

We could use a smaller sequence as well, and still demonstrate this, so we do not have to allude to the missing content.

kotp (changes requested):
That is about as far as I had time to go through, but please see the suggested changes, and there are some statements that I believe are not correct that should be changed as well.

concepts/ranges/about.md
meatball133: image

Looking in the docs are there also mention of nil: https://rubyapi.org/3.2/o/range.

kotp: Thanks.  That is why I said _think_... and did it in an REPL, to verify.  Seems like it has changed for newer versions of Ruby.

So I would then expect it to no longer be a terminated by infinity anything, and perhaps "undefined" in behavior.  I will definitely look at the source code.  Pretty sure the use will not change from `1..infinity` as `1..nil` though.  (That would be surprising, but `nil` evaluates as 0 if we do `to_i`  and as `0.0` if we do `nil.to_f` so definitely worth investigating.

kotp: Oh, you sent a picture too.  Whatever it is, I a sure it supports your statement.  Not at a good device at the moment to get interpretations for pictures.

concepts/ranges/about.md
meatball133: Yeah, I mean it could cause an infinte loop if you iterate over the sequence, but yes I think saying it is an endless sequence is better.

concepts/ranges/introduction.md
```diff
@@ -6,8 +6,11 @@ They can be used for many things like quickly creating a collection, slicing str
 They are created using the range operator `..` or `...` (inclusive and exclusive respectively).
 
 ```ruby
-1..5 # => 1, 2, 3, 4, 5
-1...5 # => 1, 2, 3, 4
+1..5  # => 1..5
+1...5 # => 1...5
+
+(1..5).to_a # => [1, 2, 3, 4, 5]
+(1...5).to_a # => [1, 2, 3, 4]
```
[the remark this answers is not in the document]
kotp: Oh, it also gives us an easy to verify side effect.  If we wanted to verify that the documentation is correct, it is easy to use `rubydoctest` to verify that it is still correct.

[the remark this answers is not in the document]
iHiD: That IRB syntax really isn't clear for me. It merging both the things you write/type and the things it's output in one format is very unclear to me. And I find it more unclear if we have multi-line things. e.g. if I want to do

```ruby
>> class Foobar
>>   def something
>>     5
>>   end
>> end
>> Foobar.new.something
=> 5
```

I just find seeing normal Ruby code without the `>>` prefix much much simpler to read:
```ruby
class Foobar
  def something
    5
  end
end

Foobar.new.something

# => 5
```

So to answer "Do we hate IRB in the documentation?", I can't speak for everyone, but I sure do!

I could maybe deal with the `irb>` syntax used in Rails docs, just on the lines that execute some actual code (see https://guides.rubyonrails.org/active_record_callbacks.html for examples) but I don't really like that either. I don't think its intuitive to someone who has no idea what irb is, and I don't think someone should need to know that to get through the track. I want people to be able to just type in what they see on the screen, not have to understand that `>>` has a special meaning and that they need to delete it if they copy + paste.

concepts/ranges/about.md
```diff
@@ -0,0 +1,127 @@
...
+(1..5).to_a # => [1, 2, 3, 4, 5]
+(1...5).to_a # => [1, 2, 3, 4]
+```
+
+The reason for having two range operators is to allow to create ranges that are inclusive or exclusive of the end value, which can be useful when for example working with indexes, that are zero based.
+
+Ranges can also be created using the `Range` initializer.
+
+```ruby
+Range.new(1, 5) # => 1, 2, 3, 4, 5
```
iHiD, line 21:
```suggestion
Range.new(1, 5) # A range containing 1, 2, 3, 4, 5
```

concepts/ranges/introduction.md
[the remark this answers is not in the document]
kotp: Fair enough.  But then what is the `# => 5` at the end of that block?  Not very descriptive.  At least the IRB is a built in tool, and given with any Ruby install.

I am sure we are both biased, given our experience with the language and the tooling, which is why I am asking.

In "school" I explain the IRB and RI tooling that is included with Ruby.  When we are in "Practice" mode, then I kind of expect that experience to be there.  When in "Learn" mode, then I would expect the syntax  that we used to have been explained at some point.

Also, that is not output, it is the evaluation, the return, of either the line or the method.  Only lines that are `print` and `puts` gives output, if we consider it is something that we would see or that would be written to a file.

This is an extract from the guides link you gave.  Even this uses the proper IRB output and return distinctions.

And to me this behaves exactly as IRB, which is not surprising, given it is indicating it is coming from an IRB run:

```ruby
irb> User.new
You have initialized an object!
=> #<User id: nil>

irb> User.first
You have found an object!
You have initialized an object!
=> #<User id: 1>
```

The output is output, the evaluation shown as it shows there.  Less "interpretation" for the student if they see this and then run the Rails console, (aka, IRB), they are very likely to see exactly the same thing.

The shown block:

```ruby
>> class Foobar
>>   def something
>>     5
>>   end
>> end
>> Foobar.new.something
=> 5
```

… will not likely show what you would be typing anyway, so it may be confusing as well.

---

I do agree with the copy/paste being something that they should be able to do where appropriate without having to wrangle the lines of text.  Absolutely.

But doesn't `# => ` have a "special meaning" as well, since we have never (that I have found) explained what that means?

concepts/ranges/about.md
kotp: Just for information, I happened to have a Pry session open, where Pry has a bug that does not show this as it is supposed to.  That was the reason it was broken as shown.  Not the syntax itself. "Bad Pry!"

concepts/ranges/about.md
```diff
@@ -0,0 +1,126 @@
...
+| Method                  | Description                                                             | Example                         |
+| ----------------------- | ----------------------------------------------------------------------- | ------------------------------- |
+| [`sum`][sum]            | Returns the sum of all the values in the range                          | `(1..5).sum # => 15`            |
+| [`size`][size]          | Returns the size of the range                                           | `(1..5).size # => 5`            |
+| [`include?`][indlude]   | Returns `true` if the range includes the given value, otherwise `false` | `(1..5).include?(3) # => true` |
+
+## Endless & Beginless ranges
+
+A range can be endless and beginless.
+The endless or beginless range has there start or end value being `nil`, but when defining the range so can nil be omitted.
```
kotp, line 65:
```suggestion
The endless or beginless range has theirr start or end value being `nil`, but when defining the range so can nil be omitted.
```

Or 

```suggestion
The endless or beginless range has its start or end value being `nil`, but when defining the range so can nil be omitted.
```

exercises/concept/chess-game/.docs/introduction.md
```diff
@@ -62,7 +62,7 @@ These methods when for example want to get the sum of all the values in the rang
 ## Endless & Beginless ranges
 
 A range can be endless and beginless.
-The endless or beginless range has there start or end value being `nil`, but when defining the range so can nil be omitted.
+The endless or beginless range has their start or end value being `nil`, but when defining the range so can nil be omitted.
```
kotp, line 65: This file looks very much out of place here.

meatball133: what do you mean?

kotp: This is coming from the `chess-game directory.  Looks out of place for a "New Concept: Range" patch.

meatball133: it is the range exercise

kotp: Right, I see.  So the range concept (as per the title) and using the Chess game to show it off.  OK.

concepts/ranges/introduction.md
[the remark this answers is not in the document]
iHiD: > But doesn't # =>  have a "special meaning" as well, since we have never (that I have found) explained what that means?

`#` is a comment
` => ` is a signifier for output/evaluation/etc (as per the examples you pasted above).

So what the blocks in this PR are saying is "given this (runnable via copy+paste) code, here's a comment that shows what the output will be."

As you say, we definitely all have our biases and familiarities for this. I don't at all claim one is right or wrong, but I would like us continuing to use the syntax that I used on the existing concept exercises (e.g. see the tasks on https://exercism.org/tracks/ruby/exercises/lasagna/)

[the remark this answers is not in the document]
iHiD: @kotp It's worth noting that this is how official Ruby docs are written:
e.g. see the examples here: https://ruby-doc.org/core-2.7.0/Array.html#class-Array-label-Creating+Arrays

kotp (approved):

PR #1585, opened 2023-10-17
Fix copy and pasted nil about.md

iHiD:


concepts/nil/about.md
```diff
@@ -1,65 +1,12 @@
...
-  def owner
-    @owner
-  end
-
-  def owner=(new_owner)
-    @owner = new_owner
-  end
-end
+# I do not have a favorite color
+favorite_color = nil
```
kotp, line 7: This could be better represented as "no  favorite color collected" while if they really have no favorite color and we have collected that information, "none" might be a better representation, rather than `nil`.

This is already merged, though, so just a note.

PR #1586, opened 2023-10-18
[New practice exercise] Gilded Rose

fpsvogel:
https://forum.exercism.org/t/add-the-gilded-rose-kata/7758

This version is a bit different from the classic Gilded Rose Kata. See https://github.com/exercism/ruby/commit/1094e5373d737a300b2609e024cb5e7441a37803 for the changes of substance _(EDIT: simplified, [see discussion below](https://github.com/exercism/ruby/pull/1586#discussion_r1364528806))_, and https://github.com/exercism/ruby/commit/6d034738d87437a2c279b2482946d5d4e2c5ab25 for the rationale. Feedback on this adjustment is welcome!

I've also edited the instructions from the original for clarity, conciseness, and emphasis on the task of refactoring.

~~### Note on the linter warnings~~
~~All of the warnings are about the starting code (which is _supposed_ to be horrible), except for one, an "Ambiguous negative number operator" warning on a line which (IMO) is unambiguous.~~

_EDIT: Added a RuboCop exclusion for the starting code, and adjusted the code re: the other warning._

PR #1594, opened 2023-10-27
Update config for Gilded Rose exercise

fpsvogel:
Adds to `config.json` for the new Gilded Rose exercise:

- Contributors list
  - Reviewers of #1586 plus BethanyG, who gave extensive feedback in the [forum thread].
- Minimum Ruby version of 3.1
  - The tests use omitted values in hash literals, a Ruby 3.1 feature.
- ~~In the root `config.json`, removes the prerequisites for Gilded Rose. This makes the exercise more easily accessible to Ruby devs who aren't already on Exercism. (If they came for Gilded Rose, did the "Hello world" exercise, and then saw that they still needed to other exercises first, they might give up.)~~
  - _EDIT: Removed, see blow._

I also wanted to specify an `icon`, but [the existing icons](https://exercism.org/docs/building/tracks/icons#h-exercise-icons) are specific to other exercises, and googling didn't turn up any suitable free icon with a permissive license. I'm not an artist, so I'll leave off the icon for someone else to contribute.

[forum thread]: https://forum.exercism.org/t/add-the-gilded-rose-kata/7758

### DIFF
--- a/config.json
+++ b/config.json
@@ -1603,7 +1603,9 @@
 
         ],
         "prerequisites": [
-
+          "strings",
+          "conditionals",
+          "enumeration"
         ],
         "difficulty": 8
       }
--- a/config.json
+++ b/config.json
@@ -1603,9 +1603,7 @@
 
         ],
         "prerequisites": [
-          "strings",
-          "conditionals",
-          "enumeration"
+
         ],
         "difficulty": 8
       }
--- a/exercises/practice/gilded-rose/.meta/config.json
+++ b/exercises/practice/gilded-rose/.meta/config.json
@@ -2,7 +2,12 @@
   "authors": [
     "fpsvogel"
   ],
-  "contributors": [],
+  "contributors": [
+    "BethanyG",
+    "ErikSchierboom",
+    "IsaacG",
+    "vaeng"
+  ],
   "files": {
     "solution": [
       "gilded_rose.rb"
--- a/exercises/practice/gilded-rose/.meta/config.json
+++ b/exercises/practice/gilded-rose/.meta/config.json
@@ -14,6 +14,7 @@
       ".meta/example.rb"
     ]
   },
+  "language_versions": ">=3.1.0",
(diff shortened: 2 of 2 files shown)


kotp (approved):
I think the removal of the prerequisites are fine.  We can see how that works out.

iHiD:
> In the root config.json, removes the prerequisites for Gilded Rose. This makes the exercise more easily accessible to Ruby devs who aren't already on Exercism. (If they came for Gilded Rose, did the "Hello world" exercise, and then saw that they still needed to other exercises first, they might give up.)

I don't love this change. The same argument could be made for pretty much any exercise. I don't think this is a special exception. If someone doesn't want to do the prereqs we have a UI-level option for that, which is to switch to Practice Mode. Instead, we'll get newbies to Ruby trying to do it as it gets unlocked straight away. So I think this change should be reverted.

kotp:
Do we need to revert this or just not accept it?  I think we may not need to revert it, it does not appear to have been accepted yet (I did not merge it only approved it pending other opinions.)

kotp:
Also, since this is active, we can force push without fed0cd5adebaa745d9460e7dc2bf847be7657894 being included.

iHiD:
Ah, I thought it was merged. Yeah, just leave that bit out please :)

config.json
```diff
@@ -1603,9 +1603,7 @@
 
         ],
         "prerequisites": [
-          "strings",
-          "conditionals",
-          "enumeration"
+
```
kotp, line 1606: As @iHiD commented, let's leave the prerequisites in.  The student has a way to get around having to do the exercises for the prerequisites, and it hopefully will avoid folks trying to refactor without having a good (enough) understanding of the prerequisites.

kotp (changes requested):

config.json
fpsvogel: Done!

kotp (approved):

PR #1595, opened 2023-10-30
[New concept + exercise]: Case

meatball133:
The goal of this concept is to dig deeper into case, and in a later pr remove `case` from conditionals and expand the conditionals concept with more info.


concepts/case/about.md
```diff
@@ -0,0 +1,129 @@
+# Case
+
+[Case][case] (often referred to as switch in other languages) is a form of control expression like if-else.
```
kotp, line 3:
```suggestion
[Case][case] (often referred to as switch in other languages) is a form of flow control expression like if-else.
```

concepts/case/about.md
```diff
@@ -0,0 +1,129 @@
...
+Case allows for chaining of multiple if-else-if statements and can be more readable and also allows for powerful constructs.
```
kotp, line 4:
```suggestion
Case allows for chaining of multiple `when` statements and can be more readable and also allows for powerful constructs.
```

Also, does it really allow for "more powerful constructs"?  Or are they just as powerful but different?

concepts/case/about.md
```diff
@@ -0,0 +1,129 @@
...
+
+A case is defined by the keyword `case` followed by an optional expression.
+Then for each case, the keyword `when` is used followed by an expression which is compared to the case expression.
+The `when` keyword should not be indented from the `case` keyword.
+After the `when` keyword is the code that should be executed if the case expression matches the when expression.
+Case allows for an optional `else` statement which is executed if no other case matches.
```
kotp, line 10:
```suggestion
Case allows for an optional `else` statement which is executed if no other `when` matches.
```

concepts/case/about.md
```diff
@@ -0,0 +1,129 @@
...
+
+The case expression is evaluated and then compared to each `when` expression.
+The expression is compared using the case subsumption operator (`===`).
```
kotp, line 13: In Ruby, we call that the "case equality operator".

concepts/case/about.md
```diff
@@ -0,0 +1,129 @@
...
+
+```ruby
+case 1
```
kotp, line 16:
```suggestion
case value
```

concepts/case/about.md
```diff
@@ -0,0 +1,129 @@
...
+
+# This is the same as:
+tmp = 1
+if 1 === tmp
+  "One"
+elsif 1 === tmp
+  "Two"
+else
+  "Other"
+end
```
kotp, line 34:
```suggestion
# This is the same as:
value = 1
if 1 === value
  "One"
elsif 1 === value
  "Two"
else
  "Other"
end
```

kotp: I would argue that the two code examples are not the same though, where 2 is explicitly matched, while the if statement will never get to the `else`.

concepts/case/introduction.md
```diff
@@ -0,0 +1,129 @@
...
+# This is the same as:
+tmp = 1
+if 1 === tmp
+  "One"
+elsif 1 === tmp
+  "Two"
+else
+  "Other"
+end
+```
```
kotp, line 35: Same issue here where the if statement is not the same as the case statement.

kotp (changes requested):
The "same" example behaves differently than the other example.

"Case equality" is used pretty consistently over the years for the `===` method as used in a case statement.

concepts/case/about.md
meatball133: It doesnt really say they are more powerful just that they allow for powerful constructs.
And it is suppose to be if-else-if.

concepts/case/about.md
meatball133: It is a type on line 30 if it is that you are referring to?

concepts/case/about.md
kotp: I can see what it says, if "it" is this copy.  But should it?

kotp: And, yes, while we can indeed use if-else-end along and even inside of case statements, the case statement does not itself use that syntax.

kotp: I would even say that it is allowed for, even without case.  So the case statement isn't what is allowing for chaining if-else-if statements.

config.json
```diff
@@ -129,6 +129,18 @@
...
+        "name": "BlackJack",
+        "uuid": "414ee9b2-c62d-41d5-aa1e-c5d82fb8562c",
+        "concepts": [
+          "case"
+        ],
+        "prerequisites": [
+          "ranges",
+          "modules"
+        ]
+      },
```
kotp, line 143:
```suggestion
      },
```

concepts/case/about.md
kotp: Oh, a "typographical" error, yeah, I believe so.  I know that it is a logic error as well, though, did not identify that it might be a typo.

concepts/case/about.md
kotp: I have seen Jorg (from stack overflow) use it in 2010 and someone else call it this in their blog in 2022, but this is only the third time I have seen this `===` called that.

concepts/case/introduction.md
```diff
@@ -0,0 +1,129 @@
...
+else
+  "Other"
+end
+# => "One"
+
+# This is the same as:
+tmp = 1
+if 1 === tmp
+  "One"
+elsif 1 === tmp
```
kotp: This has not yet been updated, I think.

kotp (approved):

concepts/case/about.md
kotp, line 4:
```suggestion
Case allows for chaining of multiple if-else-if statements and can be more readable while still providing flow control.
```

How about that?

kotp (approved):

PR #1597, opened 2023-11-07
[Refactor Concept]: conditionals 

meatball133:
The aim is to clean up the docs from mentions of integers, and also to expand on the concept.

### DIFF
--- a/concepts/conditionals/about.md
+++ b/concepts/conditionals/about.md
@@ -1,6 +1,6 @@
 # Conditionals
 
-Ruby has what is known as control expressions, these are used to control the way the program will run and they take a truthy or falsey value.
+Ruby has what is known as flow control expressions, these are used to control the way the program will run and they take a truthy or falsey value.
 There are operators that can be used to create truthy or falsey values, these are known as [comparison operators][comparison-operators].
 
 There are two main control expressions that are used to control which code will run and which will not.
@@ -47,7 +47,7 @@ It will return `1` if the left value is greater than the right value, `-1` if th
 
 ## If statement
 
-The [`if`][if] statement is used to check if a given condition is truthy or falsey.
+The [`if`][if] statement is used to check if a given condition is "truthy" or "falsey".
 If the condition is truthy then the code inside the if statement will run.
 An `if` statement ends with the `end` keyword.
 
@@ -97,8 +97,6 @@ end
 
 unless value < 2
   "1 is not greater than 2"
-else
-  "1 is greater than 2"
 end
 # => "1 is greater than 2"
 ```
--- a/concepts/conditionals/about.md
+++ b/concepts/conditionals/about.md
@@ -1,48 +1,140 @@
-# About
+# Conditionals
 
-An `if` statement can be used to conditionally execute code:
+Ruby has what is known as control expressions, these are used to control the way the program will run and they take a truthy or falsey value.
+There are operators that can be used to create truthy or falsey values, these are known as [comparison operators][comparison-operators].
+
+There are two main control expressions that are used to control which code will run and which will not.
+Also known as which given branch will run.
+
+Those two are: `if` and the `unless` expression.
+
+## Comparison operators
+
+[Comparison operators][comparison-operators] are used to compare values and return a `true` or `false` value.
+The following operators require two values to be compared of the same type.
+If the values are not of the same type then the compiler will throw an error.
+Here is a list of the operators and an example of when they give a `true` value:
+
+| Method | Description           | Example |
+| ------ | --------------------- | ------- |
+| <      | less than             | 5 < 4   |
+| <=     | less than or equal    | 4 <= 4  |
+| >      | greater than          | 3 > 1   |
+| >=     | greater than or equal | 2 >= 2  |
+
+The equal and not equal operators can be used to compare any type of value contrary to the operators already mentioned.
+The `==` operator is used to check if two values are equal, and that includes checking the type of the value.
+The `!=` works the same way but it will return `true` if the values are not equal and `false` if they are equal.
+Here is a list of the equal and not equal operators and an example of when they give a `true` value:
+
+| Method | Description  | Example |
+| ------ | ------------ | ------- |
+| ==     | equal        | 4 == 4  |
+| !=     | not equal    | 5 != 4  |
+
+## Combined comparison operator
+
+The combined comparison operator (sometimes called spaceship operator) is a special comparison operator.
+It is special in the sense that it doesn't return a truthy or falsey value but it returns a number.
+It is written as `<=>` and it is used to compare 2 values.
+It will return `1` if the left value is greater than the right value, `-1` if the left value is less than the right value, and `0` if the values are equal.
 
 ```ruby
-x = 5
-
-if x == 5
-  # Execute logic if x equals 5
-elsif x > 7
-  # Execute logic if x greater than 7
-else
-  # Execute logic in all other cases
-end
+1 <=> 2 # => -1
+2 <=> 2 # => 0
+3 <=> 2 # => 1
 ```
 
-Sometimes you want to execute a statement (or statements) if a condition is _not_ true, for situations like that, Ruby implements the `unless` keyword:
+## If statement
+
+The [`if`][if] statement is used to check if a given condition is truthy or falsey.
+If the condition is truthy then the code inside the if statement will run.
+An `if` statement ends with the `end` keyword.
 
 ```ruby
-x = 4
-unless x == 5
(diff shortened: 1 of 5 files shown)


concepts/conditionals/about.md
```diff
@@ -1,48 +1,140 @@
-# About
+# Conditionals
 
-An `if` statement can be used to conditionally execute code:
+Ruby has what is known as control expressions, these are used to control the way the program will run and they take a truthy or falsey value.
```
kotp, line 3:
```suggestion
Ruby has what is known as flow control expressions, these are used to control the way the program will run and they take a truthy or falsey value.
```

concepts/conditionals/about.md
```diff
@@ -1,48 +1,140 @@
...
+There are operators that can be used to create truthy or falsey values, these are known as [comparison operators][comparison-operators].
```
kotp, line 4: These operators are really methods.  They are known as methods, even though this resource calls them "operators".  Can we reference the Pickaxe book for this instead?  We call the "Methods", correctly, in the table that follows below, as well.

concepts/conditionals/about.md
```diff
@@ -1,48 +1,140 @@
...
+Here is a list of the operators and an example of when they give a `true` value:
+
+| Method | Description           | Example |
+| ------ | --------------------- | ------- |
+| <      | less than             | 5 < 4   |
+| <=     | less than or equal    | 4 <= 4  |
+| >      | greater than          | 3 > 1   |
+| >=     | greater than or equal | 2 >= 2  |
+
+The equal and not equal operators can be used to compare any type of value contrary to the operators already mentioned.
```
kotp, line 25:
```suggestion
The equal and not equal methods can be used to compare any type of value contrary to the operators already mentioned.
```

concepts/conditionals/about.md
```diff
@@ -1,48 +1,140 @@
...
 
-if x == 5
-  # Execute logic if x equals 5
-elsif x > 7
-  # Execute logic if x greater than 7
-else
-  # Execute logic in all other cases
+## If statement
+
+The [`if`][if] statement is used to check if a given condition is truthy or falsey.
```
kotp, line 50:
```suggestion
The [`if`][if] statement is used to check if a given condition is "truthy" or "falsey".
```

Quotes since this is not a "normal" expression, things being either truth or false, while we are introducing an idea of not strict Boolean evaluation.

Not a blocker, though, for sure.

concepts/conditionals/about.md
```diff
@@ -1,48 +1,140 @@
...
-  # Execute logic if x == 5
+  "1 is not equal to 1"
 end
+# => "1 is equal to 1"
+
+unless value < 2
+  "1 is not greater than 2"
+else
+  "1 is greater than 2"
+end
```
kotp, line 102: This is perhaps considered a poor example, as we should prefer to not use `else` with `unless`.  The indication is that the conditional statement is probably "backward" or "opposite of what is wanted".

concepts/conditionals/about.md
meatball133: In general I think operators are used in quite a few places and is often how they are referenced, sure they are methods (there are some expectations but this stays true for most cases) but in that case so does Ruby not have any operators since everything is just methods.
According to the C API docs for ruby: https://docs.ruby-lang.org/en/master/syntax/operators_rdoc.html
Are an operator defined such as (and says that there is such a thing as a operator): 
"In Ruby, operators such as +, are defined as methods on the class. [Literals](https://docs.ruby-lang.org/en/master/syntax/literals_rdoc.html) define their methods within the lower level, C language. [String](https://docs.ruby-lang.org/en/master/String.html) class, for example.

Ruby objects can define or overload their own implementation for most operators."

There are multiple resources calling them methods. And the Ruby API even calls them **conventional comparison operators**: "[Comparable](https://rubyapi.org/3.2/o/comparable) uses <=> to implement the conventional comparison operators (<, <=, ==, >=, and >) and the method between?."
Link: https://ruby-doc.org/3.2.2/Comparable.html

And according to the number concept are they already called comparisons operator and that doc did either Jeremy or Erik write: https://github.com/exercism/ruby/blob/main/concepts/numbers/about.md.

As for referencing the Pickaxe book so can we only reference publicly available material and that is a paid resource.

kotp: I do not think we really disagree here.  I just want to ensure that it is clear that these "operators" are not really anything special, they are methods.

> In general I think operators are used in quite a few places and is often how they are referenced, sure they are methods (there are some expectations but this stays true for most cases) but in that case so does Ruby not have any operators since everything is just methods. According to the C API docs for ruby: https://docs.ruby-lang.org/en/master/syntax/operators_rdoc.html Are an operator defined such as (and says that there is such a thing as a operator): "In Ruby, operators such as +, are defined as methods on the class. [Literals](https://docs.ruby-lang.org/en/master/syntax/literals_rdoc.html) define their methods within the lower level, C language. [String](https://docs.ruby-lang.org/en/master/String.html) class, for example.

And yet, even that document states and reinfoces "since operators are method implementations".

> 
> Ruby objects can define or overload their own implementation for most operators."
> 
> There are multiple resources calling them methods. And the Ruby API even calls them **conventional comparison operators**: 

https://github.com/ruby/ruby/blob/8d0eb872e8fd0cc1837318c4c242f4085df9812e/doc/syntax/operators.rdoc#L4 is probably what I would choose to reference before I would defer to the unofficial (though still very good) rubyapi.org.   Cory does a good job maintaining that site.

But coming from the official source, rather than even a recommended source from the official site.   (Even when the official language page recommended some of my published content by Matz, I would still defer to the official sources.)

> "[Comparable](https://rubyapi.org/3.2/o/comparable) uses <=> to implement the conventional comparison operators (<, <=, ==, >=, and >) and the method between?." Link: https://ruby-doc.org/3.2.2/Comparable.html

Yes, the word "operator" is used conveniently in the documentation, concerning `<=>` method.  I do not begrudge that.

> And according to the number concept are they already called comparisons operator and that doc did either Jeremy or Erik write: https://github.com/exercism/ruby/blob/main/concepts/numbers/about.md.
> 

Yes, I never said that there are no operators, just that we might want to mention that these "operators" are methods.  Since we reinforce that in the table later, as well.  (Which may reduce some confusion, since we say operator and then almost immediately list them as methods in the table.)

> As for referencing the Pickaxe book so can we only reference publicly available material and that is a paid resource.

The PIckaxe book is available online for free, while yes, it is also a paid resource, http://ruby-doc.com/docs/ProgrammingRuby/

Much of it is fairly evergreen, and these "operators" have been methods since even earlier than that book was written, and we can note where things have changed in newer versions of Ruby.  This is not one of those areas, though.

kotp (approved):
Approving, with or without reference changes.  The references are good, the detail of the operators being methods that can be modified will come out at some point, and do not necessarily need to come out here or in direct references. (In other words, not a hill to defend to the death, for sure.)  ;)  Good work on the refactoring so far, for sure!

concepts/conditionals/about.md
meatball133: I mean Ruby api is just [rubys "offical" api](https://docs.ruby-lang.org/en/3.2/), but just Class and Module Index and in my view a lot nicer user interface. And it is part of documentation on the official site so I say why not.
The other text "https://github.com/ruby/ruby/blob/8d0eb872e8fd0cc1837318c4c242f4085df9812e/doc/syntax/operators.rdoc#L4" comes from the ruby C API which is a doc intended for work on the core of Ruby or extensions of Ruby, not really intended for casual ruby developers but it does have a pretty decent explainer of what an operator is. So I am not against to include a link to the C API for explainer what an operator is.

kotp: The source code is what is used to create/generate the documentation.  It changes when you look at a version of the documentation, and can match it to the same for the repository.

What this means, if we link to the current tree link, rather than a specific blob, is that the link will be up to the minute up to date.  I did show the current at the time though, it may have already been updated.

It is also why I usually suggest generating the documentation locally, as then  you can be sure to be reading the documentation for the version you are using, rather than potential mis-matches when you go to a web resource for generated documentation there.

meatball133: > What this means, if we link to the current tree link, rather than a specific blob, is that the link will be up to the minute up to date. I did show the current at the time though, it may have already been updated.

It will link to the latest official release of Ruby so it doesn't include unstable/preview changes. But yes this could lead to mismatched version, idealy there would be some kind of system around this. But feels like it is better to have it lose, since at least I find it better to have the docs linked to the version used by tooling/test runner. But idealy we would have a script which could update the docs when we update the test runner.

PR #1601, opened 2023-11-11
Add space in chess-game's README

sid-707:
"Hello World"[..5] should evaluate to "Hello " to include the space at the end since the range is inclusive.


kotp (approved):

kotp:
Thank you for reporting, sorry about the reported spam issue (I did not mark it as such, it is probably auto-tagged for new participants there for linking, for better or worse).

PR #1605, opened 2023-11-18
Update port_palermo_test.rb

wgwz:


wgwz:
Why make this a barrier to entry for ALL changes? This is a totally obvious change, which should be accepted w/o issue. I'm not going to open a discussion for a small change like this.

kotp:
And I am not going to make you open a discussion for this.  I would not call this a "small change" though, pretty big mistake, really.

Thank you for finding it and making the change!

kotp (approved):

PR #1607, opened 2023-11-26
Update Symbol conversion text

kotp:
Associate the right object with the right call

I believe `String#to_sym` should be used to present a string as a symbol
not vice versa.


saltamay (approved):
👍

meatball133 (approved):

PR #1619, opened 2024-01-16
leap: sync

ErikSchierboom:
Sync the `leap` exercise with the latest data, as defined in https://github.com/exercism/problem-specifications/tree/main/exercises/leap.

### DIFF
--- a/exercises/practice/leap/.docs/instructions.md
+++ b/exercises/practice/leap/.docs/instructions.md
@@ -1,22 +1,3 @@
 # Instructions
 
-Given a year, report if it is a leap year.
-
-The tricky thing here is that a leap year in the Gregorian calendar occurs:
-
-```text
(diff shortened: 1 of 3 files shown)


kotp (approved):

kotp:
This should not re-run tests and so our magic comment should exist in the commit that avoids triggering retesting all student exercises, @ErikSchierboom

ErikSchierboom:
Well, that's not actually true though :) If you check [the docs](https://exercism.org/docs/building/tracks#h-maintenance), you'll find that:

> Solutions will not be retested if the merged commit either:
> - only touches .docs or .meta files, or other files that users don't interact with
> - or contains [no important files changed] in the commit body.

The former applies to this PR, so it's safe to merge.

iHiD:
Thanks for thinking of that though @kotp!

PR #1623, opened 2024-01-24
Add Reverse String

iHiD:
I wasn't sure on the best module name. Do we have a standard we use for Practice Exercises?


exercises/practice/reverse-string/.meta/solutions/reverse_string.rb
```diff
@@ -0,0 +1,5 @@
+module ReverseString
```
kotp, line 1: We might name it similar to other modules in Ruby, with "able" at the end.  But we do not have to.

kotp (approved):

ErikSchierboom (approved):

ErikSchierboom:
rubocop is complaining

exercises/practice/reverse-string/reverse_string.rb
```diff
@@ -0,0 +1,8 @@
+=begin
+Write your code for the 'Reverse String' exercise in this file. Make the tests in
+`reverse_string_test.rb` pass.
+
+To get started with TDD, see the `README.md` file in your
+`ruby/reverse-string` directory.
+=end
+
```
kotp, line 8:
```suggestion
=end
```

kotp:
There are only two practice exercises that look like this:

```
exercises/practice/reverse-string/.meta
├── config.json                        
├── solutions                          
│   └── reverse_string.rb              
└── tests.toml                         
```

The rest of them look like:

```
exercises/practice/accumulate/.meta
├── config.json                    
├── example.rb                     
└── tests.toml                     
```

iHiD:
@kotp Thanks. Fixed here and updated for the long-term fix here: https://github.com/exercism/ruby/pull/1628

Please merge if you're happy with this :)

exercises/practice/reverse-string/.meta/config.json
```diff
@@ -4,13 +4,13 @@
   ],
   "files": {
     "solution": [
-      "reverse_string.rb"
+      "reverser.rb"
```
kotp, line 7: Generally speaking, though it is not always the case, the file name has matched the slug of the exercise  (with the exception of the `-` being `_` when representing two words.

exercises/practice/reverse-string/.meta/example.rb
```diff
@@ -1,4 +1,4 @@
-module ReverseString
+module Reverser
```
kotp, line 1: I like it.

kotp (approved):

kotp:
> @kotp Thanks. Fixed here and updated for the long-term fix here: #1628
> 
> Please merge if you're happy with this :)

The file name change that is a bit more than a approve/like/merge.  It speaks to "convention" from the Exercism viewpoint.  So, I would like to confirm with you that the difference (though it does happen) in the naming style is also fine with @ErikSchierboom as well as you.

ErikSchierboom (approved):
Lgtm!

PR #1629, opened 2024-01-27
Add dnd_character exercise

mr-sigma:
Add `dnd_character` exercise based on 48in24 implementation status page: https://exercism.org/challenges/48in24/implementation_status

And on the forums: https://forum.exercism.org/t/add-dnd-character-exercise/9492

### DIFF
--- a/exercises/practice/dnd-character/.meta/example.rb
+++ b/exercises/practice/dnd-character/.meta/example.rb
@@ -1,0 +1,43 @@
+module Die
+  module_function
+
+  def roll(roll: 4, die: 6, remove_lowest: 1)
+    roll.times.map { rand(1..die) }.sort.pop(roll - remove_lowest).sum
+  end
+end
+
+module Modifiable
+  def modifier(ability_score)
+    ability_score / 2 - 5
+  end
+end
+
+class DndCharacter
+  extend Modifiable
+
+  BASE_HITPOINTS = 10
+
+  private_constant :BASE_HITPOINTS
+
+  private
+  def initialize
+    @strength     = Die.roll
+    @dexterity    = Die.roll
+    @constitution = Die.roll
+    @intelligence = Die.roll
+    @wisdom       = Die.roll
+    @charisma     = Die.roll
+
+    @hitpoints    = BASE_HITPOINTS + self.class.modifier(constitution)
+  end
+
+  public
+
+  attr_reader :strength,
+    :dexterity,
+    :constitution,
+    :intelligence,
+    :wisdom,
+    :charisma,
+    :hitpoints
+end
--- a/exercises/practice/dnd-character/dnd_character.rb
+++ b/exercises/practice/dnd-character/dnd_character.rb
@@ -1,0 +1,17 @@
+=begin
+Write your code for the 'D&D Character' exercise in this file. Make the tests in
+`dnd_character_test.rb` pass.
+
+To get started with TDD, see the `README.md` file in your
+`ruby/dnd-character` directory.
+=end
+
+class DndCharacter
+  def self.modifier
+    # Your code here
+  end
+
+  def initialize
+    # Your code here
+  end
+end
--- a/exercises/practice/dnd-character/dnd_character_test.rb
+++ b/exercises/practice/dnd-character/dnd_character_test.rb
@@ -87,7 +87,7 @@ class DndCharacterTest < Minitest::Test
 
       attributes.each do |attribute|
         assert_includes allowed_range, character.send(attribute),
-          informative_message % [attribute, allowed_range] # rubocop:disable Style/FormatString
+          informative_message % [attribute, allowed_range]
       end
 
       informative_message = %q(The character's %s must be %s)
--- a/exercises/practice/dnd-character/dnd_character_test.rb
+++ b/exercises/practice/dnd-character/dnd_character_test.rb
@@ -93,7 +93,7 @@ class DndCharacterTest < Minitest::Test
       informative_message = %q(The character's %s must be %s)
 
       assert_equal expected_hitpoints, character.hitpoints,
-        informative_message % ['hitpoints', expected_hitpoints] # rubocop:disable Style/FormatString
+        informative_message % ['hitpoints', expected_hitpoints]
     end
   end
 
--- a/exercises/practice/dnd-character/dnd_character_test.rb
+++ b/exercises/practice/dnd-character/dnd_character_test.rb
@@ -1,0 +1,119 @@
+require 'minitest/autorun'
+require_relative 'dnd_character'
+
+module Helpers
+  BASE_HITPOINTS = 10
+
+  def attributes
+    %i[strength dexterity constitution intelligence wisdom charisma]
+  end
+end
+
+class DndCharacterTest < Minitest::Test
+  include Helpers
+
+  def test_modifier_score_3
+    assert_equal(-4, DndCharacter.modifier(3))
+  end
+
+  def test_modifier_score_4
+    assert_equal(-3, DndCharacter.modifier(4))
+  end
+
+  def test_modifier_score_5
+    assert_equal(-3, DndCharacter.modifier(5))
+  end
+
+  def test_modifier_score_6
+    assert_equal(-2, DndCharacter.modifier(6))
+  end
+
+  def test_modifier_score_7
+    assert_equal(-2, DndCharacter.modifier(7))
+  end
+
+  def test_modifier_score_8
+    assert_equal(-1, DndCharacter.modifier(8))
+  end
+
+  def test_modifier_score_9
+    assert_equal(-1, DndCharacter.modifier(9))
+  end
+
+  def test_modifier_score_10
+    assert_equal 0, DndCharacter.modifier(10)
+  end
+
+  def test_modifier_score_11
+    assert_equal 0, DndCharacter.modifier(11)
+  end
+
+  def test_modifier_score_12
+    assert_equal 1, DndCharacter.modifier(12)
+  end
+
+  def test_modifier_score_13
+    assert_equal 1, DndCharacter.modifier(13)
+  end
+
+  def test_modifier_score_14
+    assert_equal 2, DndCharacter.modifier(14)
+  end
+
+  def test_modifier_score_15
+    assert_equal 2, DndCharacter.modifier(15)
+  end
+
+  def test_modifier_score_16
+    assert_equal 3, DndCharacter.modifier(16)
+  end
+
+  def test_modifier_score_17
+    assert_equal 3, DndCharacter.modifier(17)
+  end
+
+  def test_modifier_score_18
+    assert_equal 4, DndCharacter.modifier(18)
+  end
+
+  # rubocop:disable Style/FormatString, Style/RedundantPercentQ
+  def test_random_character_stats
+    100.times do
+      character = DndCharacter.new
+      allowed_range = (3..18)
+      expected_hitpoints = BASE_HITPOINTS +
+                           DndCharacter.modifier(character.constitution)
+      informative_message = %q(The character's %s must be within %s)
+
+      attributes.each do |attribute|
+        assert_includes allowed_range, character.send(attribute),
+          informative_message % [attribute, allowed_range] # rubocop:disable Style/FormatString
+      end
+
+      informative_message = %q(The character's %s must be %s)
+
+      assert_equal expected_hitpoints, character.hitpoints,
+        informative_message % ['hitpoints', expected_hitpoints] # rubocop:disable Style/FormatString
+    end
+  end
+
+  def test_stats_calculated_once
+    informative_message = <<~EXPLAIN
+      The character's %<attribute>s must not change if called more than once.
+      It was %<first>s, is now %<second>s.
+    EXPLAIN
+
+    100.times do
+      character = DndCharacter.new
+
+      (attributes << :hitpoints).each do |attribute|
+        first = character.send(attribute)
+        second = character.send(attribute)
+
+        assert_equal first, second,
+          informative_message % { attribute:, first:, second: }
+      end
+    end
+    # rubocop:enable Style/FormatString, Style/RedundantPercentQ
+  end
+end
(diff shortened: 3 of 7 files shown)


exercises/practice/dnd-character/.meta/example.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def initialize
+    @strength = roll_dice
+    @dexterity = roll_dice
+    @constitution = roll_dice
+    @intelligence = roll_dice
+    @wisdom = roll_dice
+    @charisma = roll_dice
+
+    @hitpoints = 10 + self.class.modifier(constitution)
```
kotp, line 12: This is known from the constitution and we can derive the adjustment from base 10 hitpoints.

We might want to test through the story of the character and not the story of "modifier" as a function in a language.

kotp:
I converted to draft as it is still in progress.

exercises/practice/dnd-character/.meta/example.rb
```diff
@@ -0,0 +1,53 @@
...
+  def roll(roll: 4, die: 6, remove_lowest: 1)
+    roll.times.map { rand(1..die) }.sort.pop(roll - remove_lowest).sum
+  end
+
+end
+
+module Modifiable
+
+  # This only calculates hit point modifier based on constitution
+  def modifier(constitution)
```
kotp, line 13: Let's ask for the "generic" thing.

exercises/practice/dnd-character/.meta/example.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def initialize
+    @strength     = Die.roll
+    @dexterity    = Die.roll
+    @constitution = Die.roll
+    @intelligence = Die.roll
+    @wisdom       = Die.roll
+    @charisma     = Die.roll
+
+    @hitpoints    = BASE_HITPOINTS + self.class.modifier(constitution)
```
kotp, line 36: And give, as it is here, the more specific thing.

exercises/practice/dnd-character/.meta/example.rb
```diff
@@ -0,0 +1,53 @@
...
+              :intelligence,
+              :wisdom,
+              :charisma,
+              :hitpoints
+
+end
+
+if $PROGRAM_NAME == __FILE__
+  puts DndCharacter.allocate.public_methods(false)
+end
```
kotp, line 53: We can remove the library guard once we are done exploring.  Or at least make it something that is more "example" based rather than "debugging" like.

exercises/practice/dnd-character/dnd_character_test.rb
```diff
@@ -0,0 +1,52 @@
+require 'minitest/autorun'
+require_relative 'dnd_character'
+
+module Helpers
+  BASE_HITPOINTS = 10
+  MODIFIER_TO_CONSTITUTION = {
```
kotp, line 6: Constitution modifier to affect hitpoints, so this could be something that may be a different name.

exercises/practice/dnd-character/dnd_character_test.rb
```diff
@@ -0,0 +1,52 @@
...
+end
+
+class DndCharacterTest < Minitest::Test
+  include Helpers
+
+  def test_hitpoints_correctly_calculated_from_constitution_modifier
+    character = DndCharacter.new
+    constitution_modifier = character.hitpoints - BASE_HITPOINTS
+
+    assert_includes MODIFIER_TO_CONSTITUTION[constitution_modifier], character.constitution
```
kotp, line 30: We can add a message to the user if it fails that indicates what the result should have been.

exercises/practice/dnd-character/dnd_character_test.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    assert_equal expected_hitpoints, character.hitpoints
+  end
+
+  def test_attributes_being_accessed_do_not_change
+    character = DndCharacter.new
+
+    (attributes << :hitpoints).each do |attribute|
+      assert_equal character.send(attribute), character.send(attribute)
+    end
```
kotp, line 50: How many times should we test this to have some assurance that they are stable/unchanging?

exercises/practice/dnd-character/dnd_character.rb
```diff
@@ -0,0 +1 @@
+
```
kotp, line 1: We are giving an empty file.  There is commonly a comment here that describes (for those that may have been away for a while) to write the tests in this file, and to read the Instructions or README if they are developing locally.

exercises/practice/dnd-character/.meta/example.rb
mr-sigma: Using `ability_score` as the new argument based of 5e naming conventions: https://roll20.net/compendium/dnd5e/Ability%20Scores#content

exercises/practice/dnd-character/.meta/example.rb
mr-sigma: Removed!

exercises/practice/dnd-character/dnd_character_test.rb
mr-sigma: Gave a more specific name: `CONSTITUTION_MODIFIER_TO_CONSTITUTION_MAP`

exercises/practice/dnd-character/dnd_character_test.rb
mr-sigma: As above, thank you for addressing this in the RFC!

exercises/practice/dnd-character/dnd_character.rb
mr-sigma: Added boilerplate language for the stub file!

exercises/practice/dnd-character/.meta/example.rb
mr-sigma: Using `ability_score` as the new argument based of 5e naming conventions: https://roll20.net/compendium/dnd5e/Ability%20Scores#content

exercises/practice/dnd-character/.meta/example.rb
mr-sigma: Wasn't sure what you were asking for here, the modifier method takes an ability score, and in this case we want to calculate it from constitution, so I don't believe this needs any naming changes based on the use case.

kotp: I may have been looking at some other version, you are correct.

kotp (approved):

exercises/practice/dnd-character/dnd_character_test.rb
kotp: Not a great name.  But naming is hard.  And the `_MAP` is a "type indicator" in the name, which is not common in Ruby.

Let me look at this and see if I can come up with a better name.

exercises/practice/dnd-character/dnd_character_test.rb
```diff
@@ -0,0 +1,56 @@
...
```
kotp, line 54: I do not have much confidence that this is doing well what the test says it is doing.

```ruby
    (attributes << :hitpoints).each do |attribute|
      expected = character.send(attribute)
      10.times do
        assert_equal expected, character.send(attribute)
      end
```
This would check each 10 (arbitrary choice) times, against what it was the answer for the first call

exercises/practice/dnd-character/dnd_character_test.rb
```diff
@@ -0,0 +1,56 @@
+require 'minitest/autorun'
+require_relative 'dnd_character'
+
+module Helpers
+  BASE_HITPOINTS = 10
+  CONSTITUTION_MODIFIER_TO_CONSTITUTION_MAP = {
```
kotp, line 6: This table seems to be `ABILITY_SCORE_TO_MODIFIER`.

kotp (changes requested):

exercises/practice/dnd-character/dnd_character.rb
```diff
@@ -0,0 +1,6 @@
+=begin
+Write your code for the 'D&D Character' exercise in this file. Make the tests in
+`dnd_character_test.rb` pass.
```
kotp, line 3:
```suggestion

`dnd_character_test.rb` pass.
```

exercises/practice/dnd-character/dnd_character.rb
```diff
@@ -0,0 +1,6 @@
...
+To get started with TDD, see the `README.md` file in your
```
kotp, line 4:
```suggestion
`dnd_character_test.rb` pass.

To get started with TDD, see the `README.md` file in your
```

iHiD:
@mr-sigma @kotp Unless I'm missing something, the tests.toml and the tests themselves don't seem to be aligned. Compare this to C# for example: https://github.com/exercism/csharp/blob/main/exercises/practice/dnd-character/DndCharacterTests.cs

kotp:
> @mr-sigma @kotp Unless I'm missing something, the tests.toml and the tests themselves don't seem to be aligned. Compare this to C# for example: https://github.com/exercism/csharp/blob/main/exercises/practice/dnd-character/DndCharacterTests.cs

The tests are not one to one in alignment, this is true.  And CSharp having an exact one to one based on Problem Specification makes sense to me, since it is being generated, and it appears without any additional changes or removals from the Problem Specifications supplied information.

We should find that the tests do not need to be exact matches one to one, and may even be difficult to match, given the random nature of character generation, against a user written design, without likely driving a specific design or solution (and this is a practice exercise currently.)

The tests focus on the modification algorithm being applied, and the tests that attempt to ensure that the character attributes are changing randomly per character being generated, but also unchanging for the character once it is created.

We should find that the tests align in result, though, since deviation from that would be problematic if and when we decide to automate the tests, for example, and when bringing in problem descriptions from upstream.

Also, I think this pull request should still be in "Draft Mode" so we are still working through some things regarding the test and how it influences the solution.

iHiD:
My understanding is that tests should align 1-1 with the tests.toml. 

I've looked through many other tracks in case I misunderstood this, and they are all 1-1 to problem specs:

- https://github.com/exercism/javascript/blob/main/exercises/practice/dnd-character/dnd-character.spec.js
- https://github.com/exercism/bash/blob/main/exercises/practice/dnd-character/dnd_character.bats
- https://github.com/exercism/elixir/blob/main/exercises/practice/dnd-character/test/dnd_character_test.exs
- https://github.com/exercism/wren/blob/main/exercises/practice/dnd-character/dnd-character.spec.wren

I'll wait for @ErikSchierboom to weigh in on this on Tuesday, but my understanding of exercises Exercism-wide is that they should 1-1 align to tests.toml.

kotp:
> My understanding is that tests should align 1-1 with the tests.toml.
> 
> I've looked through many other tracks in case I misunderstood this, and they are all 1-1 to problem specs:
> 
> * https://github.com/exercism/javascript/blob/main/exercises/practice/dnd-character/dnd-character.spec.js
> * https://github.com/exercism/bash/blob/main/exercises/practice/dnd-character/dnd_character.bats
> * https://github.com/exercism/elixir/blob/main/exercises/practice/dnd-character/test/dnd_character_test.exs
> * https://github.com/exercism/wren/blob/main/exercises/practice/dnd-character/dnd-character.spec.wren
> 
> I'll wait for @ErikSchierboom to weigh in on this on Tuesday, but my understanding of exercises Exercism-wide is that they should 1-1 align to tests.toml.

I will defer.  When I was creating the generator, there were exercises that did not necessarily match up with Ruby, and I had made exceptions, but that was years ago, so it could have changed.  I understand that many are like that, I just was not aware of the requirement to do so other than it being convenient to do so.  (Some of the generators would conditional do things bringing in some tests, and making others or not do others.)

exercises/practice/dnd-character/dnd_character_test.rb
```diff
@@ -0,0 +1,102 @@
...
+    100.times do
+      character = DndCharacter.new
+      allowed_range = (3..18)
+      expected_hitpoints = BASE_HITPOINTS + DndCharacter.modifier(character.constitution)
+
+      attributes.each do |attribute|
+        assert_includes allowed_range, character.send(attribute)
+      end
+
+      assert_equal expected_hitpoints, character.hitpoints
```
kotp, line 89: This one is in the correct order, the expected first, and then character hitpoints being actual.

Indeed, we could remove the `expected_` from the name, or we could say only `expected` because the actual shows what kind of thing is expected based on the method being called on the object under test.

mr-sigma:
@kotp I added your patches, rebased, and ran `rubocop -d -A` to align with the repo's style guidelines. Should be all set to go, thanks for the feedback!

kotp (approved):
I sent a patch, but am approving early.

I will accept your choice on the what you see there and what you choose to apply.

This is definitely looking good.

mr-sigma:
@kotp, patch applied!

exercises/practice/dnd-character/dnd_character_test.rb
```diff
@@ -0,0 +1,119 @@
...
+    100.times do
+      character = DndCharacter.new
+      allowed_range = (3..18)
+      expected_hitpoints = BASE_HITPOINTS +
+                           DndCharacter.modifier(character.constitution)
+      informative_message = %q(The character's %s must be within %s)
+
+      attributes.each do |attribute|
+        assert_includes allowed_range, character.send(attribute),
+          informative_message % [attribute, allowed_range] # rubocop:disable Style/FormatString
```
kotp, line 90:
```suggestion
          informative_message % [attribute, allowed_range]
```

exercises/practice/dnd-character/dnd_character_test.rb
```diff
@@ -0,0 +1,119 @@
...
+      end
+
+      informative_message = %q(The character's %s must be %s)
+
+      assert_equal expected_hitpoints, character.hitpoints,
+        informative_message % ['hitpoints', expected_hitpoints] # rubocop:disable Style/FormatString
```
kotp, line 96:
```suggestion
        informative_message % ['hitpoints', expected_hitpoints]
```

kotp (changes requested):
Remove redundant inline disables.

kotp (approved):

kotp:
Thank you very much @mr-sigma

PR #1635, opened 2024-02-13
Update Bundler to version 2.5.6

smaboshe:
This commit updates Bundler to version 2.5.6 which is now available.

Ref:
- https://github.com/rubygems/rubygems/releases/tag/bundler-v2.5.6

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -51,4 +51,4 @@ DEPENDENCIES
   simplecov
 
 BUNDLED WITH
-   2.2.22
+   2.5.6


kotp (approved):

kotp:
Any reason to not go to 3.5.6 instead?

smaboshe:
> Any reason to not go to 3.5.6 instead? Sorry, specify the gem system version 3.5.6? And from the Gemfile not the Gemfile.lock?

No reason that I can think of 🙂. Let me try this locally.

smaboshe:
This turned out to be a thrilling side quest!

I learned that requiring `gem`, `gems` or `rubygems` inside `Gemfile` does not keep Rubygems updated. For example requiring `rubygems` and running `bundle install` resulted in:
```
Could not find gem 'rubygems' in rubygems repository https://rubygems.org/ or installed locally.
The source does not contain any versions of 'rubygems'
```

`gem update --system` is the recommended way to keep the Rubygems version up to date.

I also learned that requiring a particular version of Bundler in `Gemfile` can lead to conflicts.

My take away is that `bundle --update bundler` is the blessed way to keep Bundler up to date and its configuration is tracked in `Gemfile.lock` (not `Gemfile`).

That said, if there is a way to specify the gem system version in `Gemfile`, I'm excited to learn how.

iHiD:
It's also worth noting that whatever we do here, we should really match in the [Ruby Test Runner](https://github.com/exercism/ruby-test-runner). The dockerfile for that is here: https://github.com/exercism/ruby-test-runner/blob/main/Dockerfile

My experience of installing bundler in dockerfiles is one that has a similar story to your sidequest.

We recently discovered that it's important to pass pass the bundler version in as a param in the GH Action. Here's a PR to guide if you go down this route (from a different repo): https://github.com/exercism/website/pull/6751/files

smaboshe:
Thank you, @iHiD. Happy to take a closer look.

smaboshe:
From what I am learning, it is no longer necessary to specify the exact version of Bundler in the Ruby Test Runner. If I can find a link to the change that introduced this, I'll share it.

When we use the current Bundler (`gem install bundler` without specifying the version) it will detect the appropriate version install and then use that. I confirmed this by building the Ruby Test Runner `Dockerfile` locally and got the following:
```
Bundler 2.5.6 is running, but your lockfile was generated with 2.4.18. Installing Bundler 2.4.18 and restarting using that version.
```

Ref:
- https://github.com/exercism/ruby-test-runner/blob/e00080cc06e81ca24f6732a8405030b75f0eca04/Dockerfile#L8

smaboshe:
I have opened https://github.com/exercism/ruby-test-runner/pull/139 as a possible approach to keeping Bundler "in sync" on the Ruby Test Runner.

iHiD:
> it will detect the appropriate version install and then use that

Yeah, the problem with that is that we then lose the ability to cache Docker layers as all the gems will be re-installed, which takes a while for things like nokogiri. For the website, that adds 30mins to the deployment time.

The best way we've found to get round that is to read the bundler version from the Gemfile.lock and use that to cache properly (as per the PR I sent)

kotp:
We might be able to have a bin/bundle script that uses `bundle _2.2.2_ install` for example, reflecting the current Gemfile.lock version.  That will help to ensure that the version of Bundler is used (prerequisite, that version is installed, with `gem install bundler 2.2.2` (again as reflected by what is shown as Bundled with 2.2.2 in `Gemfile.lock`).

PR #1636, opened 2024-02-13
Add spiral matrix exercise

mr-sigma:
Forum thread: https://forum.exercism.org/t/add-spiral-matrix-exercise/9795

Ran into some issues with Rubocop's ABC complexity for generating the matrix, as well as it being unhappy with using `attr_reader` in some cases so would be interested in feedback for the example solution, but don't need it to be a blocker for getting the code in!

### DIFF
--- a/exercises/practice/spiral-matrix/.docs/instructions.md
+++ b/exercises/practice/spiral-matrix/.docs/instructions.md
@@ -1,0 +1,24 @@
+# Instructions
+
+Given the size, return a square matrix of numbers in spiral order.
+
+The matrix should be filled with natural numbers, starting from 1 in the top-left corner, increasing in an inward, clockwise spiral order, like these examples:
+
+## Examples
+
+### Spiral matrix of size 3
+
+```text
+1 2 3
+8 9 4
+7 6 5
+```
+
+### Spiral matrix of size 4
+
+```text
+ 1  2  3 4
+12 13 14 5
+11 16 15 6
+10  9  8 7
+```
--- a/exercises/practice/spiral-matrix/.docs/instructions.md
+++ b/exercises/practice/spiral-matrix/.docs/instructions.md
@@ -2,7 +2,7 @@
 
 Given the size, return a square matrix of numbers in spiral order.
 
-The matrix should be filled with natural numbers, starting from 1 in the top-left corner, increasing in an inward, clockwise spiral order, like these examples:
+The matrix should be filled with natural numbers, starting from 1 in the top-left corner, increasing in an inward, clockwise spiral order.
 
 ## Examples
 
--- a/exercises/practice/spiral-matrix/.meta/example.rb
+++ b/exercises/practice/spiral-matrix/.meta/example.rb
@@ -1,0 +1,48 @@
+class SpiralMatrix
+  private
+  attr_reader :counter, :dx, :dy, :x, :y
+
+  def initialize_matrix
+    Array.new(size) { Array.new(size, 0) }
+  end
+
+  def generate_matrix
+    while matrix_includes_zeroes?
+      @matrix[y][x] = counter
+
+      @dy, @dx = @dx, -@dy if next_step_invalid?
+
+      @x += dx
+      @y += dy
+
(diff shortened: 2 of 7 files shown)


exercises/practice/spiral-matrix/spiral_matrix.rb
```diff
@@ -0,0 +1,17 @@
...
+`spiraL_matrix_test.rb` pass.
+
+To get started with TDD, see the `README.md` file in your
+`ruby/spiral-matrix` directory.
+=end
+
+class SpiralMatrix
+  def initialize
+    # Your code here
+  end
```
iHiD, line 12: I have a preference not to have methods defined in stubs with the wrong signatures (e.g. here initialize needs to take a size). And equally, I'd rather not force the implementation by having a param defined, so I think my preference is just to remove this method and the the one below. 

But please tell me if there's a specific reason you felt they would be good to include.

mr-sigma: This was how I'd been asked to implement it in the Dnd Character exercise MR, was copying here, but happy to change whatever!

iHiD: Yeah, I think you should remove the code and just leave the comment at the top please. That's how all the other practice exercises are implemented (e.g. https://github.com/exercism/ruby/blob/main/exercises/practice/binary-search/binary_search.rb :)

kotp: I think it was a focus on providing a comment so that we did not submit an empty file.  Leave the practice exercises as mostly unguided.

exercises/practice/spiral-matrix/spiral_matrix.rb
```diff
@@ -0,0 +1,7 @@
+=begin
+Write your code for the 'Spiral Matrix' exercise in this file. Make the tests in
+`spiraL_matrix_test.rb` pass.
+
+To get started with TDD, see the `README.md` file in your
+`ruby/spiral-matrix` directory.
+=end
```
kotp, line 7: This is good.

exercises/practice/spiral-matrix/.docs/instructions.md
```diff
@@ -0,0 +1,24 @@
+# Instructions
+
+Given the size, return a square matrix of numbers in spiral order.
+
+The matrix should be filled with natural numbers, starting from 1 in the top-left corner, increasing in an inward, clockwise spiral order, like these examples:
```
kotp, line 5:
```suggestion
The matrix should be filled with natural numbers, starting from 1 in the top-left corner, increasing in an inward, clockwise spiral order.
```

Because a colon and then header level (section change) does not quite make sense to me.

kotp (approved):
Looks good pending one change concerning the colon use before a heading.

exercises/practice/spiral-matrix/.docs/instructions.md
mr-sigma: @kotp this is the auto-generated text from `bin/configlet create --practice-exercise spiral-matrix`, should this be changed upstream in this case?

exercises/practice/spiral-matrix/.meta/example.rb
```diff
@@ -0,0 +1,48 @@
...
+      @counter += 1
+    end
+  end
+
+  def next_step_invalid?
+    @y + @dy == @size ||
+      (@y + @dy).negative? ||
+      @x + @dx == @size ||
+      (@x + @dx).negative? ||
+      !(@matrix[@y + @dy][@x + @dx]).zero?
```
kotp, line 27:
```suggestion
      (@matrix[@y + @dy][@x + @dx]).positive?
```

exercises/practice/spiral-matrix/.docs/instructions.md
kotp: It probably does.

kotp:
Looking good!  Thank you @mr-sigma

PR #1640, opened 2024-03-05
Update parallel_letter_frequency_test.rb

iHiD:
Two changes:
1. I think the `texts` should always be an array. It's a string in one case.
2. There is punctuation used in the later tests that does not occur in the punctuation one.


kotp:
This is one commit for two changes.

I like the the second. The first, I think, is tested in such a way that we probably want to not need to re-test any that are submitted.  (not that there are many yet.)

kotp (approved):

iHiD:
Thanks! We'll need to rerun things as the working test runner only got deployed this morning, so I'll merge and do that now.

PR #1644, opened 2024-03-15
Add Knapsack exercise

fpsvogel:
Forum thread: https://forum.exercism.org/t/add-knapsack-exercise-48in24/10320


kotp:
On Thu, Mar 14, 2024 at 10:21:35PM -0400, Felipe Vogel wrote:
> ---
>  exercises/practice/knapsack/knapsack_test.rb | 97 ++++++++++++++++++++
>  1 file changed, 97 insertions(+)
> 
> diff --git a/exercises/practice/knapsack/knapsack_test.rb b/exercises/practice/knapsack/knapsack_test.rb
> index e69de29bb2..1a84e0e49f 100644
> --- a/exercises/practice/knapsack/knapsack_test.rb
> +++ b/exercises/practice/knapsack/knapsack_test.rb
> @@ -0,0 +1,97 @@
> +require 'minitest/autorun'
> +require_relative 'knapsack'
> +
> +class KnapsackTest < Minitest::Test
> +  Item = Data.define(:weight, :value)

Loving the use of `Data` here.

> +
> +  def test_no_items
> +    # skip
> +    max_weight = 100
> +    items = []

Consider making the assertion line have some kind of potential hint.

In the case of the exercse "all your base" on Ruby, we do something like
this:

```ruby
assert_equal expected, converted, hint
```

In this case, we can do:

```ruby
def test_no_items
  maximum_weight = 100
  items = []
  expected = 0
  actual = Knapsack.new(maximum_weight).max_value(items)

  assert_equal expected, actual, 'For no items, the resulting value must be 0'
```

This can help layout the reading and provides a reason or a hint as to
what you must do to make that test pass.

> +    assert_equal 0, Knapsack.new(max_weight).max_value(items)
> +  end
> +
---- ✂  Snip Snip  ✂  ----

> +      Item.new(weight: 118, value: 229),
> +      Item.new(weight: 120, value: 240)
> +    ]
> +    assert_equal 1458, Knapsack.new(max_weight).max_value(items)
> +  end
> +end

exercises/practice/knapsack/.docs/hints.md
```diff
@@ -0,0 +1,34 @@
...
+>
+> So the main idea is basically to split the entire knapsack problem into smaller knapsack problems. Instead of testing 10 items with capacity 50, you first try (after the trivial case of 0) 1 item and capacity 10, 20, 30, 40 and 50 (or how ever many sub tasks you want to create) and then take another item and start again at capacity 10.
+>
+> If you see item 1 fits into capacity 20+ then all these slots in the table now contain this value. Then you look at item 2 from capacity 10-50 again. Lets assume item 2 fits into capacity 20 as well. Then now you check whether it is a new maximum or not and if it is then you update the table. Now you look at capacity 30 for item 2, you see that item 2 fits, this means 10 capacity would remain if you take it, however there, as of now, was no item that fits into 10 capacity, thus the solution remains the same as before. At 40 this changes, you now realize that even if you include item 2 there are 20 capacity remaining, thus you can fill that space with the previous solution, which was item 1. Thus for 40 capacity, as of now, the optimal solution is to take item 1 and 2. And so on.
```
kotp, line 29: We probably should have these as single line sentences, or "sentence per line".

Even though it is a quote and quotes are unlikely to be modified/fixed, as it is history we are stating.  Still, the style is indicated.

kotp:
About the failing test, it is due to the `Data` use in the test, which I commented on, as something I do like.  But we should only use it if it is defined.  This will require either a version check in the test with an alternative solution for that structure and the desired behavior, or a choice not to use `Data` in the test at all.

kotp:
> ~~This will require either a version check in the test with an alternative solution for that structure and the desired behavior, or a choice not to use `Data` in the test at all.~~

Nah, just a conscious decision to bring it in even with the 3.1 version failing, for a known reason.  This will pass on the platform, as it is using Ruby 3.3 according to the repository.

fpsvogel:
@kotp Thanks for the feedback! I've made the changes that you suggested. Plus, I added punctuation and fixed typos in the explanation quoted from Reddit, in order to make it more readable.

exercises/practice/knapsack/knapsack_test.rb
```diff
@@ -0,0 +1,127 @@
...
+
+  def test_no_items
+    # skip
+    max_weight = 100
+    items = []
+    expected = 0
+    actual = Knapsack.new(max_weight).max_value(items)
+
+    assert_equal expected, actual,
+      "When there are no items, the resulting value should be 0."
```
kotp, line 15: It should be or it must be?  Is this a suggestion or a requirement?

kotp (commented):
Looking good, but a few changes are probably warranted in communicating what must be rather than what should be.

kotp (approved):

exercises/practice/knapsack/knapsack_test.rb
kotp: Yeah, they "shoulda" probably not done that. ;)

kotp: Doesn't "shoulda" sound like a tool for testing warnings and deprecation, rather for things that must happen otherwise you get failures?  But naming is hard, and the name _is_ kind of cute, definitely memorable.

fpsvogel: True. Too bad the gem name "muster" is already taken 😄

PR #1647, opened 2024-03-20
Add Zebra Puzzle exercise

fpsvogel:
Forum thread: https://forum.exercism.org/t/add-zebra-puzzle-exercise-48in24/10368

Note that I've disabled a RuboCop rule in part of the example solution. LMK if that's something we should avoid.


fpsvogel:
Sorry if the force push was badly timed! I didn't realize the PR was reopened already.

I didn't actually make any changes from when I submitted the PR. I only cleaned up the commit history to simplify it.

kotp (approved):

PR #1650, opened 2024-03-26
Add easier hint and example solution for Knapsack exercise

fpsvogel:
After recently implementing Knapsack (#1644), I realized that there's a much simpler brute-force approach that is still performant with the small input sets that are in the tests.

This PR:

- Adds hints on this brute-force approach as a starting point.
- Replaces the example solution, just to add variety to the example solutions across different tracks. (The existing example solution is based directly on the C++ version.)

### DIFF
--- a/exercises/practice/knapsack/.docs/hints.md
+++ b/exercises/practice/knapsack/.docs/hints.md
@@ -2,13 +2,14 @@
 
 ## General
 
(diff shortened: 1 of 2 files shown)


kotp (approved):

kotp:
[no important files changed]

PR #1654, opened 2024-04-05
Try to fix CI

ErikSchierboom:


ErikSchierboom:
@kotp I've upgraded a bunch of stuff, but now I'm getting a different error. Would you mind looking into it?

kotp:
Spent quite some time researching.

Without any success so far for a fix.

But pushing to see if the CI passes with this change...

kotp:
The last commit does fix the knapsack, I think, but fails for where Data is being used for the zebra exercise.

ErikSchierboom:
> The last commit does fix the knapsack, I think, but fails for where Data is being used for the zebra exercise.

Any chance you could fix that too?

kotp:
> > The last commit does fix the knapsack, I think, but fails for where Data is being used for the zebra exercise.
> 
> Any chance you could fix that too?

Ruby 3.0 has just transitioned to End of Life, Ruby 3.1 will be end of life in the next 11 months.  Perhaps we make this "reportive" rather than required?

kotp:
Thank you @ErikSchierboom !

ErikSchierboom:
You too!

PR #1662, opened 2024-04-21
Changes for upcoming default frozen strings

kotp:
The changes remove the magic comment for frozen strings, but also makes
mutable strings were necessary to avoid unnecessary test failures for
students.

### DIFF
--- a/exercises/concept/bird-count/bird_count_test.rb
+++ b/exercises/concept/bird-count/bird_count_test.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 require 'minitest/autorun'
 require_relative 'bird_count'
 
--- a/exercises/concept/lasagna/lasagna_test.rb
+++ b/exercises/concept/lasagna/lasagna_test.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 require 'minitest/autorun'
 require_relative 'lasagna'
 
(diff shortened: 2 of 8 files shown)


meatball133:
Generally, I am for this but the direction of Ruby as it currently seems to be moving towards frozen string literals, as can be seen by the planned changes in 3.4: https://github.com/ruby/ruby/blob/master/NEWS.md, which states:

> String literals in files without a frozen_string_literal comment now behave as if they were frozen. If they are mutated a deprecation warning is emitted. These warnings can be enabled with -W:deprecated or by setting Warning[:deprecated] = true. To disable this change, you can run Ruby with the --disable-frozen-string-literal command line argument. [[Feature #20205](https://bugs.ruby-lang.org/issues/20205)]

kotp:
> Generally, I am for this but the direction of Ruby as it currently seems to be moving towards frozen string literals, as can be seen by the planned changes in 3.4: https://github.com/ruby/ruby/blob/master/NEWS.md, which states:
> 
> > String literals in files without a frozen_string_literal comment now behave as if they were frozen. If they are mutated a deprecation warning is emitted. These warnings can be enabled with -W:deprecated or by setting Warning[:deprecated] = true. To disable this change, you can run Ruby with the --disable-frozen-string-literal command line argument. [[Feature #20205](https://bugs.ruby-lang.org/issues/20205)]

Absolutely and that is why I made the changes.

This is moving toward the default frozen strings, and all of the warnings are no longer there when checking for areas where frozen strings from us will cause an exception.

This is why I have done the `+'''` where I have to signal those are not immutable strings, while still being string literals, as an example.

meatball133 (approved):

PR #1688, opened 2024-06-12
Include example for unless statement

wascoding:
Includes an else example in the code making it more clear.


iHiD:
Hello. I think generally unless/else is considered an antipattern. Generally speaking if/else is a better choice in my experience, so I'm not sure we want to teach that here. I'd suggest opening a post on the forum (forum.exercism.org) if you'd like to discuss further. Thanks! :)

kotp:
Discussion happening [here](https://forum.exercism.org/t/include-example-for-unless-statement/11667).

PR #1713, opened 2024-08-21
fix introduction.md in 'assembly-line' exercise

pontabi:
Please see the forum
http://forum.exercism.org/t/misinfo-in-ruby-concept-exercise/12633/2


kotp (approved):
Approving, since, while the statement is not false, the resulting change does indicate how it _should_ be done, without giving an opinion about it definitively.  In other words, leaving out the fact that you _can_ also avoids encouraging doing so.

I like it.

PR #1719, opened 2024-09-30
Match ruby version with test runner

meatball133:
The test runner uses 3.3.0 so makes no sense the track uses 3.2


iHiD:
Seems sensible to me :)

iHiD (approved):

kotp (commented):
So should we update the test runner to version 3.3.2 or reduce the version here?  It seems that we are going backward, and eventually we are likely going to be reverting this change.

kotp (approved):
My eyes misread, when I ran the audio, doh!  Yeah, definitely approved.

meatball133:
I wrote my message wrong I meant to say 3.3

PR #1722, opened 2024-10-03
Sync palindrome products exercise with problem specification

themetar:
[As discussed on the [forum](https://forum.exercism.org/t/why-is-the-palindrome-products-exercise-not-in-sync-with-problem-spec/13149)]

* Added the missing test cases
* Updated the example solution to comply to new tests


exercises/practice/palindrome-products/.meta/example.rb
```diff
@@ -1,43 +1,40 @@
...
 
 class Palindromes
-  attr_reader :range
-  def initialize(options)
-    max = options.fetch(:max_factor)
-    min = options.fetch(:min_factor) { 1 }
-    @range = (min..max)
-  end
+  private
+  attr_reader :min, :max
```
kotp, line 5:
```suggestion
  private
  
  attr_reader :min, :max
```

exercises/practice/palindrome-products/.meta/example.rb
```diff
@@ -1,43 +1,40 @@
...
+    enum.lazy.
+      filter { |number| palindrome? number }.
+      map { |number| Palindrome.new number, factors(number) }.
+      find { |palindrome| !palindrome.factors.empty? }
   end
 
-  def smallest
-    sort.first[1]
+  public
+  attr_reader :smallest, :largest
```
kotp, line 33:
```suggestion
  public
  
  attr_reader :smallest, :largest
```

kotp (changes requested):
You should be able to apply the suggestions, they are simply to provide a blank line after visibility methods `private` and `public` as they do not affect only the following line.

exercises/practice/palindrome-products/.meta/example.rb
themetar: vscode kept triggering rubocop and auto-formatting my code. I assumed that that's exercism's style.

I think this is the rule that's being followed:
https://github.com/exercism/ruby/blob/fa35ccc41372394db18b4c4a140e27b2597d38d6/.rubocop.yml#L38

themetar:
I've added the blank lines. @kotp

kotp (approved):

PR #1728, opened 2024-11-01
Fix typo ruby don't exist null in flatten-array

yuler:


kotp (commented):
No need, that is what the slash is for in this document, which comes from upstream.

PR #1730, opened 2024-11-09
Impliment new template based generator

meatball133:


.github/workflows/generator-tests.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-22.04
+    container:
+      image: crystallang/crystal
+    steps:
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@52753b7da854d5c07df37391a986c76ab4615999
+        with:
+          ruby-version: "3.3"
+          bundler-cache: true
+        run: rake test:generator
```
kotp, line 32:
```suggestion
        run: rake test:generator

```

generatorv2/lib/generator.rb
```diff
@@ -0,0 +1,102 @@
...
+  end
+
+  def toml(path = "./exercises/practice/#{@exercise}/.meta/tests.toml")
+    raise "Toml not found: #{path}" unless File.exist?(path)
+
+    uuid = TomlRB.load_file(path)
+    uuid.filter do |_k, v|
+      v.none? { |k, inner_value| k == "include" && !inner_value }
+    end
+    uuid.map { |k, _v| k }
```
kotp, line 61:
```suggestion
    uuid.keys
```

No tested.

generatorv2/lib/generator.rb
```diff
@@ -0,0 +1,102 @@
...
+  def underscore(str)
+    str.each_char.reduce("") do |acc, x|
+      acc << if [' ', '-'].include?(x)
+               '_'
+             else
+               x.downcase
+             end
+      acc
+    end
+  end
```
kotp, line 40:
```suggestion
def underscore(str)
  str.gsub(/[-\s]/, '_').downcase
end

```

bin/generate.rb
```diff
@@ -0,0 +1,41 @@
+require 'optparse'
```
kotp, line 1:
```suggestion
#!/usr/bin/env ruby
require 'optparse'
```

We would also likely drop the file name extension, and would set the executable bit.

generatorv2/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+Last Updated: 2024/11/9
+
+The generator is a powerful tool that can be used to generate tests for exercises based on the canonical data.
+The generator is written in Ruby and is located in the `bin` directory.
+
+## How to use the generator
+
+### Things to do before running the generator
+
```
kotp, line 11:
```suggestion

Run `bundle install` to install the required libraries.
```

bin/generate
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env ruby!
```
kotp, line 1:
```suggestion
#!/usr/bin/env ruby
```

generatorv2/lib/generator.rb
```diff
@@ -0,0 +1,49 @@
+require 'toml-rb'
```
kotp, line 1: What problems or solutions are we avoiding/gaining by using TOML as opposed to YAML (or even JSON)?  This is the point at which TOML as a tool and dependency is being introduced, and its associated maintenance cost, so we should evaluate that.  We should have a net positive effect by bringing this in.

bin/generate
```diff
@@ -0,0 +1,42 @@
...
+
+parser.on('--verify', 'Verify all exercises') do
+  exercises = Dir.entries('./exercises/practice').select { |f| File.directory? File.join('./exercises/practice', f) }
+  exercises.each do |exercise|
+    if File.exist?("./exercises/practice/#{exercise}/.meta/test_template.erb")
+      current_code = File.read("./exercises/practice/#{exercise}/#{exercise}_test.rb")
+      f = Tempfile.create
+      Generator.new(exercise).generate(f.path)
+      generated_code = f.read
+      raise RuntimeError.new("The result generated for: #{exercise}, doesnt match the current file") if current_code != generated_code
```
kotp, line 33:
```suggestion
      raise RuntimeError.new("The result generated for: #{exercise}, doesn't match the current file") unless current_code == generated_code
```

Typographical error (missing apostrophe) and preference of "positive conditional statement".

Prefer using `unless positive_conditional_statement` rather than `if negative`, this can keep all (or at least most) of our conditional statements positive.


We should also consider if `RuntimeError` is the best error here.  It could be `VerificationError` instead, which would be more specific.

We might also `fail` instead of `raise` for the communication that this is purposefully failed here, given the right conditions.

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+Before running the generator you have to make sure a couple of files are in place.
+
+1. `tests.toml` file
+
+It is located under the `.meta` folder for each exercise.
+The toml file is used to configure which exercises are generated and which are not.
+Since the generator grabs all the data from the canonical data, so does this enable new tests that won't automatically be merged in.
+Instead so does new tests have to be added to the toml file before they show up in the test file.
+
+If there is a test that isn't needed or something that doesn't fit Ruby you can remove it from the toml file.
```
kotp, line 22:
```suggestion
If there is a test that isn't needed or something that doesn't fit Ruby you can remove it from the configuration file.
```

In case we move from TOML to another type of configuration file, this reference will not need to change in the documentation.

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The generator makes sure that the exercise is in the config.json so you need to add it there before running the generator.
+
+#### Things to note
+
+The script which grabs info from the toml file is quite sensitive, writing the toml file in an incorrect way can brick the generator.
+
+Here are some examples of how you should **NOT** work with the toml file.
+
+Make sure that the uuid is the only thing inside of `[uuid]`, if there is, for example, an extra space so would that break it.
```
kotp, line 35: Isn't "uuid" supposed to be capitalized as an acronym?

```suggestion
Make sure that the UUID is the only thing inside of `[uuid]`, if there is, for example, an extra space so would that break it.
```

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+The template is located under the `.meta` for each exercise.
+
+This template has to be manually written for each exercise.
+The goal although is to make it so that you only have to write the template once and then it will be able to be used to generate new tests.
+
+The template file is written in [Embedded Ruby(ERB)][erb].
+ERB enables you to write Ruby code inside of the template file.
+It also means that the templates can be highly customizable since you can write any Ruby code you want.
+
+When writing the template file it is recommended to look at already existing template files to get a better understanding of how it works.
```
kotp, line 81:
```suggestion
When writing the template file, it is recommended to look at already existing template files to get a better understanding of how it works.
```

generatorv2/lib/generator.rb
meatball133: The `test.toml` files are written in toml, using another file format isn't really an option since rewriting it would require parsing it. 

I wrote a manual parser for the Crystal generator (but it is likely not as good of an implementation) because I couldn't find any good shards (library). This library should be fine, the toml specification isn't changing, but we could have our own solution.

meatball133: The goal is more or less with the toml files is that, if someone adds extra test cases or reimplements test cases for a certain exercise will that not affect the generation process until you sync the exercise. Meaning the ci won't fail in the majority of cases if not all exercises are synced.

bin/generate
```diff
@@ -23,15 +37,16 @@ parser.on('-a', '--all', 'Generate all exercises') do
...
     if File.exist?("./exercises/practice/#{exercise}/.meta/test_template.erb")
       current_code = File.read("./exercises/practice/#{exercise}/#{exercise}_test.rb")
       f = Tempfile.create
       Generator.new(exercise).generate(f.path)
       generated_code = f.read
-      raise RuntimeError.new("The result generated for: #{exercise}, doesnt match the current file") if current_code != generated_code
+      raise VerificationError unless current_code == generated_code
     end
+    rescue VerificationError => e
+      $stderr.puts e.message % {exercise:}
```
kotp, line 50: ℹ️ Be aware that this syntax is relatively new, and so we should ensure that the minimum Ruby version is set to allow for this "valueless hash" syntax.

meatball133: I mean it is tooling and we have set the ruby version to 3.3 so I think we dont have to have backwards compatibility.

kotp: It means that it is backward compatible to 3.3.0, up to 3.3.6, currently.  Locally I am running 3.3.6, and so the backward compatibility is there.

bin/generate
```diff
@@ -47,7 +47,7 @@ parser.on('--verify', 'Verify all exercises') do
       fail VerificationError unless current_code == generated_code
     end
     rescue VerificationError => e
-      $stderr.puts e.message % {exercise:}
+      stderr.puts e.message % {exercise:}
```
kotp, line 50:
```suggestion
      STDERR.puts e.message % {exercise:}
```

My mistake, not sure that this is under test, and so I missed the case of this constant.

kotp (changes requested):

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+Make sure that the UUID is the only thing inside of `[UUID]`, if there is, for example, an extra space so would that break it.
+Here is an example
+
+```toml
+# This would break it since it is an extra space between UUID and `]`
+[1e22cceb-c5e4-4562-9afe-aef07ad1eaf4 ]
+# This would break it since it is an extra space between UUID and `[`
+[ 1e22cceb-c5e4-4562-9afe-aef07ad1eaf4]
+```
+
```
kotp, line 44: It may be worth writing a "validator" option for the genator.  Perhaps even initiating the `bundle install` command that is needed.  If we know it has to happen we can provide the mechanism, avoiding mistakes from the users, even if it is seemingly "simple".

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+```
+# skip
+skip
+skip
+```
+
+### The Test Generator
+
+If all the earlier steps are done so can you run the generator.
```
kotp, line 107:
```suggestion
If all the earlier steps are done you run the generator.
```

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The generator uses a template file to generate the test file.
+The template is located under the `.meta` for each exercise.
+
+This template has to be manually written for each exercise.
+The goal although is to make it so that you only have to write the template once and then it will be able to be used to generate new tests.
+
+The template file is written in [Embedded Ruby(ERB)][erb].
+ERB enables you to write Ruby code inside of the template file.
+It also means that the templates can be highly customizable since you can write any Ruby code you want.
```
kotp, line 79:
```suggestion
It also means that the templates can be highly customized allowing you to write any Ruby code you want or need.
```

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+include = false
+```
+
+### Template
+
+The generator uses a template file to generate the test file.
+The template is located under the `.meta` for each exercise.
+
+This template has to be manually written for each exercise.
+The goal although is to make it so that you only have to write the template once and then it will be able to be used to generate new tests.
```
kotp, line 75:
```suggestion
The goal is to make it so that you only have to write the template once and then it will be able to be used to generate new tests.
```

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The template file is written in [Embedded Ruby(ERB)][erb].
+ERB enables you to write Ruby code inside of the template file.
+It also means that the templates can be highly customizable since you can write any Ruby code you want.
+
+When writing the template file, it is recommended to look at already existing template files to get a better understanding of how it works.
+The template is getting a slightly modified version of the canonical data, so you can check out the [canonical data][canonical data] to see the data structure.
+The modification is that the cases which are not included in the toml file will be removed from the data structure.
+
+When writing the template so is it a special tool that can help with giving `# skip` and `skip` tags for tests.
```
kotp, line 85: What about (for this line and the next:

> When writing the template, there is a method called `status` that will write the required skip line, where the first test will have the skip commented out.

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+You simply have to call the `status` method.
+It will return either `# skip` or `skip` depending on if it is the first test case or not.
+
+Here is an example:
+
+```
+<%= status()%>
+<%= status()%>
+<%= status()%>
```
kotp, line 94: Should this be rewritten so that we can call it as `<%= skip? %>`?

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+```
+# skip
+skip
+skip
+```
+
+### The Test Generator
+
+If all the earlier steps are done so can you run the generator.
+To run the generator you need to have a working Ruby installation and installed all gems in the Gemfile.
```
kotp, line 108:
```suggestion
To run the generator you need to have a working Ruby installation with the gems installed, via `bundle install`.
```

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+The generator is located in the `bin` directory and is called `generator.rb`.
```
kotp, line 109: We should consider renaming the executable without the `.rb` extension.

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+To run the generator so do you have to be in the root directory and run the following command:
+
+```shell
+bundle exec ./bin/generate -e <exercise_slug>
```
kotp, line 114: Either the line commented prior to this comment is incorrect, or this line is incorrect.

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+Where `<exercise_slug>` is the same name as the slug name which is located in the `config.json` file.
+
+For more commands so can you run the following command:
```
kotp, line 119:
```suggestion
For more commands and options, you can see this by running the command:
```

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+```shell
+bundle exec ./bin/generate --help
+```
+
+### Errors and warnings
+
+The generator will give you errors and warnings if something is wrong.
+That includes if the exercise is not in the `config.json` file, if the exercise is not in the toml file, or if the template file is missing.
+Also if it has a problem getting the `canonical-data.json` file so will it give you an error.
```
kotp, line 129:
```suggestion
It will also report an error if it can not read the `canonical-data.json` file.
```

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+The generator also uses a formatter which will give you errors if the generated file is not formatted correctly.
```
kotp, line 130:
```suggestion
The generator will check that the generated file is formatted correctly, reporting an error if there is a problem.
```

generatorv2/README.md
```diff
@@ -0,0 +1,134 @@
...
+The file will still be generated even if formatter gives errors, therefore can you check the file and see what is wrong and fix it in the template.
```
kotp, line 131:
```suggestion
The file will still be generated even if the formatter reports errors, So that you can check the file and see what is wrong and fix it in the template.
```

kotp (changes requested):

generatorv2/README.md
meatball133: I think, that shouldn't actually be an issue, unless it breaks against toml rules which I think it doesn't. The docs very some what of a copy paste from the Crystal docs. And that one has that issue,

generatorv2/test/utils_test.rb
```diff
@@ -22,15 +22,15 @@ def test_underscore_with_two_words
       Generator.new("two-fer").underscore("two-fer")
   end
 
-  def test_status
+  def test_skip
```
kotp, line 25: Should the name be `skip_test` as that is what it does, rather than it testing a skip?

meatball133:
yes, that I think could have be a good option. Been hit by a heatwave and generally been quite busy so sorry for not responding for a while

PR #1735, opened 2024-12-02
Updated description about nil

carmageddon888:


kotp (commented):
I like the change.  It is definitely more accurate.

kotp (approved):

kotp:
Normally and probably even for these, it is better to discuss them in the forms.  But I have reopened, and am applying this shortly.

PR #1738, opened 2025-01-12
Update chess_game.rb

Horlamedhey:
The method 'nick_name' is not consistent with the instructions' version (nickname), and since 'nickname' is a whole word, it makes sense without the underscore.


kotp:
Whenever you think it is ready for review, go ahead and click the "Ready for review" button to get this out of "Draft".

PR #1739, opened 2025-01-15
exercises/practice/anagram: add an append with instructions about the order of the return values

IsaacG:


kotp (approved):
We could have also relaxed the requirements.  But this is good.

IsaacG:
Clarify what this track expects. Context: https://forum.exercism.org/t/14816/

IsaacG:
@kotp Do you have merge permissions? I don't.

PR #1742, opened 2025-03-10
Update unless / else example in conditionals concept

clauddzy:
Tried to clear up the example. Text now matches condition and added the else case.

Discussion is located [here].

[here]: https://forum.exercism.org/t/confusing-unless-else-example-in-conditionals-concept/16293/7

### DIFF
--- a/concepts/conditionals/about.md
+++ b/concepts/conditionals/about.md
@@ -17,7 +17,7 @@ Here is a list of the operators and an example of when they give a `true` value:
 
 | Method | Description           | Example |
 | ------ | --------------------- | ------- |
-| <      | less than             | 5 < 4   |
+| <      | less than             | 4 < 5   |
 | <=     | less than or equal    | 4 <= 4  |
 | >      | greater than          | 3 > 1   |
 | >=     | greater than or equal | 2 >= 2  |
@@ -66,7 +66,7 @@ end
 
 ## Unless statement
 
-The `unless`unless statement works very similarly to the `if` statement but it will run the code inside the `unless` statement if the condition is falsey.
+The `unless` statement works very similarly to the `if` statement but it will run the code inside the `unless` statement if the condition is falsey.
 
 ```ruby
 value = 1
@@ -108,7 +108,7 @@ end
 The `elsif` statement can be used in conjunction with the if statement.
 The `elsif` statement will be executed if the if branch is not executed and the condition of the elsif statement is truthy.
 Elsif statements can be chained together and the first truthy condition will be executed.
-There can also be an else statement at the end of the if statement which will run if non of the earlier statement has been true.
+There can also be an else statement at the end of the if statement which will run if none of the earlier statements have been true.
 
 ```ruby
 value = 1
--- a/concepts/conditionals/about.md
+++ b/concepts/conditionals/about.md
@@ -112,14 +112,14 @@ There can also be an else statement at the end of the if statement which will ru
 
 ```ruby
 value = 1
-if value != 1
-  "1 is not equal to 1"
+if value == 0
+  "1 is equal to 0"
 elsif value > 2
   "1 is greater than 2"
 else
-  "1 is not equal to 1 and 1 is not greater than 2"
+  "1 is not equal to 0 and 1 is not greater than 2"
 end
-# => "1 is not equal to 1 and 1 is not greater than 2"
+# => "1 is not equal to 0 and 1 is not greater than 2"
 ```
 
 ## if and unless as suffix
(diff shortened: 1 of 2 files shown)


concepts/conditionals/about.md
```diff
@@ -96,9 +96,11 @@ end
 # => "1 is equal to 1"
 
 unless value < 2
-  "1 is not greater than 2"
+  "1 is not less than 2"
+else
+  "1 is less than 2"
```
kotp, line 101: Because there is an `else` here it strongly suggests that this should be an `if` statement instead, where we swap the conditional statements to be correct.  Avoiding `else` with `unless` is very idiomatic, as it prevents mental gymnastics that otherwise happens.

concepts/conditionals/about.md
```diff
@@ -96,9 +96,11 @@ end
...
 end
-# => "1 is greater than 2"
+# => "1 is less than 2"
```
kotp, line 103: This may be the only change needed.  @meatball133 ??

concepts/conditionals/introduction.md
```diff
@@ -83,9 +83,11 @@ end
 # => "1 is equal to 1"
 
 unless value < 2
-  "1 is not greater than 2"
+  "1 is not less than 2"
+else
+  "1 is less than 2"
```
kotp, line 88: See the other file where I comment on this.

concepts/conditionals/introduction.md
```diff
@@ -83,9 +83,11 @@ end
...
 end
-# => "1 is greater than 2"
+# => "1 is less than 2"
```
kotp, line 90: This may be the warranted change, in my opinion, as stated earlier.

kotp: As Ellnix expressed, this is more than just a comment it is an expression of evaluation.  This will not end up existing in this form per the discussion on the forums, fortunately.

kotp:
Reopened assuming  acceptance of decided "work to be done" in the discusson.

clauddzy:
I removed references to `unless` in the **Else statement** paragraph.
I replaced the `unless` example with an `if` example where the `else` branch is executed. That way there are examples for both cases.
I also fixed some other typos I found in the files.

concepts/conditionals/about.md
```diff
@@ -17,7 +17,7 @@ Here is a list of the operators and an example of when they give a `true` value:
 
 | Method | Description           | Example |
 | ------ | --------------------- | ------- |
-| <      | less than             | 5 < 4   |
+| <      | less than             | 4 < 5   |
```
kotp, line 20: Good eyes.

concepts/conditionals/about.md
```diff
@@ -95,18 +95,20 @@ else
...
+# => "1 is not greater than 2"
 ```
 
 ## "Cascading-if" statements
 
 The `elsif` statement can be used in conjunction with the if statement.
 The `elsif` statement will be executed if the if branch is not executed and the condition of the elsif statement is truthy.
 Elsif statements can be chained together and the first truthy condition will be executed.
-There can also be an else statement at the end of the if statement which will run if non of the earlier statement has been true.
+There can also be an else statement at the end of the if statement which will run if none of the earlier statements have been true.
```
kotp, line 111: Good here as well.

concepts/conditionals/introduction.md
```diff
@@ -82,18 +82,20 @@ else
...
 
 ## "Cascading-if" statements
 
 The `elsif` statement can be used in conjunction with the if statement.
 The `elsif` statement will be executed if the if branch is not executed and the condition of the elsif statement is truthy.
 Elsif statements can be chained together and the first truthy condition will be executed.
-There can also be an else statement at the end of the if statement which will run if non of the earlier statement has been true.
+There can also be an else statement at the end of the if statement which will run if none of the earlier statements have been true.
 
 ```ruby
```
kotp, line 100: Can you also attend to this block of code?

The value is one, but the evaluated response states that it is not equal to 1, though it is.

clauddzy: I see what you mean. I will make the changes

kotp: Thank you.

clauddzy: The resulting expression is now mathematically correct.
Another way to go about it would be to simply remove the not in the first part of the condition condition, resulting: `"1 is equal to 1 and 1 is not greater than 2"`. Personally I think the logic in the example I proposed is a bit simpler and lets a beginner focus on the if / elsif statements evaluation, which is the main point of the section.
But I'm fine with either way.

kotp (approved):

kotp:
Thank you for the work @clauddzy!

clauddzy:
My pleasure! Yes, this is the first one.

Issue #1532, opened 2023-04-27
Add analyzer for leap


kytrinyx:
We have some logic in the leap test suite that ensures that people don't use the built-in functions from `Date`.

Since the instructions don't say anything about not using built in functions (see https://github.com/exercism/ruby/blob/main/exercises/practice/leap/.docs/instructions.md), we should create an analyzer for this functionality.

kotp:
Alternative middle step may be the add the mention in the documentation, as an append.

I would be as delighted, though, to see a refinement on the Date class, as much as a "not Date class" solution.  I believe an analyzer could detect if the built in method from Date is used, but the tests likely can not.

kytrinyx:
> Alternative middle step may be the add the mention in the documentation, as an append.

That would work, too.

> I would be as delighted, though, to see a refinement on the Date class, as much as a "not Date class" solution. I believe an analyzer could detect if the built in method from Date is used, but the tests likely can not.

Agreed. I think an analyzer would be a really nice improvement for this exercise.

PatrickMcSweeny:
I would like to work on this if it is still active.

kotp:
That would be awesome @PatrickMcSweeny , and I have modified the labels and assignment assigned to you.

PatrickMcSweeny:
Thanks. Do you or @kytrinyx have any more details to add as to what functionality the analyzer should have?

PatrickMcSweeny:
I found the [ruby-analyzer repo](https://github.com/exercism/ruby-analyzer) and I think the existing analyzers in there should be enough to get me started.

kotp:
> I found the [ruby-analyzer repo](https://github.com/exercism/ruby-analyzer) and I think the existing analyzers in there should be enough to get me started.

This is where I would have pointed you to, let us know if you need some assistance.

PR #1752, opened 2025-05-19
Append leap instructions to instruct user not use Ruby Date class

PatrickMcSweeny:
The discussion in #1532 called for appending the instructions for leap to instruct the user not to use the Date class in Ruby.

Discussion on forums is:  [Append Ruby leap instructions to instruct user not use Date class](https://forum.exercism.org/t/append-ruby-leap-instructions-to-instruct-user-not-use-date-class/17515).

### DIFF
--- a/exercises/practice/leap/.docs/instructions.append.md
+++ b/exercises/practice/leap/.docs/instructions.append.md
@@ -1,0 +1,3 @@
+# Restrictions
+
+Avoid using `Date#leap?` from the Standard Library.


exercises/practice/leap/.docs/instructions.md
```diff
@@ -1,3 +1,3 @@
 # Instructions
 
-Your task is to determine whether a given year is a leap year.
+Your task is to determine whether a given year is a leap year without using any methods from Ruby's Date class.
```
BNAndras, line 3: [Track-specific details should go in an `instructions-append.md`
](https://exercism.org/docs/building/tracks/practice-exercises#h-file-docs-instructions-append-md). `instructions.md` is sourced from an upstream problem-specifications repo that all Exercism tracks share for canonical exercise documentation and test cases. If you update the instructions directly, our tooling will see that they're out of sync and suggest that the maintainers resync the canonical documentation, potentially losing your edits in the process.

kotp: Look for examples of using an appends file to accomplish what you are attempting to do.

kotp: Example result from `find` command:

```sh
./exercises/practice/ocr-numbers/.docs/instructions.append.md
./exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
./exercises/practice/series/.docs/instructions.append.md
./exercises/practice/dnd-character/.docs/instructions.append.md
./exercises/practice/accumulate/.docs/instructions.append.md
./exercises/practice/kindergarten-garden/.docs/instructions.append.md
./exercises/practice/high-scores/.docs/instructions.append.md
./exercises/practice/robot-name/.docs/instructions.append.md
./exercises/practice/anagram/.docs/instructions.append.md
./.appends
```

PatrickMcSweeny:
I updated my commit to create a separate append file and instruct the user not to use the date class specifically.

exercises/practice/leap/.docs/instructions.append.md
```diff
@@ -0,0 +1,3 @@
+# Restrictions
+
+Create your own implementation rather than using the Date class from the Ruby standard library.
```
kotp, line 3: This might also work:

```suggestion
Avoid using `Date#leap?` from the Standard Library.
```

Simpler English, and uses syntax that you would see to reference the exact method.

BNAndras: Yeah, I like that. Out of curiosity, is standard library supposed to be capitalized in Ruby circles? Google search seems to have search results for it both being capitalized and not capitalized. Generally, standard library doesn't refer to something specifically with that name so I don't see it capitalized typically.

kotp: Probably Ruby standard library is good.  The link provides a direct link to the specific thing in the standard library.

PatrickMcSweeny: The tests also prevent the user from using the 'gregorian_leap' or 'julian_leap' methods, so I updated it to be 'Avoid using the Date class from the Standard Library.'

kotp: That does not do what you think it does (if you think it prevents the user from using those).

You can see this if you test it, by writing this code in the example.rb file, and run `rake test:leap`:

```ruby
  def leap?
    Date.gregorian_leap?(number)
  end
```

I always thought it was just there to be informative to the reader, but never thought it was meant to be "effective code".

There are trivial ways to get around it, even if it might work in certain circumstances.

kotp: What I would suggest doing, I think, is to leave the advice to the analyzer, and suggest to not use `Date#leap?` explicitly here.

PatrickMcSweeny: I updated my commit to what you suggested earlier.

kotp (approved):
I will approve it, but consider the case of "Standard Library" and make a judgement call to leave it or change it to Standard library or standard library. I think the Core and Standard libraries may be title cased as they are the names given to each library, collection of classes and modules, in Ruby that are supplied when installed.

PatrickMcSweeny:
This docs site uses title case, so I will go with that.

https://docs.ruby-lang.org/en/3.3/standard_library_rdoc.html

PR #1757, opened 2025-06-12
Clarify blackjack instructions

PatrickMcSweeny:
I opened a topic about this [here](https://forum.exercism.org/t/suggestion-for-blackjack-exercise-instructions/17857), and I didn't sense any pushback, so I figured I would create a PR. This is similar to a change made in the Go track in [this PR](https://github.com/exercism/go/pull/2855).

### DIFF
--- a/exercises/concept/blackjack/.docs/instructions.md
+++ b/exercises/concept/blackjack/.docs/instructions.md
@@ -62,7 +62,7 @@ Depending on your two cards and the card of the dealer, there is a strategy for 
 Although not optimal yet, you will follow the strategy your friend Alex has been developing, which is as follows:
 
 - If you have a pair of aces you must always split them.
-- If you have a Blackjack (two cards that sum up to a value of 21), and the dealer does not have an ace, a figure or a ten then you automatically win.
+- If you have a Blackjack (two cards that sum up to a value of 21), and the dealer does not have an ace, a face card or a ten then you automatically win.
   If the dealer does have any of those cards then you'll have to stand and wait for the reveal of the other card.
 - If your cards sum up to a value within the range [17, 20] you should always stand.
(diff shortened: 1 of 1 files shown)


exercises/concept/blackjack/.docs/instructions.md
```diff
@@ -62,7 +62,7 @@ Depending on your two cards and the card of the dealer, there is a strategy for
 Although not optimal yet, you will follow the strategy your friend Alex has been developing, which is as follows:
 
 - If you have a pair of aces you must always split them.
-- If you have a Blackjack (two cards that sum up to a value of 21), and the dealer does not have an ace, a figure or a ten then you automatically win.
+- If you have a Blackjack (two cards that sum up to a value of 21), and the dealer does not have an ace, a face card (Jack, Queen, King) or a ten then you automatically win.
```
kotp:
```suggestion
- If you have a Blackjack (two cards that sum up to a value of 21), and the dealer does not have an ace , 10, or a face card (Jack, Queen, King) then you automatically win.
```

The 10 could be, and probably should be, mentioned first.

PatrickMcSweeny: Changing "ten" to "10" would be inconsistent with the rest of the instructions.

kotp: Also, this work is more than what is claimed in the discussion, as we have not only changed the word from "figure" to "face card" but have given a list of face cards here.

kotp: > Changing "ten" to "10" would be inconsistent with the rest of the instructions.

You are correct, it was here to talk about, not to commit directly.  The names we are using as names and values as values, we definitely should keep doing that.

kotp (approved):

SleeplessByte (approved):

PR #1765, opened 2025-07-25
Make so the test file name is formated in snakecase by generator

meatball133:


generatorv2/lib/generator.rb
```diff
@@ -38,6 +38,10 @@ def camel_case(str)
     str.split(/[-_]/).map(&:capitalize).join
   end
 
+  def to_snake(str)
+    str.gsub(/([a-z\d])([A-Z])/, '\1_\2').tr("-", "_").downcase
```
kotp: Why not...

```suggestion
    str.tr("-", "_").downcase
```

Are there tests for this file?

generatorv2/lib/generator.rb
```diff
@@ -38,6 +38,10 @@ def camel_case(str)
     str.split(/[-_]/).map(&:capitalize).join
   end
 
+  def to_snake(str)
```
kotp: I see `camel_case` right above here, should this be `snake_case` instead?

kotp (changes requested):
Suggested change.

How are we testing this?  (I did not see the test file updated)  Via `rake test:generator` to answer my own question.

generatorv2/lib/generator.rb
kotp: Also I see `underscore` is very similar.  And the tests file has not been update to reflect testing of this new public method.

meatball133: Didn't think about that method. Was quite tired when I made this commit, but good suggestion.

kotp (approved):

PR #1770, opened 2025-08-30
Fix exercise naming in generator

meatball133:
The fix for the exercise name was only applied for the generator not the bin script. The underscore method in the generator will now clear all characters which is not numbers or letters. Noticed this would conflict when a description on problem spec had for example "," in it. Further, are there some formatting changes to make so it formats the files correctly.

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -9,6 +9,10 @@ def exercises
     .select { |file| File.directory? File.join('./exercises/practice', file) }
 end
 
+def underscore(str)
+  str.gsub(/[^a-zA-Z0-9_-]/, '').gsub(/[-\s]/, '_').downcase
+end
+
 class VerificationError < StandardError
   MESSAGE = 'The result generated for %<exercise>s, does not match the current file'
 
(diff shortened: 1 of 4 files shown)


bin/generate
```diff
@@ -9,6 +9,10 @@ def exercises
     .select { |file| File.directory? File.join('./exercises/practice', file) }
 end
 
+def underscore(str)
+  str.gsub(/[^a-zA-Z0-9_-]/, '').gsub(/[-\s]/, '_').downcase
```
kotp: Should this use the character class instead of listing these, or is this limited to only the mentioned characters?

generatorv2/lib/generator.rb
```diff
@@ -21,17 +21,17 @@ def generate(result_path = "./exercises/practice/#{@exercise}/#{underscore(@exer
...
 
     File.write(result_path, result)
     RuboCop::CLI.new.
-      run(['-x', '-c', '.rubocop.yml', '-o', NullDevice.path, result_path])
+      run(['-a', '-c', '.rubocop.yml', '-o', NullDevice.path, result_path])
   end
 
   def underscore(str)
-    str.gsub(/[-\s]/, '_').downcase
+    str.gsub(/[^a-zA-Z0-9_\- ]/, '').gsub(/[-\s]/, '_').downcase
```
kotp: Should we use a character class here as well?

bin/generate
meatball133: Good point, I don't write regex that often, so I am not that good at it. Likely something I should work on.

kotp (approved):

PR #1773, opened 2025-08-31
Add templates for exercises batch 3

meatball133:
Includes:

- Circular Buffer
- Crypto square
- Custom set
- Darts
- Diamond
- Differnce of square
- Dnd character

### DIFF
--- a/exercises/practice/circular-buffer/.meta/test_template.erb
+++ b/exercises/practice/circular-buffer/.meta/test_template.erb
@@ -4,7 +4,7 @@ require_relative 'circular_buffer'
 class CircularBufferTest < Minitest::Test
 <% json["cases"].each do |cases| %>
     def test_<%= underscore(cases["description"]) %>
-        <%= skip? %> 
+        <%= skip? %>
         buffer = CircularBuffer.new(<%= cases["input"]["capacity"] %>)
         <%- cases["input"]["operations"].each do |operation| -%>
             <%- if operation["operation"] == "read"-%>
--- a/exercises/practice/darts/.meta/test_template.erb
+++ b/exercises/practice/darts/.meta/test_template.erb
@@ -4,7 +4,7 @@ require_relative 'darts'
 class DartsTest < Minitest::Test
 <% json["cases"].each do |cases| %>
     def test_<%= underscore(cases["description"]) %>
-        <%= skip? %> 
+        <%= skip? %>
         darts = Darts.new(<%= cases["input"]["x"] %>, <%= cases["input"]["y"] %>)
         assert_equal(<%= cases["expected"] %>, darts.score)
     end
--- a/exercises/practice/difference-of-squares/.meta/test_template.erb
+++ b/exercises/practice/difference-of-squares/.meta/test_template.erb
@@ -5,7 +5,7 @@ class DifferenceOfSquaresTest < Minitest::Test
 <% json["cases"].each do |cases| %>
     <% cases["cases"].each do |sub_case| %>
         def test_<%= underscore(sub_case["description"]) %>
-            <%= skip? %> 
+            <%= skip? %>
             assert_equal <%= sub_case["expected"] %>, Squares.new(<%= sub_case["input"]["number"] %>).<%= sub_case["property"] == "differenceOfSquares" ? "difference" : sub_case["property"].gsub(/([A-Z])/, '_\1').downcase.sub(/^_/, '') %>
         end
     <% end %>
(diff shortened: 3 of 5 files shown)


generatorv2/lib/utils.rb
```diff
@@ -47,7 +47,7 @@ def additional_json(json)
   def remove_tests(uuid, json)
     json["cases"].each_with_object([]) do |x, acc|
       if x["cases"]
```
kotp:
```suggestion
      acc << if x["cases"]
```

This, along with ...

generatorv2/lib/utils.rb
```diff
@@ -47,7 +47,7 @@ def additional_json(json)
...
-        acc << { "cases" => remove_tests(uuid, x) }
+        acc << { "cases" => remove_tests(uuid, x), "description" => x["description"] }
       elsif uuid.include?(x["uuid"])
         acc << x
```
kotp: ... with this...

```suggestion
        { "cases" => remove_tests(uuid, x), "description" => x["description"] }
      elsif uuid.include?(x["uuid"])
        x
```

... removes the unconditional work from the conditional branches.

I did not touch the double quotes, though.  Unsure if that is already attended to in another place, if we are doing that.

generatorv2/lib/utils.rb
meatball133: This seems to create error when I do it that way

kotp: Can you show the patch, how you attempted this?

exercises/practice/diamond/diamond_test.rb
```diff
@@ -2,34 +2,105 @@
...
+            "  B B  \n"\
+            "   A   \n"\
+
+    assert_equal string, answer
+  end
+
+  def test_largest_possible_diamond
+    skip
+    answer = Diamond.make_diamond('Z')
+    string = "                         A                         \n"\
```
kotp, line 52: Consider using HereDoc for this, rather than consecutive string "trick".

kotp: Same for the others.

kotp (changes requested):

exercises/practice/circular-buffer/.meta/test_template.erb
```diff
@@ -0,0 +1,30 @@
+require 'minitest/autorun'
+require_relative 'circular_buffer'
+
+class CircularBufferTest < Minitest::Test
+<% json["cases"].each do |cases| %>
+    def test_<%= underscore(cases["description"]) %>
+        <%= skip? %>
```
kotp:
```suggestion
        <%= skip? %>
```

exercises/practice/darts/.meta/test_template.erb
```diff
@@ -0,0 +1,12 @@
+require 'minitest/autorun'
+require_relative 'darts'
+
+class DartsTest < Minitest::Test
+<% json["cases"].each do |cases| %>
+    def test_<%= underscore(cases["description"]) %>
+        <%= skip? %>
```
kotp:
```suggestion
        <%= skip? %>
```

exercises/practice/difference-of-squares/.meta/test_template.erb
```diff
@@ -0,0 +1,13 @@
+require 'minitest/autorun'
+require_relative 'difference_of_squares'
+
+class DifferenceOfSquaresTest < Minitest::Test
+<% json["cases"].each do |cases| %>
+    <% cases["cases"].each do |sub_case| %>
+        def test_<%= underscore(sub_case["description"]) %>
+            <%= skip? %>
```
kotp:
```suggestion
            <%= skip? %>
```

kotp (commented):
Trailing whitespace.

generatorv2/lib/utils.rb
meatball133: https://github.com/exercism/ruby/pull/1773/commits/169890685e3437d176c3baa7edbf56dcc0bf6f57. And you see that when main was merged in it failed. But when rollbacked it worked: https://github.com/exercism/ruby/pull/1773/commits/1b71ff2d75a71cdad1dc0352f97431fa83eb18db

exercises/practice/diamond/diamond_test.rb
meatball133: I tested it out but the problem I find is that for this exercise is trailing spaces important. So what end up happening is that each line is appended with #{'       '}. Which in my opinion, looks just weird.

kotp: I agree with that, many people do not notice the white space that is not supposed to be there, they are surely going to miss the whitespace that is supposed to be there.

I have done it with a HereDoc and dynamic format codes, and have done it with append and multiplied spaces, so there are a few ways to address that... but for the tests, it needs to be clear to the reader.

kotp (approved):

PR #1779, opened 2025-08-31
Fix verification script 

meatball133:
When the script threw an error, it would rescue it. Now it doesn't do that; instead, it instantly fails.

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -48,7 +48,7 @@ parser.on('--verify', 'Verify all exercises') do
       Generator.new(exercise).generate(f.path)
       generated_code = f.read
       File.delete(f.path)
-      fail VerificationError.new(exercise) unless current_code == generated_code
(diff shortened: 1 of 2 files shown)


bin/generate
```diff
@@ -48,10 +48,8 @@ parser.on('--verify', 'Verify all exercises') do
       Generator.new(exercise).generate(f.path)
       generated_code = f.read
       File.delete(f.path)
-      fail VerificationError unless current_code == generated_code
+      fail VerificationError.new(exercise) unless current_code == generated_code
```
kotp: Just to note:

```suggestion
      fail VerificationError, exercise unless current_code == generated_code
```

That would be identical behavior, and I believe how you generally see this constructed.

kotp (approved):
Approved with potential change.

PR #1781, opened 2025-09-01
Fix typo in moviegoer

lemonade-37:
Typos in the Moviegoer documentation file have been corrected.
Please review.

Reference:
- [Moviegoer](https://exercism.org/tracks/ruby/exercises/moviegoer)
- [Exercism Community Topics（Fix typo in Moviegoer HINTS.md）](https://forum.exercism.org/t/fix-typo-in-moviegoer-hints-md/19038)

### DIFF
--- a/exercises/concept/moviegoer/.docs/hints.md
+++ b/exercises/concept/moviegoer/.docs/hints.md
@@ -6,7 +6,7 @@
 
 ## 2. Check if a moviegoer is allowed to see scary movies
 
-- [Compare][doc-integer-gtoe] the moviegoer's age with the minimum age allowed to see scary movies. You don't even need thee ternary operator in this task.
+- [Compare][doc-integer-gtoe] the moviegoer's age with the minimum age allowed to see scary movies. You don't even need the ternary operator in this task.
 
 ## 3. Check if a moviegoer is entitled to free popcorn
 
--- a/exercises/concept/moviegoer/.docs/instructions.md
+++ b/exercises/concept/moviegoer/.docs/instructions.md
@@ -24,7 +24,7 @@ The cinema has a simplified age-verification system.
 If you are 18 or over you can watch scary movies.
 If you are younger, you cannot.
 
-Implement the `Moviegoer.watch_scary_movie?` method.
+Implement the `Moviegoer#watch_scary_movie?` method.
 It should return whether someone is allowed to watch the movie or not.
 
 ```ruby


exercises/concept/moviegoer/.docs/instructions.md
```diff
@@ -24,7 +24,7 @@ The cinema has a simplified age-verification system.
 If you are 18 or over you can watch scary movies.
 If you are younger, you cannot.
 
-Implement the `Moviegoer.watch_scary_movie?` method.
+Implement the `Moviegoer#watch_scary_movie?` method.
```
kotp, line 27: This is not a typographical error, it is as intended.  The method called is an instance method, as we can see and verify from the tests:

```ruby
def test_regular_ticket_price
  assert_equal 15, Moviegoer.new(59).ticket_price
end
```

exercises/concept/moviegoer/.docs/hints.md
```diff
@@ -6,7 +6,7 @@
 
 ## 2. Check if a moviegoer is allowed to see scary movies
 
-- [Compare][doc-integer-gtoe] the moviegoer's age with the minimum age allowed to see scary movies. You don't even need thee ternary operator in this task.
+- [Compare][doc-integer-gtoe] the moviegoer's age with the minimum age allowed to see scary movies. You don't even need the ternary operator in this task.
```
kotp, line 9: This change is good.

exercises/concept/moviegoer/.docs/instructions.md
lemonade-37: Since it was originally documented within exercises/concept/moviegoer/moviegoer.rb, I believe “Moviegoer#watch_scary_movie” is correct. What do you think?

```ruby
  def watch_scary_movie?
    raise 'Please implement the Moviegoer#watch_scary_movie method'
  end
```

Additionally, instance methods of the same format are also written using “#” in the problem statement.
Therefore, I thought it would be better to standardize the notation for instance methods.

> Write the `Moviegoer#ticket_price` method using a ternary statement.

kotp: I know for a fact that it is correct.  `Class#method_name` is for instance methods and `Class.method_name`, as well as the less used `Class::method_name` is for class methods.  The first example is not usable invocable code, but is how we have documented Ruby code since well before 1.8.5.

kotp (approved):

kotp:
Sorry for the noise, we are saying the same thing, I am being round-about with it, and did not realize it.

PR #1786, opened 2025-09-08
Add templates for exercises batch 6

meatball133:
Includes:

- Largest series product
- Leap
- Linked list
- Luhn
- Matching brackets
- Matrix

Note:

Linked list have breaking changes in form of adding two extra methods that exist in problem spec.

### DIFF
--- a/exercises/practice/leap/.meta/test_template.erb
+++ b/exercises/practice/leap/.meta/test_template.erb
@@ -3,7 +3,7 @@ require_relative 'leap'
 
 class Date
   def leap?
-    raise "Implement this yourself instead of using Ruby's implementation."
+    raise "Write this functionality yourself instead of using Ruby's `leap?` method."
   end
 
   alias gregorian_leap? leap?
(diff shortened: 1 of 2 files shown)


exercises/practice/leap/.meta/test_template.erb
```diff
@@ -0,0 +1,20 @@
+require 'minitest/autorun'
+require_relative 'leap'
+
+class Date
+  def leap?
+    raise "Implement this yourself instead of using Ruby's implementation."
```
kotp:
```suggestion
    raise "Write this functionality yourself instead of using Ruby's `leap?` method."
```

kotp: Because "implement the implement you implement yourself, rather than the implement that Ruby implements" can be very confusing, if accurate and proper use, but I tend to avoid this word in multicultural and multilingual audiences.  

English is hard enough as it is, using a word with 4 senses when people tend to not take care or be clear which sense they are talking about can be doubly hard for those that do not have English as a first language.

kotp (approved):
Approved, with normal historical nag, and suggested replacement phrasing.

PR #1787, opened 2025-09-14
Add templates for exercises batch 7

meatball133:
Includea:

- Meetup
- Nth prime
- Palindrome products
- Pangram
- Pascal triangle
- Phone number


exercises/practice/palindrome-products/palindrome_products_test.rb
```diff
@@ -2,33 +2,25 @@
 require_relative 'palindrome_products'
 
 class PalindromesTest < Minitest::Test
-  def test_smallest_palindrome_from_single_digit_factors
+  def test_find_the_smallest_palindrome_from_single_digit_factors
```
kotp: Not really a fan of the phrase "test find the" but "test for the" makes sense to me.

exercises/practice/pangram/pangram_test.rb
```diff
@@ -4,71 +4,71 @@
 class PangramTest < Minitest::Test
   def test_empty_sentence
     # skip
-    sentence = ""
+    sentence = ''
     result = Pangram.pangram?(sentence)
-    refute result, "Expected false, got: #{result.inspect}. #{sentence.inspect} is NOT a pangram"
+    refute result, "Expected false, got: #{result.inspect}. #{sentence.inspect} is not a pangram"
```
kotp: Have we slipped from using "expected, actual" as a pattern?  The result, I would suspect, would be resolution of the tests, rather than an evaluation of what is actually being under test.

exercises/practice/meetup/.meta/test_template.erb
```diff
@@ -0,0 +1,12 @@
+require 'minitest/autorun'
+require_relative 'meetup'
+
+class MeetupTest < Minitest::Test
+<% json["cases"].each do |cases| %>
+    def test_<%= underscore(cases["description"]) %>
+        <%= skip? %>
+        meetup = Meetup.new(<%= cases["input"]["month"] %>, <%= cases["input"]["year"] %>).day(:<%= cases["input"]["dayofweek"].downcase %>, :<%= cases["input"]["week"] %>)
```
kotp: Conceptually, and as a pattern, the tests are `assert expected, actual` and we can use these names.  If we are careful with this, a high percentage of template portions will follow this pattern, the only difference being what is referenced by `actual` as to what is called, while the expected is always what we expect.

kotp (approved):
Approving, as these are "good", but we might consider the templates being more cookie cutter.

exercises/practice/pangram/pangram_test.rb
meatball133: I am not sure what you mean. Could you give an example when this happen?

kotp: In history?  I think the old templating system, I used expected, actual was used for that.

Structurally, or as a pattern, I think it is helpful to have:

```ruby
class TriangleTest < Minitest::Test
<% json["cases"].each do |cases| %>
    def test_<%= underscore(cases["description"]) %>
        <%= skip? %>
        actual = Triangle.new(<%= cases["input"]["count"] %>).rows
        expected = <%= cases["expected"] %>
        assert_equal expected, actual, 'Helpful message if appropriate"
    end
<% end %>
```

Which in every assertion line being something like `assert expected, actual, message`, and the details of what the expected and actual being on their own lines, which can make reading the tests easier as well.

meatball133: Okay I I think I understand. And sorry for taking sometime for reply. I have had quite a bit this week so I only have the weekend over to open source work. 

A big part of templating is just that we can in aftermath make changes really quickly, like 1 line change or something can do it for an entire test file. My goal with the templating is to make the styling the same as the old one mostly as a proof on concept that the generator can handle it. In some cases have I changed the test cases in either dissucused style changes or to make the exercise in line with problem spec.

I have no problem start implementing these style changes now but on the other hand there is quite a few already merged in templates so as I see it we also have to go back and change those as well.

kotp: Right on.  I did consider not saying anything at all... and it could be an "after the fact" change, since it is definitely pattern work, and should not affect test functionality.

If I remember correctly, I was at least co-author of the original templating system that was here before.  And also knew that students are encouraged to read the tests (at least for the practice exercises) as there were no concept exercises back then at all.

If you want to see where we can keep the simplistic assertion lines by making expected, actual be the pattern, that would be appreciated I think even further than only myself.  But I would also be tickled pink, if that were possible.

kotp: > And sorry for taking sometime for reply. I have had quite a bit this week so I only have the weekend over to open source work.

<details><summary>Late?  Early?</summary>

GandalfAWizardIsNeverLaterGIF

</details>

meatball133: Yeah I have nothing against the practice but I think there are somethings which might have to be disscused. Like the palindrome products do have multiple asserations in the same test cases. 

e.g:

```ruby
    palindromes = Palindromes.new(max_factor: 9)
    palindromes.generate
    largest = palindromes.largest
    assert_equal 9, largest.value
    assert_includes [[[1, 9], [3, 3]], [[3, 3], [1, 9]]], largest.factors
```

How should a situation like this be structured?

kotp: Two different concerns for the test, so probably very similar to a method with the word "and" as part of the name.  It probably should be two different tests.

But also this test makes perhaps very little sense (not looking at the name of the test at the moment to get "context") at first blush, since it would be impossible to get 9 if 9 were not involved.  I think one or the other might be important, but together, one follows the other, while the other does not necessarily follow the other.

It likely is simply a "bad test" in general.

If this is an attempt at "DRY" tests are not generally the place for that, either.

meatball133: Yes I was thinking the same to split it into two test cases. But I think for a scenario like this (from binary tree exercise):

```ruby
    four = Bst.new 4
    four.insert 2
    four.insert 6
    four.insert 1
    four.insert 3
    four.insert 7
    four.insert 5
    assert_equal 4, four.data
    assert_equal 2, four.left.data
    assert_equal 1, four.left.left.data
    assert_equal 3, four.left.right.data
    assert_equal 6, four.right.data
    assert_equal 5, four.right.left.data
    assert_equal 7, four.right.right.data
```

Would this turn into a ton of test cases and make the file very long. Is still the approach of spliting each assert into its own test applicable here?

kotp: This could easily be a single assertion, though.  Store the expected results in a collection expected values and assert that the collection of actual values are identical to the collection of expected values.

I have never really loved that test either. ;)

meatball133: Yeah but like should we also do a singel assertion in palindromes? It is here I think a distiction in how we handle these kinds of cases have to be made (and it could be a mix of those two it doesnt have to be one or the other).

kotp: Can you link to what you are looking at?  I went to the [palindromes Line 5](https://github.com/exercism/ruby/blob/10c4b33f6bff97ef5159bf765a759fbcf3352580/exercises/practice/palindrome-products/palindrome_products_test.rb#L5) to see what I could see.

This is how it is:

```ruby
  def test_smallest_palindrome_from_single_digit_factors
    palindromes = Palindromes.new(max_factor: 9)
    palindromes.generate
    smallest = palindromes.smallest
    assert_equal 1, smallest.value
    assert_equal [[1, 1]], smallest.factors
  end
```

This is how it may be (untested, here for demonstration and not refactored to "base":

```ruby
  def test_smallest_palindrome_from_single_digit_factors
    palindromes = Palindromes.new(max_factor: 9).generate # if generate returns the instance
    actual = palindromes.smallest.value, palindromes.smallest.factors
    expected = 1, [1, 1]
    assert_equal expected, actual
  end
```

This may be improved further, not looking at the exercise in full. But kind of pattern can be useful for us overall.  It also presents a certain pattern of familiarity for the student, as they know what to look for while reading the test files, and how to decipher them a bit more easily, I think.

This generator, yeah, it is for our benefit, but everything exposed to the student, the result that comes from the generator, should be for their benefit.

meatball133: Alright I made a new [commit] with the changes we have disscussed let me know if it was this you had in mind. Since what I was thinking about was that the palindrome products case with `assert_includes` might be a bit confusing but it might actually work.

[commit]: https://github.com/exercism/ruby/pull/1787/commits/6177ebae9ae32b170bf55a5131d14483b97ac4aa

kotp: I modified your message for what I would encourage for readability for the reader.  It gets the reader to the points that you have in mind (hopefully) to look at without having to spend time working to interpret what is actually referred to.

Not a criticism, but you likely have what is in mind, the reader needs to interpret this to understand, and I hope that the changes reflect what your thoughts were.

exercises/practice/palindrome-products/palindrome_products_test.rb
```diff
@@ -2,94 +2,94 @@
...
-  def test_find_the_largest_palindrome_from_single_digit_factors
+  def test_for_the_largest_palindrome_from_single_digit_factors
     skip
     palindromes = Palindromes.new(max_factor: 9)
     palindromes.generate
-    largest = palindromes.largest
-    assert_equal 9, largest.value
-    assert_includes [[[1, 9], [3, 3]], [[3, 3], [1, 9]]], largest.factors
+    actual = palindromes.largest.value, palindromes.largest.factors
+    expected = [[9, [[1, 9], [3, 3]]], [9, [[3, 3], [1, 9]]]]
```
kotp: Likely one extra pair of `[]` brackets... Gotta go back and look at how this is created.  But also, perhaps not really problematic.

Just knowing that the assignment itself will be evaluated as an Array is good information to have in mind as a Ruby developer though.

We should avoid nesting to the extent that we can though, as it can be confusing as a learner.  (It goes the other way, though, if you are unaware.)

kotp (commented):
Definitely looking better, not tested locally.

kotp (approved):

exercises/practice/palindrome-products/.meta/test_template.erb
```diff
@@ -16,7 +16,7 @@ class PalindromesTest < Minitest::Test
                 palindromes = Palindromes.new(max_factor: <%= cases["input"]["max"] %> <%= cases["input"]["min"] != 1 ?  ", min_factor: #{cases["input"]["min"]}" : "" %>)
                 palindromes.generate
                 actual = palindromes.<%= cases["property"] %>.value, palindromes.<%= cases["property"] %>.factors
-                expected = <%= cases["expected"]["factors"].size > 1 ? "#{cases["expected"]["factors"].permutation(2).to_a.map{|factors| [cases["expected"]["value"], factors]}}" : "#{cases["expected"]["value"] ? cases["expected"]["value"] : "nil"}, #{cases["expected"]["factors"]}"%>
+                expected = <%= cases["expected"]["factors"].size > 1 ? "#{ cases["expected"]["factors"].permutation(2).to_a.map{|factors| [cases["expected"]["value"], factors].to_s}.join(", ")}" : "#{cases["expected"]["value"] ? cases["expected"]["value"] : "nil"}, #{cases["expected"]["factors"]}"%>
```
kotp, line 19: No space inside delimiter for interpolation.  Will not cause a problem, but looks odd since it is "imbalanced".  However, space after method name and block.  Space inside block delimiter (before the block argument list delimiter, and space inside ending of the block delimiter.

This may be fixed in a post generation process if we are using something like a linter, though.  But we can and should address it in the generator.

kotp:
Any time you are done with this, it is surely at a "good enough, avoid perfection as the enemy of being in place."

PR #1792, opened 2025-09-23
Add templates for exercises batch 9

meatball133:
Includes:

- Raindrops
- Resistor color
- Resistor color duo
- Resistor color trio
- Reverse string
- Rna transcription

### DIFF
--- a/exercises/practice/raindrops/.meta/test_template.erb
+++ b/exercises/practice/raindrops/.meta/test_template.erb
@@ -1,0 +1,13 @@
+require 'minitest/autorun'
+require_relative 'raindrops'
+
+class RaindropsTest < Minitest::Test
+<% json["cases"].each do |cases| %>
+    def test_<%= underscore(cases["description"]) %>
+      <%= skip? %>
+      actual = Raindrops.convert(<%= cases["input"]["number"] %>)
+      expected = '<%= cases["expected"] %>'
+      assert_equal expected, actual
+    end
+    <% end %>
+end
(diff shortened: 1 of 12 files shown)


exercises/practice/resistor-color-trio/.meta/example.rb
```diff
@@ -27,6 +27,12 @@ def value
...
 
   def to_s
-    value < 1000 ? "#{value} ohms" : "#{(value.to_f/1000).to_i} kiloohms"
+    index = 0
+    value_copy = value
+    while value_copy >= 1000 && index < LABELS.size - 1
+      value_copy /= 1000
+      index += 1
+    end
+    "#{value_copy} #{LABELS[index]}"
```
kotp: Trying to understand why the example files are changed here... 

I thought this was about templates, and example files are not generally given by the upstream.

meatball133: problem spec added tests which were not coverd by the original example.

kotp: This should likely be a patch outside of this patch-set then.

Thanks for doing the work for the additional (breaking changes, I presume) functionality for the example.rb file though.

kotp (approved):

exercises/practice/resistor-color-trio/.meta/example.rb
meatball133: Yeah likely, I could still split it if you like?

kotp: That would be good, because if a revert were needed for one or the other, then they would not be both affected.

meatball133:
Depends on: https://github.com/exercism/ruby/pull/1794

kotp (approved):

PR #1794, opened 2025-09-29
Update resistor-color-trio

meatball133:
Branched of #1792, update the exercise to be up to spec with problem spec. Adds tests for higher numbers.

### DIFF
--- a/exercises/practice/resistor-color-trio/.meta/example.rb
+++ b/exercises/practice/resistor-color-trio/.meta/example.rb
@@ -1,5 +1,6 @@
 class ResistorColorTrio
   COLOR_CODES = %i[black brown red orange yellow green blue violet grey white].freeze
+  LABELS = %w[ohms kiloohms megaohms gigaohms teraohms].freeze
 
   def initialize(colors)
     @colors = colors.map(&:to_sym)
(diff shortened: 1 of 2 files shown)


exercises/practice/resistor-color-trio/.meta/example.rb
```diff
@@ -1,17 +1,17 @@
...
+  LABELS = %w[ohms kiloohms megaohms gigaohms teraohms].freeze
 
   def initialize(colors)
     @colors = colors.map(&:to_sym)
     @tens, @ones, @zeros = @colors
   end
 
   def label
-    "Resistor value: #{to_s}"
+    "Resistor value: #{self}"
```
kotp, line 11: So much better!

kotp:
> > Branched of #1792, update the exercise to be up to spec with problem spec. Adds tests for higher numbers.
> 
> Should we add "and adds template"? Otherwise, I would expect the update for the spec but not the template.

On further though, and given the reason for this was so that "one could be reverted..." including the template here is causing the same concern.  The template should be done in the other PR where templates are done.

meatball133:
I will remove the template but I will still keep the updated test file, since I think that is what you are meaning

PR #1807, opened 2025-12-13
Update supported versions of Ruby

kotp:
- description in README
- version for test in CI


meatball133:
But I still think testing 3.3 makes sense, it just that I think all the tooling and the "main" version should be brought to 3.4

kotp:
> But I still think testing 3.3 makes sense, it just that I think all the tooling and the "main" version should be brought to 3.4

We can avoid deprecated and dropped syntax and test for 3.4, knowing that (some) students will be using prior versions, though.

We could update it to 3.3 though, knowing that we have a end of life coming up, and moving it up at that time, though.

I did conscientiously say "want" rather than "need" to indicate the version of 3.4, since it should not be a necessity currently, but an indication that joy will be more likely had with the newer version.

So `exercise-tests.yml` could be:

```yml
ruby-version: [3.3, 3.4]
```

While leaving the documentation as it is changed here.  What do you think?

kotp:
The currently failing test is "normal", though undesirable.

meatball133 (approved):
> The currently failing test is "normal", though undesirable.

I am a bit afraid if that tests fails all the time thinking it is an exercise test though

IsaacG (approved):

IsaacG:
Can we get 3.4 working?

kotp:
> Can we get 3.4 working?

We can, and I think it is stable... it should not "sometimes fail".

kotp:
Thank you @meatball133  and @IsaacG I think this is good for the next 3 months (3.3 end of life comes in about 6 months, and 4.0+ is in my sight).

PR #1824, opened 2026-03-27
[SimpleCalculator] added note about exception scope

VaiaPatta1985:
A note was added to instructions clarifying that UnsupportedOperation is expected to be defined for a specific scope. Solves issue #6772


kotp (approved):
Nice, leaves the requirement in, the fact that it is different from the others as a learning point.

Let me know if there is something that can gently get the student to be able to know how to do this, without necessarily going through mentoring the exercise, but in a way that it does not give the solution directly away.

PR #1835, opened 2026-05-18
Fixed last_will_test #5 for Actual and Expected

hurley-matt:
Discussed in the forums [here].

[here]: https://forum.exercism.org/t/fixed-last-will-test/51845

### DIFF
--- a/exercises/concept/last-will/last_will_test.rb
+++ b/exercises/concept/last-will/last_will_test.rb
@@ -36,6 +36,6 @@ class LastWillTest < Minitest::Test
 
   def test_code_fragments_fit
     # skip
-    assert_equal ::EstateExecutor.assemble_code, 1_925_550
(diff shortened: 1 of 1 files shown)


exercises/concept/last-will/last_will_test.rb
```diff
@@ -36,6 +36,6 @@ def test_code_fragment_number_assembly_method_in_correct_namespace
 
   def test_code_fragments_fit
     # skip
-    assert_equal ::EstateExecutor.assemble_code, 1_925_550
+    assert_equal 3253, ::EstateExecutor.assemble_code
```
kotp, line 39: Definitely is in the right order... expected and actual in that order.

More to "fixup" in the semantics of the test.

Since it is already here, would you like to fix the rest as well?

VaiaPatta1985 (commented):
This PR changes the expected answer of the last test to an incorrect value, and should not be merged.

PR #1841, opened 2026-06-22
test: correct assert_equal argument order in last_will tests

Lemmah:
Discussed in forums [here:](http://forum.exercism.org/t/fixed-last-will-test/51845/15)

## Summary

This PR updates the `last_will` exercise tests to follow the standard Minitest convention of passing arguments to `assert_equal` as:

```ruby
assert_equal expected, actual
```

instead of:

```ruby
assert_equal actual, expected
```

## Why?

While both forms produce the same pass/fail result, reversing the arguments causes failure output to be misleading.

For example, when a learner's implementation returns `677` instead of the expected `67`, the current tests report:

```
Expected: 677
  Actual: 67
```

which inverts the intended meaning and can make debugging more difficult.

After this change, the same failure is reported as:

```
Expected: 67
  Actual: 677
```

which aligns with standard Minitest behavior and clearly communicates what the implementation should have returned.

## Changes

* Reordered arguments in all `assert_equal` assertions to use `expected, actual`.
* No behavioral changes to the tests themselves.
* Improves readability and debugging experience for learners.

## Impact

This change does not affect test outcomes. It only improves the clarity of failure messages when assertions fail.

### DIFF
--- a/exercises/concept/last-will/last_will_test.rb
+++ b/exercises/concept/last-will/last_will_test.rb
@@ -3,39 +3,38 @@ require_relative 'last_will'
 
 class LastWillTest < Minitest::Test
   def test_family_secrets_unaltered
-    # skip
-    assert_equal ::Zhang.bank_number_part(1), 8_541
-    assert_equal ::Zhang.bank_number_part(3), 8_541 * 3 % 10_000
-    assert_equal ::Khan.bank_number_part(1), 4_142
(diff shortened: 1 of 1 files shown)


kotp (approved):

kotp:
Thank you for the fix.  It is sent in and does not invalidate current exercises.

You will want to check the forums, in case work is happening (such as the test generators) so that conflicting work is not done for no good reason.

I appreciate the contribution!

PR #1849, opened 2026-08-02
Fix: Transpose - must skip only first test

kotp:
Only the first test should be skipped when delivered to the student.

fixes: 1846


IsaacG (approved):

IsaacG:
Was this generated with the test generator? Does the test generator need fixing?

